PR #17, opened 2023-03-06
Added geo data update option

pulsarice:
I'm not a programmer. I took some lines of code from NidukaAkalanka/x-ui-english and some from maddie/update-v2ray-geo.sh and made some tweaks. sorry for any mistakes or formatting errors.

### DIFF
--- a/x-ui.sh
+++ b/x-ui.sh
@@ -408,6 +408,36 @@ show_xray_status() {
     fi
 }
 
+update_geo() {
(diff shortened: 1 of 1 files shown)


MHSanaei:
hi
thank you for your work 
but we don't need this because, after every release, we update geo files with workflows/release.yml, 
it's good for NidukaAkalanka because they update very late 
I hope I see you more with a nice pr

PR #30, opened 2023-03-17
update

MHSanaei:


pchreza:
hi bro tnx for your god x-ui 
How can several outbands be connected to x-ui as proxy servers?

Mostafa-kaveh:
Hello
Thank you very much for your hard work
In the new update, I cannot add a user with a new email

allahyar:
Hello
SSL that is created, the panel does not accept, what is the problem?
image

Moein-1986:
Hello
Tls is not working in new update
Please fix that

pchreza:
> Hello SSL that is created, the panel does not accept, what is the problem? image
In this error, it says that there is no such file in this path, you should check your path or move your certificate files.

PR #346, opened 2023-05-05
Some new buttons for bot and ability to use userId in tgId

masoud-hidden:
Added buttons:
IP Log
IP Limit
Enable/Disable

Ability to use the telegram User ID in the Telegram Username field.


MHSanaei:
nice thank you for your great work, I just checked everything works fine except "Ability to use a telegram User ID in Telegram Username"
if I'm not wrong you mean we can set a user id for the client?
I test that but doesn't work

masoud-hidden:
> nice thank you for your great work, I just checked everything works fine except "Ability to use a telegram User ID in Telegram Username" if I'm not wrong you mean we can set a user id for the client? I test that but doesn't work

Yes, You can set a user id for the client. Add a number sign (#) before the user id and set it to the Telegram Username field in the client settings. Test it with a normal telegram user not an admin user.
Like this:
User ID = 987654321
Telegram Username (in the client settings) = #987654321

MHSanaei:
i test it again but as you can see but still doesn't work like when we use username
Screenshot 2023-05-06 014818

masoud-hidden:
> i test it again but as you can see but still doesn't work like when we use username Screenshot 2023-05-06 014818

Can I see what you received from the bot when the user id is set?

MHSanaei:
Your configuration is not found!
Please ask your Admin to use your telegram username in your configuration(s).

Your username: @MHSanaei1

masoud-hidden:
I will fix it soon.

masoud-hidden:
I can remove the (#) in the next commit, Is it good?

MHSanaei:
yes it will be even better thanks for that

masoud-hidden:
Don't forget to remove the # from the client settings.

MHSanaei:
well done

PR #431, opened 2023-05-13
[HOTFIX] Redirect `/xui` to `/panel`

hamid-gh98:
- Added Redirect Middleware to redirect all paths from `/xui` to `/panel`
- Changed API path to `/panel/api/inbound` (it will also redirect from the old path to this new path)
- Update readme.md


MHSanaei:
work well except when they have root path

hamid-gh98:
> work well except when they have root path

Yeah, I saw that. I fixed it in the new PR that I just submitted.

PR #432, opened 2023-05-13
[HOTFIX] Add basePath to Redirect Middleware

hamid-gh98:
Hi,
- Add basePath when redirecting with middleware


MHSanaei:
we have another problem I guess it comes with your last PR
when they use phone or minimize the panel 
this button doesn't work
photo_2023-05-14_01-47-20

MHSanaei:
@hamid-gh98 do you have any idea about this?

hamid-gh98:
> @hamid-gh98 do you have any idea about this?

I just changed its className. I don't think it's causing the issue, but i will check it.

MHSanaei:
@hamid-gh98 
مشکل از این کامیت هست 
https://github.com/MHSanaei/3x-ui/commit/0e266b88f0eb70bf46d3ecc4df148794bc307205

MHSanaei:
قسمت siderDrawer 
در صفحه common_sider.html مشکل داره

hamid-gh98:
> قسمت siderDrawer در صفحه common_sider.html مشکل داره

I just tested it with the old 'common_sider.html' file, but it's still like this. i think something else caused this issue.

If you log change function, you'll see that it changes the 'visible' value immediately (in JS), but updating the UI takes some time.

PR #551, opened 2023-05-31
Updated install.sh to match new release.yml

somebodywashere:
Also reverted xray path in release.yml back to xray-linux-${{ matrix.platform }} (cuz didn't find where to fix this path in other files, probably it's just easier to keep this path in workflow).


MHSanaei:
ah you are right i forgot Linux before matrix.platform
just need to add linux like this 
mv xray xray-linux-${{ matrix.platform }}
don't need to remove linux on install.sh

somebodywashere:
No problem, I thought -linux eas dropped on purpose 😅 So just need to fix a Workflow then.

MHSanaei:
[forgot linux tag name](https://github.com/MHSanaei/3x-ui/commit/48dce38e14ffe3f96d4a9a5b85f6971be34d5448)

PR #580, opened 2023-06-09
Removed ipLimit from the panel, changed the code to work with fail2ban

somebodywashere:
Hello! A wasted a couple of days trying to make stable and reliable Device limit with the base code (thorught cmd module you used).
But I made a decision to drop it completly from the panel because it still not that reliable, stable, pretty heave for machine and not configurable at all for end user.
So I just made logging of ips that get added to disAllowedIIps and configured fail2ban. It works like a charm now for me (tested 1 full day).

I did not fully removed whole code, because I noticed that disAllowed Ips somehow interact with XrayBlockedIPS.

fail2ban is pretty lightweight service that could be configured to use with UFW (I'm actually doing it) and perfectly suits for limiting devices aswell.

In my mind I think IP Limit should be optional and added to the main menu to configure (like WARP). Made a quick draft of commands to work:

`sudo apt-get install fail2ban -y`
`sudo cp /etc/fail2ban/jail.conf /etc/fail2ban/jail.local`
`sudo echo $'\n[3x-ipl]\nenabled=true\nfilter=3x-ipl\naction=iptables-allports\nlogpath=/var/log/daemon.log\nmaxretry=3\nfindtime=150\nbantime=300' >> /etc/fail2ban/jail.local`
`sudo echo $'[Definition]\nfailregex = [LIMIT_IP].+SRC= <HOST>\nignoreregex =' >> /etc/fail2ban/filter.d/3x-ipl.conf`

My config for jail:

[3x-ipl]
enabled=true
filter=3x-ipl
action=iptables-allports
logpath=/var/log/daemon.log
maxretry=3
findtime=150
bantime=300

Still deciding of what will fit the best. Also you can integrate a few "presets" of configs like above.
image


MHSanaei:
hi @somebodywashere 
thank you for your great work 
I just test it works fine there is something that could be better if you move it to somewhere else 
ip log shouldn't be here on x-ui log

Screenshot 2023-06-14 182538

somebodywashere:
> hi @somebodywashere thank you for your great work I just test it works fine there is something that could be better if you move it to somewhere else ip log shouldn't be here on x-ui log

Hello. Sure, I agree with you and will try to do so in a few days. 

Actually I configured some "recidive" system for myself, (15m ban 2 times, next ban is 1hr, and next ones 3-6-12-24) and I'm not sure about does it need to be included in a panel in a simple manner or nah. For example smth like 5m-30m-3h-12h?

What do you think. I'm thinking it's pretty overkill and doesn't needed to be included in a base.

TalaDev:
من xui رو اپدیت کردم و دستورات رو اجرا کردم اما کار نمیکنه

realartin:
could you please port this for @hiddify1 panel ?

AliBelali:
@somebodywashere thanks for your great work
I think fail2ban have a little problem with docker.
As it uses iptables (or ufw) and it isn't normally accessable in docker (it needs net-admin flag that is not accessable in some scenarios) do you suggest any alternatives

AliBelali:
@somebodywashere thanks for your great work
I think fail2ban have a little problem for servers that are behind cdn. 
actually I think it wouldn't work at all specially with websocket because when using a cdn or reverse proxy in another machine , packets will be forwarded to x-ui server with "src ip of cdn machine in **packet's header**" and "src ip of client in **http header**" x-ui uses src ip in http header to detect client's source ip address (it's logical because otherwise it sees all clients with same ip that is ip of cdn machine) so if a user breaks ip limitation , fail2ban gets client's ip from x-ui (ip in http header) and will drop client's ip with firewall but firewalls work with packet's header (not http header specially if using tls - In tls http header is encrypted) and ip header of all client are same as cdn ip so it fail2ban has no effect
@MHSanaei

PR #991, opened 2023-08-30
Fail2ban banned log command added

AmoAlireza:
hi
command `x-ui banlog` added, now you can check the banned log without going to IP Limit Management.


MHSanaei:
Thank you 🙏

PR #1026, opened 2023-09-06
[sub] Show used traffic for unlimited users

AmoAlireza:
hi
if user total flow and expiry date was unlimited, now will show the total flow used.

flow


MHSanaei:
Thank you for your PR.

But I think it's useless because we don't need to know the traffic usage when it's unlimited.

PR #1268, opened 2023-11-28
Change port check

### LINKED ISSUE #860: Sub and Web could not use same port
### Welcome

- [X] Yes, I'm using the latest major release. Only such installations are supported.
- [X] Yes, I'm using the supported system. Only such systems are supported.
- [X] Yes, I have read all WIKI document,nothing can help me in my problem.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, log, etc).

### Description of the problem,screencshot would be good

Hello, please match ip:port pairs for exclude these error (at now match only port) [here](https://github.com/MHSanaei/3x-ui/blob/329889ec006bdaed62609dbe012c950a87ea91c3/web/entity/entity.go#L83C1-L83C1)

I use 3x-ui with nginx reverse-proxy (hide all in 80,443 port)
In my setup ip:port used:
```
12.34.56.78:80 - nginx (redirect to https)
12.34.56.78:443 - nginx (stream_ssl_preread, routes to xray,panel,subscription)
127.0.1.1:443 (xray-reality)
127.0.1.1:80 (panel)
127.0.1.2:80 (subscription)
```
It's absolutely correct, but can't be saved with error:
`Modify Settings Fail: Sub and Web could not use same port: 80`
At now I edit database directly, for made do work (and it's work as expected).

If use different ports (8080 or other) subscription/panel generated links was broken:
http://domain.example:8080/sub/1kdeji89387frgj59

Also good idea make checkbox "use https" to generated links (at now all QR http://) 
for case when 3x-ui behind reverse-proxy.

### Version of 3x-ui

3X: v1.7.5
Xray: v1.8.1

### x-ui log reports or xray log

Not need

guard43ru:
We can use same port on difference IP's.
Fix #860

### DIFF
--- a/web/entity/entity.go
+++ b/web/entity/entity.go
@@ -82,8 +82,8 @@ func (s *AllSetting) CheckValid() error {
 		return common.NewError("Sub port is not a valid port:", s.SubPort)
 	}
 
-	if s.SubPort == s.WebPort {
-		return common.NewError("Sub and Web could not use same port:", s.SubPort)
+	if (s.SubPort == s.WebPort) && (s.WebListen == s.SubListen) {
+		return common.NewError("Sub and Web could not use same ip:port, ", s.SubListen, ":", s.SubPort, " & ", s.WebListen, ":", s.WebPort)
 	}
 
 	if s.WebCertFile != "" || s.WebKeyFile != "" {


MHSanaei:
thank you

PR #1299, opened 2023-12-08
Fixing build for ARM64

vlasov-y:
I have found out that your arm64 Docker images does not work because they still contain build for x86_64. I have managed to make Dockerfile build respective version properly basing on TARGETOS and TARGETARCH env variables. I have tested locally - now both amd64 and arm64 images contain proper binaries.

1. I have followed your releases GitHub Actions pipeline and used cross-compilation
2. Switched builder to debian bookworm based image
3. Switched runner to debian:bookworm
4. Set runner platform to TARGETPLATFORM to use arm64 on image layer
5. Copy go.mod first and then everything else - it helps proper build layer caching

Notes about 3:
Dynamically compiled binary on debian, does not work on alpine.

Now I see this on the linux/arm64/v8 build.

image

And I see this on linux/amd64 build.

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,52 +1,74 @@
 # ========================================================
 # Stage: Builder
 # ========================================================
-FROM --platform=$BUILDPLATFORM golang:1.21-alpine AS builder
+FROM --platform=$BUILDPLATFORM golang:1.21.5-bookworm AS builder
 WORKDIR /app
+
 ARG TARGETARCH
 ARG TARGETOS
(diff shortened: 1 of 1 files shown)


vlasov-y:
Tried to run, received
```
2023/12/08 16:34:33 Binary was compiled with 'CGO_ENABLED=0', go-sqlite3 requires cgo to work. This is a stub
```

So CGO is required... How to fix it?

vlasov-y:
Succeeded to use cross compilation.

PR #1500, opened 2024-01-02
socks5 proxy option added to telegram bot settings

surbiks:
In some cases, we need to install the panel on a server that does not have access to Telegram. A proxy must be used for this purpose. In this patch, we added proxy settings to the Telegram bot section. for this, you only need to define a socks5 inbound for the panel at the address 127.0.0.1 with a random port and place this setting in the Telegram bot section. In this way, the Telegram bot uses the panel itself to connect.


MHSanaei:
Hi, thank you for your PR.

If possible, could you provide translations for all languages?

Thanks again!

surbiks:
> Hi, thank you for your PR.
> 
> If possible, could you provide translations for all languages?
> 
> Thanks again!

Hi
Yes, translations added.

SimpoZi:
Hi, is there a possibility to add a notification to the client in the telegram about the imminent expiration?

PR #1566, opened 2024-01-12
A few updates

quydang04:
### Hello, in this pull request, I have made some small changes.

1. I have fixed some typos and edited support for Debian 11 and above because Debian 10 will drop support in July 2024.
2. I have added support for more OSs, specifically RockyLinux. Another distro worthy of being an alternative to Centos, along with Almalinux.

_Can you review it and consider updating it in the next version?
Thank you very much._


MHSanaei:
Thanks 👍

PR #1571, opened 2024-01-12
Small edit & fixes

shahin-io:
Hello, dear @MHSanaei 

I’ve made some adjustments across the board for enhancement.


web/html/xui/form/tls_settings.html
```diff
@@ -171,7 +171,7 @@
                     <template slot="title">
                         <span>{{ i18n "reset" }}</span>
                     </template>
-                    Short ID
+                    Short IDs
```
MHSanaei: We have just one short id not like alireza 3-4 ids 
So it's not correct
Undo your change

shahin-io: Done.

Issue #1456, opened 2023-12-26
Two inbounds on the same port, different IP stack


pulsarice:
### Welcome

- [X] Yes, I'm using the latest major release. Only such installations are supported.
- [X] Yes, I'm using the supported system. Only such systems are supported.
- [X] Yes, I have read all WIKI document,nothing can help me in my problem.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, log, etc).

### Description of the problem,screencshot would be good

Can you implement a feature to use the same port on different IPs for inbounds?
I want to have two inbounds both listening on port 443, the first one on IPv4 and the second one on IPv6
Thanks.

### Version of 3x-ui

2.0.2

### x-ui log reports or xray log

null

PR #1595, opened 2024-01-16
Fix inbounds with the same port and different IPs

amintn:
Currently in 3x-ui, we cannot create multiple inbounds on the same port but with different listening IPs, although it is possible in the xray core. related issues: #520 and #1456

The problem is that "checkPortExist" function only checks the database for existence of the said port and doesn't check the listening IP. I've changed this function 
to also take listening IP into consideration.

Basically I've changed the function to check If there's an inbound with listening IP of INADDR_ANY ("", 0.0.0.0, ::, ::0) then no more inbounds are allowed to be created and also if there's an inbound with a specific IP, then no more inbounds with the same IP port or INADDR_ANY and the same port are allowed to be created.

With this approach users who don't use this feature (leave listen field empty) will not be affected and their inbounds will still listen on all IPs and those users who want this feature will be able to use it.

Also unique constraint on port field in inbounds table had to be dropped because there now can be more than one inbound per port.

Screenshot of it in action
panel
terminal

### DIFF
--- a/database/model/model.go
+++ b/database/model/model.go
@@ -37,7 +37,7 @@ type Inbound struct {
 
 	// config part
 	Listen         string   `json:"listen" form:"listen"`
-	Port           int      `json:"port" form:"port" gorm:"unique"`
+	Port           int      `json:"port" form:"port"`
 	Protocol       Protocol `json:"protocol" form:"protocol"`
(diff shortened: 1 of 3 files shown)


MHSanaei:
thank you

PR #1650, opened 2024-01-20
add group user with the same subscription id to all inbounds

alirahimi818:
add group user with the same subscription id to all inbounds (just added)

Untitled


MHSanaei:
thanks for your pr 
I checked, and it looks good, but when there is no inbound, we have an issue.

Screenshot 2024-01-21 151219

alirahimi818:
Hi,

Is there a problem with this feature?

PR #1668, opened 2024-01-23
persian datepicker bug fixed

alirahimi818:
persian datepicker bug fixed


MHSanaei:
After your changes, we have a bug on export and import inbound. I couldn't find the problem.

alirahimi818:
OK, I found one, but after fixing it, I still get the error (but this error is present in the current version 2.1.1).

PR #1678, opened 2024-01-25
Progressive Web App

Incognito-Coder:


StayAway32:
regards to your wonderfull job of adding pwa and thanks to Mhsanaei and everyone else involved to make such a wonderfull fork

can you also consider if the project could be converted to use h2 instead of http:1.1 ?
i know the panel is just for admin but some of us have it open 24 hour a day being monitored by multiple people and i think the change and benefits that comes with using h2 could be significant
i am inexperienced with http matters as such i cant be of much use ! but if there is anything i can do for this matter or anyother to help the project call me in :)

MHSanaei:
thank you

PR #1704, opened 2024-01-29
fix bug in edit SOCKS and HTTP outbound

surbiks:
Hi. In the outbound tab, when you want to edit Socks and HTTP, the outbound settings are not loaded, and an error is displayed in the browser console. This problem is fixed in this patch.


MHSanaei:
i didn't know that , thank you

PR #1711, opened 2024-01-29
Show outbound traffic in outbounds table

surbiks:


MHSanaei:
nice i like this one 
thank you

PR #1759, opened 2024-02-06
Manage balancers in settings UI

surbiks:
Hi. One of the parts that needed to be managed in the UI was the balancers part. The management of this section was added to the xray settings. Also, the ability to define and manage routes based on balancer has been added in the routing section.


MHSanaei:
Thanks for enhancing the UI with balancer management and route definition. Great work!

PR #1788, opened 2024-02-10
Enhance IP Limitation using Xray Stdout and iptables

mehdikhody:
This pull request aims to improve IP limitation mechanisms by replacing the fail2ban tool and log files with a more efficient solution. Instead of relying on external tools and file operations, this enhancement proposes leveraging xray stdout to extract client email and connected IP addresses.

Key Changes:
1. **Direct Xray Integration:** Instead of relying on intermediate log files, the system now reads IP events directly from the stdout of the xray tool. This eliminates the need for file I/O operations and simplifies the overall process.
2. **Native iptables Commands:** The IP limitation mechanism now utilizes native iptables commands to ban disallowed IPs. This approach improves efficiency and reduces latency by avoiding external tools and file operations.
3. **Real-time Response:** With direct integration and native command usage, the system can respond to IP events in real-time as they occur, enhancing the responsiveness of the IP limitation mechanism.
4. **Simplified Workflow:** The new flow significantly simplifies the workflow by removing unnecessary intermediate steps.
5. **Enhanced Reliability:** By eliminating potential points of failure associated with file I/O operations, the reliability of the IP limitation mechanism is improved, ensuring consistent and accurate enforcement of IP restrictions.
6. **Elimination of fail2ban:** there is no need of fail2ban or any kind of log files.
7. **Improved Efficiency:** By avoiding unnecessary I/O operations and utilizing system-native functionalities, this enhancement offers a more streamlined and efficient IP limitation mechanism.
8. **Better for lower specs machines:** This potentially can prevent the database look, because of lower I/O operations.
9. **Fallback to Previous Mechanism:** If the access log is populated, the system will revert to using fail2ban and log files for IP limitation.

Your review and feedback on these proposed changes are highly appreciated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 *.tar.gz
 access.log
 error.log
+.env
 tmp
 main
 backup/
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ require (
 	github.com/gin-contrib/gzip v0.0.6
 	github.com/gin-gonic/gin v1.9.1
 	github.com/goccy/go-json v0.10.2
+	github.com/joho/godotenv v1.5.1
 	github.com/mymmrac/telego v0.28.0
 	github.com/nicksnyder/go-i18n/v2 v2.4.0
 	github.com/op/go-logging v0.0.0-20160315200505-970db520ece7
--- a/go.sum
+++ b/go.sum
@@ -131,6 +131,8 @@ github.com/jinzhu/inflection v1.0.0 h1:K317FqzuhWc8YvSVlFMCCUb36O/S9MCKRDI7QkRKD
 github.com/jinzhu/inflection v1.0.0/go.mod h1:h+uFLlag+Qp1Va5pdKtLDYj+kHp5pxUVkryuEj+Srlc=
 github.com/jinzhu/now v1.1.5 h1:/o9tlHleP7gOFmsnYNz3RGnqzefHA47wQpKrrdTIwXQ=
 github.com/jinzhu/now v1.1.5/go.mod h1:d3SSVoowX0Lcu0IBviAWJpolVfI5UJVZZ7cO71lE/z8=
+github.com/joho/godotenv v1.5.1 h1:7eLL/+HRGLY0ldzfGMeQkb7vMd0as4CfYvUVzLqw0N0=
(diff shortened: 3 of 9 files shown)


MHSanaei:
Thanks for your PR. 
but I don't want to add something else for IPlimit because soon we will have this
https://github.com/XTLS/Xray-core/pull/2277
and all our code about IP limits will be removed.

mehdikhody:
Thank you for your feedback. Will this work on panels with older xray-core as well?

woxie:
I just tested it and it works better than the current implementation. Thank you!

mehdikhody:
The path to version 1.9 is quite long, especially considering that we are currently on version 1.8.7. Many devices still do not support version 1.8.7, and people many will set up the panel on an older xray-core versions. I have tested this feature with another developer, and it works. However, the current IP limit is still not precise.

I request that you add this feature until the presentation of the new core version.

somebodywashere:
Your pros is all on point.
But there's many drawbacks for my point of view.
1. First of all, on 9th. If we get rid of fail2ban, we need to get rid of it fully, not based on access.log usage or not.
1.1 Also you can configure fail2ban to listen stdout aswell.
2. Bantime is not configurable.
3. You can configure precision by doing tweaks in jail configuration. (Not from the panel or bash, but you can in /etc/fail2ban/jail.d/3xipl.conf file).

It's precise as it is right now because you don't want to block an IP if it appears just one or two times. Maybe someone just switching from Wi-Fi to Mobile network and back? You NEED to make sure it's really two people/IPs connected for a some period of time.
In your implementation it's unable to correct/adapt precision in anyway.

mehdikhody:
> You NEED to make sure it's really two people/IPs connected for a some period of time. In your implementation it's unable to correct/adapt precision in anyway.

Every 10 seconds (`check_client_ip_job` interval), all IPs will be unblocked once, and then the additional IPs will be blocked again.

Chipmax64:
> Thank you for your feedback. Will this work on panels with older xray-core as well?

I greet you.
I decided to try your IP restriction option.
I configured the reality connection and logging, set a limit of no more than one ip.
In the ip log window I see 3 ips. But there is no blocking; all three clients are working.
What could be the problem?

PR #1801, opened 2024-02-11
Update Uninstall Option

jalalsaberi:
after uninstall, script will delete itself and show Install & Upgrade command for installing again if user need that.


MHSanaei:
thank you

PR #1854, opened 2024-02-20
Fix Dockerfile

LOVECHEN:
Fix wrong image


MHSanaei:
thank you i forgot that

PR #1864, opened 2024-02-21
Update xray.html

shahin-io:
Dear @MHSanaei 

I’ve added missing balancerTag desc along with minor style changes.


MHSanaei:
thank you

PR #1889, opened 2024-02-23
Optimize Dark Theme & Ultra Dark 

TaraRostami:
Note: The Ultra Dark theme will be added later as an option, and users can activate it as desired.

Like below:
Enable Ultra


Arman-nsb:
Thanks, this is way better! Also, the hover effect on the dashboard widgets doesn't look good. I'd suggest removing it.

shahinDOT:
I believe it would be beneficial to implement a dropdown menu in the settings to allow users to choose the type of dark mode that suits their preferences. Generally, users select one of these themes for night vision style. With this approach, the side menu will remain clean, as it was before, while also maintaining consistency. This is a common practice and is often preferred.

PR #1997, opened 2024-03-08
Some fixes and improvements

hamid-gh98:
Hello,

1. Fix access log path (maybe some ppl set other paths for access log)
2. Fix select options for logs in xray settings by the paths in config.json
3. refactor api controller
4. add status code to pureMsgJson to use in future
5. response with 401 to unauthorize the requests
6. Fix session timeout by redirecting via axios interceptors

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 .cache
 .sync*
 *.tar.gz
+*.log
 access.log
 error.log
 tmp
--- a/README.md
+++ b/README.md
@@ -311,9 +311,9 @@ If you want to use routing to WARP before v2.1.0 follow steps as below:
   
   ```sh
     "log": {
-    "access": "./access.log",
-    "dnsLog": false,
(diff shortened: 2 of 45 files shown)


MHSanaei:
hi hamid 
thanks for your pr 
there is two problem with your changes
one of them as you see 
Screenshot 2024-03-10 232019

another one when there is no inbound or iplimit active we get this 
 WARNING - Access log is set to 'none', check your Xray Configs

hamid-gh98:
Hello, hossein.

thanks for your review.

1. now we only warning if f2b is installed.
2. logs paths fixed (the problem was when config log path is set to 'none')

now both is fixed

FlameFandub:
Hello. Can you help me? 
image
I use this version of pannel. When I create new user, i cant setup ip connections restrict for each one in "Client" part. 
Can you help me how to fix this?

PR #2055, opened 2024-03-14
Add install.sh support for alpine linux

maple367:
Add install.sh support for alpine linux.
Please ignore the edit in README.md


a-poluyanov:
Why did you reject this change? I've been waiting for it for almost a year and now I came looking for it and it's still not there
The problem is that you are using 3x-ui in Docker -> Docker is based on alpine
screeen

I want to install fail2ban inside the container to block ip and have this function finally work
But I get an error
screeen

Moreover, he mistakenly says that everything was installed successfully! How? After all, it didn’t find systemctl, couldn’t add packages, and therefore the IP restriction function doesn’t work.

I don’t pretend that you should accept edits from this particular person, but you definitely need to support alpine, otherwise why are you releasing your product in Docker based on alpine linux?

I think that alpine should be left as a base for Docker, since it is the best option for security, but add support for it so that fail2ban works inside the Docker image or offer people another solution that should work, it doesn’t work now

PR #2103, opened 2024-03-20
Add gormStatement.RaiseErrorOnNotFound in few places

somebodywashere:


somebodywashere:
For some reason error still is not raised. I'll just change everything for manual check

somebodywashere:
Doesn't work for me, idk why :(

PR #2225, opened 2024-04-12
Update custom.css

jahlib:
remove  max-height 400px from .ant-popover-inner-content


jahlib:
maybe 1000px not full remove, but it need to be larger, maybe 800px

TaraRostami (commented):
This item should not be deleted or changed.

jahlib:
why not? I don’t insist, but the number of my users does not fit into 400px, I think I’m not the only one

PR #2229, opened 2024-04-20
fix dns ui

rammiah:
after show a sever with domains, open a server without domains will show last server domains
fix-dns


MHSanaei:
Thank you

PR #2230, opened 2024-04-20
Update x-ui.sh with ufw port settings

Athameral:
It really costs time when adding rules for a large range, if for loop is used in bash. Changed it to built-in port range support (with colon) in ufw.


MHSanaei:
thank you

PR #2232, opened 2024-04-23
Update README.md

AliBelali:
There is no need to clone the whole project when using "docker compose".
 docker-compose.yml file is enough.


AliBelali:
There were no ports specified in the "docker run" command, and the host firewall didn’t perform destination NAT to the container in case of using "docker run"

PR #2248, opened 2024-05-09
[fix] always serve panel even if xray failed to run

hamid-gh98:


evstratbg:
hi @MHSanaei,

sorry for mentioning here, but there is no "issue" tab. There is a problem, sometime, when I add users using API, a user gets added without a `method` key, which results xray being stopped

in sqlite, in `inbounds` table, `settings` column, `clients` key, I found 

```
    {
        "alterId": 0,
        "email": "fewfewfewf",
        "enable": true,
        "expiryTime": 0,
        "id": "ewfwef",
        "limitIp": 2,
        "password": "PASSWORD,
        "subId": "",
        "tgId": "",
        "totalGB": 0
    },
    ```
method is missing, and xray got stopped because of it. Can you help me with this?

Issue #2284, opened 2024-05-22
Loss of access to the panel


KoPobkaTopta:
### Welcome

- [X] Yes, I'm using the latest major release. Only such installations are supported.
- [X] Yes, I'm using the supported system. Only such systems are supported.
- [X] Yes, I have read all WIKI document,nothing can help me in my problem.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, log, etc).

### Description of the problem,screencshot would be good

After entering the listening domain, access to the panel is lost.In the old version (v2.3.1), everything works fine.Error 403

### Version of 3x-ui

<details>

```console
# v2.3.3
```

</details>

### x-ui log reports or xray log

none

allury:
I'm also running into this issue and my panel certificate is set to the correct path

meta-syfu:
![Screenshot (425)](https://github.com/MHSanaei/3x-ui/assets/135529333/8468a2e8-8225-44ef-ae9c-a81c0cfb69d1)

Yes. This port is API port and cause of this xray doesn't run and rapidly errors

MHSanaei:
use `x-ui log` and find your problem

KoPobkaTopta:
> use `x-ui log` and find your problem

Nothing was written there. Right now I’m on the old version. So I can’t check it again.Nothing was written there. Right now I’m on the old version. So I can’t check it again.

MHSanaei:
If you can update it one more time

KoPobkaTopta:
> If you can update it one more time

I've just updated the 3x-ui. Upon entering the panel, he writes access is denied.If I remove the listening domain, I can log in to the domain, and if I set it, the HTTP ERROR 403 error

KoPobkaTopta:
> If you can update it one more time

I just tried to update again and writes this in the logs: WARNING - get x ray traffic failed: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp 127.0.0.1:62789: connect: connection refused"

PR #2292, opened 2024-05-23
feat: Enhance host extraction from headers

najahiiii:
- Refactor SUBController subs and subJsons methods to extract host from X-Forwarded-Host header, falling back to X-Real-IP header and then to the request host if unavailable.
- Update html function to extract host from X-Forwarded-Host header, falling back to X-Real-IP header and then to the request host if unavailable.
- Update DomainValidatorMiddleware to first attempt to extract host from X-Forwarded-Host header, falling back to X-Real-IP header and then to the request host.

Fixes: #2284


MHSanaei:
Thank you

PR #2305, opened 2024-05-25
Add CLI to set panel cert

### LINKED ISSUE #2304: Set panel cert path from CLI
The only way to set panel cert is to from the panel using http.
By adding a command to set the panel cert we can avoid using panel in http mode to set the panel cert

MaisamV:
We will be able to set panel cert using bellow command:
/usr/local/x-ui/x-ui cert -webCert /path/to.cert -webCertKey /path/to.key
thiw PR will resolves #2304


MHSanaei:
Thanks, I'll add an option to certificate management "set your cert to the panel" button and ask for the cert path

PR #2427, opened 2024-06-24
Update settings.html

GFW4Fun:
The code is copied and applied from the Alireza0/xui

- Fix Restart Panel in New URL
- Fix Restart Panel in 'x-ui-pro'


MHSanaei:
the panel already handles redirection to a new path and port

MHSanaei:
Hello, 
thank you for your PR, but the changes made are unnecessary

PR #2433, opened 2024-07-09
fix wrong splithttp default

mmmray:
This default is defined as 1MB, but maxUploadSize is to be specified in bytes. This confusion could've come from poorly written documentation in xray, but it has been updated.

in general I wish that panels would not set defaults at all and instead just omit parameters (in sharelinks, inbounds, ...) that the user didn't set explicitly. If I want to change the defaults in xray's codebase, it seems that all the panels will have to update the default too.

I see marzban doing the same kind of things.


MHSanaei:
Thanks for your pull request! I've merged it.

I'm glad the docs are updated. I follow your work on Xray Core and appreciate it.

Thanks again!

PR #2446, opened 2024-08-06
API - Add Client to inbounds

MrR4di0k4l:
## API - Add Client to inbounds

If the list of inboundIds is sent empty, a user will be created on all existing inbounds, otherwise, the inbound ID must be entered as follows:
"inboundIds": [1, 2]

## Example
```python
import requests

url = "http://localhost:2053/panel/api/inbounds/addClientInbounds"

payload = {'settings': '{"clients":[{"id":"983ecd09-6257-411d-a58c-fa7473ca3c0c","flow":"","email":"SpeedVpn","limitIp":2,"totalGB":32212254720,"expiryTime":1712927113000,"enable":true,"tgId":"","subId":"subId","reset":0}]}', 'inboundIds': []}
headers = {
  'Accept': 'application/json',
  'Cookie': '3x-ui=MTcyMjkzOTQwNnxEWDhFQVFMX2dBQUJFQUVRQUFCMV80QUFBUVp6ZEhKcGJtY01EQUFLVEU5SFNVNWZWVk5GVWhoNExYVnBMMlJoZEdGaVlYTmxMMjF2WkdWc0xsVnpaWExfZ1FNQkFRUlZjMlZ5QWYtQ0FBRUVBUUpKWkFFRUFBRUlWWE5sY201aGJXVUJEQUFCQ0ZCaGMzTjNiM0prQVF3QUFRdE1iMmRwYmxObFkzSmxkQUVNQUFBQUZQLUNFUUVDQVFWaFpHMXBiZ0VGWVdSdGFXNEF8uVa-901gifdCXLToceyxtfnjMOvHDZer53Ks2O0V6gA='
}

response = requests.request("POST", url, headers=headers, json=payload)

print(response.text)

```


MHSanaei:
@alireza0 Can you please check this?

MrR4di0k4l:
I tested it with more than 700 users and 20 entries and it was ok. In addition, the problem of the following script is also solved with this api.
```sh
https://github.com/M4mmad/3xui-multi-protocol
```

CoCEmu:
hello,
how can i add custom v2ray config as outband in xray configuration?

gospodenkods:
Got the message
CPU usage is 2.97%, which is above the threshold of 0%
Are the CPU usage thresholds configured somewhere?
 Unfortunately, I couldn't find it anywhere

SomeBytee:
Hi, sorry that I'm asking my question here, i couldn't find any better way 
Is it possible to have multiple inbounds for a single user? 
for example, imagine we have 3 inbounds ( with different ports -8080-8081-8082, and I assign client-number-1 to each of them with the same UUID, 
but the problem is since the port is different on each inbound if the inbound-number-1 reach his limit, the other two inbounds still functional.

long story short
I want a single user to be manageable with a single inbound, not in all inbounds with different usages.
let me know if I had to make a child inbound or something or having a routing rule for this , this will helps a lot

Thanks in Advance!

Lopersios:
Hello my friend, I would like to know if you can implement a limit on the use of a key. how it was done on another fork https://github.com/FranzKafkaYu/x-ui

Lopersios:
Hi friend are you planning to add one key one user feature

PR #2495, opened 2024-08-17
Adjust Persian translate

MatinHQ:
Those definitions are singular in Persian, so it doesn't make sense to add a plural form

Also adjusted spacing for `expireIn` key

image


MHSanaei:
thank you

PR #2503, opened 2024-08-28
Correct Russian README

voronin9032:
Looks like Russian README was translated from English using machine translation. This pull request makes the README sound more natural and adds various fixes to it. In the original translation, some terms were translated but they should be in English, some things were translated incorrectly and so on. Now, the README is clearer and more unambiguous.


MHSanaei:
Yes, I used AI to translate it. Thanks for your PR

PR #2505, opened 2024-08-29
better autocomplete for password and token inputs

nukhtarov:
Some password managers were not handling inputs correctly, this version fixes that


MHSanaei:
thank you

PR #2507, opened 2024-08-31
fix TypeError: this.shortIds.split is not a function

Rudomitori:
This PR fixes a bug with the following repro steps:
1. Open page "Inbounds"
2. Click "Add Inbound"
3. Select option "Reality" in field "Security"
4. Reset field "Short ID" via the icon button
5. Click button "Create"
  Expected: A new item is created
  Actual: Nothing happends and there is an error in the console: "TypeError: this.shortIds.split is not a function"
  
I've not tested the fix, but I don't see what it can break


MHSanaei:
Thank you for the fix, I wasn’t aware of that. I’ve tested everything, and it’s working perfectly.

PR #2579, opened 2024-10-10
simplified basic routing

alireza0:
Convert Basic Xray configuration to a simple block/direct dropdown list


MHSanaei:
Tack så mycket

PR #2581, opened 2024-10-11
feat(tgbot): Allow restart core via telegram bot

najahiiii:
Usage `/restart force`


MHSanaei:
Hi @najahiiii 
Thanks for your works, but i don't think we need a restart Xray option on Tgbot, do you have any specific reason for this?

najahiiii:
Hi, Sometimes I need to restart xray-core and I personally lazy to open server nor open web panel for restart it. so using tgbot will help

PR #2584, opened 2024-10-16
feat(tgbot): Add the option to change the telegram API server

MadrinX:
If someone has a [custom bot API  server](https://core.telegram.org/bots/api#using-a-local-bot-api-server) or proxied API server this lets them use it.

### DIFF
--- a/web/assets/js/model/setting.js
+++ b/web/assets/js/model/setting.js
@@ -16,6 +16,7 @@ class AllSetting {
         this.tgBotEnable = false;
         this.tgBotToken = "";
         this.tgBotProxy = "";
+        this.tgBotAPIServer = "";
         this.tgBotChatId = "";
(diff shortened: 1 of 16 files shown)


MadrinX:
Hi, 
Indeed, my target is a proxy issue. If you have a tunneled server and the panel in the country is limited, it is safer to use the proxied API server than risky socks5.

you don't need to have a bot server, if you have a reverse proxy to the Telegram API, this option will also work for you (and you don't have to worry about being blocked in this case Because, for DPI, this is an http server use some json data over tls).

@najahiiii

najahiiii:
I see.
yea that will help some restriction too

I personally setup my own telegram bot api server too, for my other project.

anyway, thankyou for your work.

MHSanaei:
thank you

PR #2585, opened 2024-10-18
修复简体中文名称依然使用繁体名称的问题。

kilvn:


kilvn:
The old "Simplified Chinese" is a traditional Chinese text, not a true simplified name, which is ironic

MHSanaei:
I don’t know about the Chinese names, but we have two of them.
One of them is simplified Chinese, and the other is traditional Chinese.
Please check both for me. Thank you.

kilvn:
> I don’t know about the Chinese names, but we have two of them. One of them is simplified Chinese, and the other is traditional Chinese. Please check both for me. Thank you.

The “簡體中文” in the current JS file is the text typed by the Taiwanese person who added TW voice to this JS file using his Taiwanese language input method. This is not Simplified Chinese, and every time I log in to the backend, I feel awkward and sarcastic. Why is the text in Simplified Chinese traditional.

PR #2586, opened 2024-10-18
improve Web Base Path and Logs Management

sxlmnwb:
aditional:
`recheck webBasePath on config_after_install`

cc: @MHSanaei, @sxlmnwb

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -143,7 +143,7 @@ config_after_install() {
     local existing_webBasePath=$(/usr/local/x-ui/x-ui setting -show true | grep -Eo 'webBasePath: .+' | awk '{print $2}')
 
     # Check if username and password exist
-    if [[ -n "$existing_username" && -n "$existing_password" ]]; then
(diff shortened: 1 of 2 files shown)


MHSanaei:
Thanks for your PR, but I don't want the user to be able to remove the path. I even want to remove the path from the panel. Then I thought that if someone finds a way to log in, maybe they won't need to change the path anymore.
Sorry about that.

sxlmnwb:
> Thanks for your PR, but I don't want the user to be able to remove the path. I even want to remove the path from the panel. Then I thought that if someone finds a way to log in, maybe they won't need to change the path anymore. Sorry about that.

If you want to delete the path, it means that to access the panel, we just have to use the specified port? Is that right? @MHSanaei

MHSanaei:
I intend to remove the path from the setting on the panel so you can edit it from there.
You can change the port to anything you want; it doesn't matter.
The path is crucial and should not be left empty.

PR #2594, opened 2024-11-04
DNS Configuration 

zoheirkabuli:
I had DNS leak issue on my `Hetzner` server, updating DNS to Cloudflare fixed my problem, I think it's useful to update DNS by default

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -198,6 +198,24 @@ config_after_install() {
     /usr/local/x-ui/x-ui migrate
 }
 
+update_dns() {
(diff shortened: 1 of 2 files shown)


somebodywashere:
This is bad idea, especially adding immutable flag. That would lead to many issues with other packages (systemd-resolved, openresolv, etc..).

zoheirkabuli:
@somebodywashere 
thanks for your suggestion, I reverted changes and added the DNS Configuration to `x-ui` command

PR #2602, opened 2024-11-16
Error: "No Such File or Directory" While Installing Certs

Azoo224:
Hello, I found an issue with the "update - CF SSL Certificate" feature in the 3x-ui v2.4.7 update. During script execution, the following error occurs: 
"[INF] Certificate issued successfully, Installing... 
[Sat Nov 16 01:54:41 AM PST 2024] The domain 'example.com' seems to already have an ECC cert, let's use it.
[Sat Nov 16 01:54:41 AM PST 2024] Installing cert to: /root/cert-CF/example.com/fullchain.pem
/root/.acme.sh/acme.sh: line 5925: /root/cert-CF/example.com/fullchain.pem: No such file or directory
[ERR] Certificate installation failed, script exiting...
". 

Later, I discovered that the directory "${certPath}/${CF_Domain}" must exist before installing the certificate; otherwise, acme will throw such error. Therefore, I am submitting this PR and hope you can review it as soon as possible.

### DIFF
--- a/x-ui.sh
+++ b/x-ui.sh
@@ -1183,8 +1183,10 @@ ssl_cert_issue_CF() {
         fi
 
         # Install the certificate
+        mkdir -p ${certPath}/${CF_Domain}
+
         ~/.acme.sh/acme.sh --installcert -d ${CF_Domain} -d *.${CF_Domain} \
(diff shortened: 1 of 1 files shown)


MHSanaei:
Thank you

PR #2608, opened 2024-11-29
docker: refactor

s0me1newithhand7s:
- [x] Yes i've tested locally

## Why
1. Edited shebangs to be *right*, as [calling binaries trough `env` is right way](https://stackoverflow.com/questions/16365130/what-is-the-difference-between-usr-bin-env-bash-and-usr-bin-bash).
2. Switched to `golang:bookworm` in builder and `ubuntu:latest` in 3x-ui itself. It isn't so important for builder, but for `x-ui.sh` itself - it is.


s0me1newithhand7s:
i can rebase if needed

aski425:
please, it to "bookworm", because after I accidentally drop it, it is not convenient to restore it, because you have to do everything from the beginning, I would like it to be possible through cli and with docker can be done.

s0me1newithhand7s:
and **pls** don't silently close this PR
i am open to cooperate

KiGamji:
why would you switch from a smaller, lighter distro, to debian and ubuntu

s0me1newithhand7s:
> why would you switch from a smaller, lighter distro, to debian and ubuntu

"Switched to golang:bookworm in builder and ubuntu:latest in 3x-ui itself. It isn't so important for builder, but for x-ui.sh itself - it is."

alireza0:
IMHO:
1. Using `x-ui.sh` in docker is not recommended
2. Using unstable debian version `bookworm` is not recommended
3. Using very large `ubuntu` image has not any good point

Instead you can use one of these solutions:
1. Only use `noble` for all stages.
2. Use separate `alpine` image without any `x-ui.sh` script beside another image `noble` ( in both builder and 3x-ui ) with `x-ui.sh` script. Then you could have two images: `3x-ui:alpine` and `3x-ui:noble`

s0me1newithhand7s:
> Using `x-ui.sh` in docker is not recommended
fully agreed but why not fixing it bc we actually can?

> Only use `noble` for all stages.
OK, i'll do it ASAP

> Use separate `alpine` image without any `x-ui.sh` script beside another image `noble` ( in both builder and 3x-ui ) with `x-ui.sh` script. Then you could have two images: `3x-ui:alpine` and `3x-ui:noble`
sound a bit over-complicated as my goal was to *just fix*, but i'll do my best

s0me1newithhand7s:
let's cooperate then

KiGamji:
> 2\. Using unstable debian version `bookworm` is not recommended

huh? bookworm was released on 10 June 2023

s0me1newithhand7s:
> > 2. Using unstable debian version `bookworm` is not recommended
> 
> huh? bookworm was released on 10 June 2023

nwm i've switched to `golang:alpine` since building platform have no impact, at all

s0me1newithhand7s:
> I still don't get why would you use ubuntu/debian over alpine for the whole Dockerfile — the resulting image would be more heavy

i'm not rewriting whole `Dockerfile` - it is just a *refactor*

KiGamji:
the question still remains

s0me1newithhand7s:
> the question still remains

because it wasn't written originally this way, so i consider not to

KiGamji:
I did some tests — with alpine it took me 315 seconds to build and takes 285 MB of storage, while with ubuntu:devel it now takes 296 seconds and takes 404 MB.

Also, the container now crashes:
3x-ui  | Server ready
3x-ui  | /app/DockerEntrypoint.sh: 7: exec: /app/x-ui: not found

s0me1newithhand7s:
> I did some tests — with alpine it took me 315 seconds to build and takes 285 MB of storage, while with ubuntu:devel it now takes 296 seconds and takes 404 MB.
> 
> Also, the container now crashes: 3x-ui | Server ready 3x-ui | /app/DockerEntrypoint.sh: 7: exec: /app/x-ui: not found

hm strange, mine haven't this issue

s0me1newithhand7s:
but thanks for reporting, i'll try to fix it

s0me1newithhand7s:
sorry for not updating you; i'm trying to rewrite `Dockerfile` and test it locally

fuwaneko:
> > I still don't get why would you use ubuntu/debian over alpine for the whole Dockerfile — the resulting image would be more heavy
> 
> i'm not rewriting whole `Dockerfile` - it is just a _refactor_

You have no idea what you're doing, haven't you? I see absolutely no point in the changes proposed in this PR. Can you please provide any good rationale for these “refactors”?

s0me1newithhand7s:
> You have no idea what you're doing, haven't you? I see absolutely no point in the changes proposed in this PR. Can you please provide any good rationale for these “refactors”?

i'm trying to abandon process of using unsupported `alpine linux`. unsupported by who? by x-ui and 3x-ui.
why? even at docker it's means, as my friends and me found. :D

KiGamji:
unsupported? huh???

PR #2614, opened 2024-12-19
Fix firewall

LoST202:
I entered my "corrections" into the ufw code. I don't think I would have come here if you hadn't thought about why when trying to execute a command from the menu. I hope for your understanding. 

P.S My first commit in my live


MHSanaei:
Thank you

PR #2627, opened 2024-12-29
Update translation

LoST202:
Added "Comment" into all languages


MHSanaei:
Remove firewall change

PR #2631, opened 2024-12-30
Language improvements

LoST202:
- Added a "comment" in all languages + corrected the Russian translation

- Added Finnish languages

[edit 17:41]: new languages added


MHSanaei:
we don't need any more languages for the panel, you also have so many problem on them
I'll remove them

LoST202:
> we don't need any more languages for the panel, you also have so many problem on them
> I'll remove them 

Ok

PR #2632, opened 2025-01-02
fail2ban service in docker container

simfishing:
Add option to enable (default) or disable the fail2ban service in docker container


almirus:
Will this setting appear in the UI automatically?

Lakhtion:
How to set up an IP limit in a container correctly? The script x-ui still does not see the service, IP addresses do not end up in jail. Logs are enabled in the panel, the limit 1 is set for test client. I check it wich fail2ban-client command inside container.

PR #2633, opened 2025-01-05
Update dependencies

alphax-hue3682:


MHSanaei:
Hi @alphax-hue3682 
sorry but i don't need a PR for update dependencies because i can do it myself when i want to make a new version

PR #2638, opened 2025-01-13
Fixed ru translation

mortefy:
"noRecommendKeepDefault" = "Не рекомендуется оставлять настройки по умолчанию"  The meaning here is distorted. The phrase translates to 'It is not recommended to keep the default settings,' which contradicts the original meaning. The correct translation is: -"Рекомендуется оставить настройки по умолчанию"
of course, if the original English phrase is translated correctly - 'It is recommended to keep the default'


mortefy:
I also noticed that some translations in other languages differ from the English version at the same point.

PR #2640, opened 2025-01-14
wireguard modal fix

Alexandrhhh:
There was a bug with listing all configs in the wireguard modal window, now under each peer there is only its config


MHSanaei:
thank you so much

NeoBeZ:
WG don't connect to mikrotik

PR #2644, opened 2025-01-15
Group Management of Subscription Clients

alirahimi818:
- Ability to add client to all inbounds
- Group editing
- Group deletion
- Group reset of usage
- Group enable/disable
- Synchronization of client traffics (every 10s)
- Ability to enable or disable the synchronizer


alirahimi818:
Many of the cases have been tested, but make sure to test again before merging and releasing.
I'll try to create a PR on Alireza's version as well, but it will be in a few weeks when I have some free time.

MHSanaei:
Thanks. I don't see any problems. Let's test one version; I hope everything works well.

PR #2652, opened 2025-01-31
Some Improvement

Incognito-Coder:
**Hi**
added switcher to Enable/Disable outbound traffic.
fixed `getSubGroupClients `function that cause ui hangs on set enable/disable clients.


MHSanaei:
thank you

PR #2657, opened 2025-02-06
Added REST api

KiselevAlexander:
- Added token based auth for API v2
- Added route for get server status


pomeo:
documentation in API.es.md and API.ru.md is mixed up

DevxRF:
> * Added route for get server status
my idea : 
really better in login response, return the server status on message because this better for get server status and update seasion cookies and btter security

KiselevAlexander:
> > * Added route for get server status
> >   my idea :
> >   really better in login response, return the server status on message because this better for get server status and update seasion cookies and btter security

I added token based auth and now there is no way to respond with server status because there is no authorization request.

lavrentijav:
Hello! 
There are no examples for creating inbounds of other protocols, as I understand them, they have completely different settings. Can the api create keys for wireguard itself?
Thank you in advance.

KiselevAlexander:
@MHSanaei  something wrong with this PL? 
Why it closed without comments?

PR #2658, opened 2025-02-06
Update themeSwitch.html

eldricks02:
This style prevented the text and icon of the ultra button from being displayed exactly relative to the theme switch from light to dark


TaraRostami:
> This style prevented the text and icon of the ultra button from being displayed exactly relative to the theme switch from light to dark

Hello
Can you share a screenshot of the problem you saw?

PR #2661, opened 2025-02-09
Make wget verify certificates

sausagenoods:
I don't know why `--no-check-certificate` was added to the wget commands in the first place. This option introduces a security vulnerability.

Without certificate validation in these update functions, users can be MITM'ed and served malicious binaries/scripts. There is no further integrity checking such as checksum comparisons later on too.

This PR makes wget verify certificates by removing the `--no-check-certificate` option.

### DIFF
--- a/x-ui.sh
+++ b/x-ui.sh
@@ -180,7 +180,7 @@ update_menu() {
         return 0
     fi
 
-    wget --no-check-certificate -O /usr/bin/x-ui https://raw.githubusercontent.com/MHSanaei/3x-ui/main/x-ui.sh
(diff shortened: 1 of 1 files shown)


ivansavra1:
> Hey @MHSanaei I’m curious on why you used that flag to begin with, what was the rationale for that? I can’t see what benefit would justify such an important security risk.

@vaxilu Care to comment on this?

vaxilu:
> > Hey @MHSanaei I’m curious on why you used that flag to begin with, what was the rationale for that? I can’t see what benefit would justify such an important security risk.
> 
> @vaxilu Care to comment on this?

有些新手的服务器根证书是过期的，或者服务器时间错误等各种原因，会导致证书验证失败，他们也不懂如何解决，索性直接跳过证书验证

PR #2668, opened 2025-02-24
Virtuozzo linux support

atarwn:
This may be the first time you've heard of VzLinux, but it's a great and lightweight distribution based on RHEL that I use on my servers. I just added `virtuozzo` to the `x-ui.sh` and `install.sh` scripts to ensure they work correctly on this system.


MHSanaei:
Ok thank you 😊

PR #2729, opened 2025-03-06
Change the cpu usage interval from second to minute

UltraMegaPotato:
Hi!

Now the Telegram bot sends a warning if CPU usage during the last second exceeds the threshold. This is quite useless because such small peaks can happen and do not affect system performance. However, what's more crucial is the CPU usage over a 1-minute period. It indicates that the system is busy for a significant amount of time, and you really should be warned.


MHSanaei:
hmm you are right, thank you

PR #2738, opened 2025-03-08
Tgbot menu commands

sergeev-ms:
The `/restart` command is not working properly because now it's `/restart force` and it can't be used as menu command. 
I have the suggestion to rename it to `/restart`. `force` just don't make any sense.
image


MHSanaei:
Hi
i remove the force here
https://github.com/MHSanaei/3x-ui/commit/a0dd101d9778f11dfe4866e7bc71c5119bcbf6fa
your code doesn't write well and you forgot to add isAdmin for restart 
i have to close this PR 
see you soon thanks

sergeev-ms:
@MHSanaei Hi. Thanks for you answer.
> you forgot to add isAdmin for restart

1. We create the menu on bot startup and we don't have information about isAdmin
2. isAdmin is checked in [answerCommand ](https://github.com/MHSanaei/3x-ui/blob/c35179d9240647884f592622d036bfad228f7bda/web/service/tgbot.go#L261)method and non-admin user will receive a message `tgbot.commands.unknown`
3. Now any user can send `/restart` or `/restart force` you can't prevent it, only answer if no authorized.

> your code doesn't write well
Сan you tell me that is not "write well" and I can improve it?

PR #2744, opened 2025-03-09
chore: add new grpc params for outbound

shishkevichd:


MHSanaei:
If you looking for a bug there is one you can fix
On the log page, the xray log doesn't show on the correct level 
You can see them on the debug level

shishkevichd:
> If you looking for a bug there is one you can fix 
>On the log page, the xray log doesn't show on the correct level 
>You can see them on the debug level

Check out #2746 and give a comment. This PR fixes the problem you're talking about, among other things.

PR #2772, opened 2025-03-14
chore: pretty 'Overview' page

shishkevichd:
<details><summary>Screenshot</summary>
<p>

<b>Light</b>

image

<b>Dark</b>

image

</p>
</details>


MHSanaei (commented):
I think the IPv6 doesn't fit; it should be smaller.
It would be better if we could see the Xray core version displayed like the panel version.
When Xray encounters an error, the status also shows an error, but it doesn’t look good when tested.
One more thing: they asked to replace 'Xray' and 'Manage' at the top with 'Uptime' and '3x-UI'

erfjab:
Beautiful and pleasant

TaraRostami (changes requested):
Well done.
Review the following changes:
https://github.com/MHSanaei/3x-ui/pull/2774

shishkevichd:
> All good, but I think it will be better with not just buttons, but with buttons and words.

> We also can't see current version of Xray-core running (it also corelates with first issue, I guess it should be near "wrench" icon)

Made labels for the buttons

> I think the IPv6 doesn't fit; it should be smaller.

> When Xray encounters an error, the status also shows an error, but it doesn’t look good when tested.

Fixed

TaraRostami:
There's no need to use `:sm="24" :lg="12`" on all the `Card`. Also, it's better to avoid use multiple CSS overrides.

MHSanaei (commented):
Well done! Everything looks good to me.

PR #2773, opened 2025-03-14
feat: custom subscription title in panel

am-monkey:
PR contains changes which are adding ability to define custom subscripion title from panel
- added input for custom title in "Panel settings --> Subscription"
- updated controllers in /sub
- subId replaced with subTitle string from panel
- added translations EN/RU

Снимок экрана 2025-03-14 в 19 28 10


lozhka13:
finally!

MHSanaei:
Thanks, but is it really necessary?

erfjab:
> Thanks, but is it really necessary?


For some users, yes.

MHSanaei:
Ok

Rzity:
Greetings, I also ask you to add a custom subscription header for each client individually, that is, add the Subtitle in the client configuration.

am-monkey:
It renders but not correct.
If We will add this line of code, will it fix?
Снимок экрана 2025-03-20 в 13 57 15
@MHSanaei

MHSanaei:
@am-monkey I don't know i didn't test it :D

PR #2865, opened 2025-04-04
Code refactoring

shishkevichd:
### What is a pull request?
This pull request is aimed at improving the code. Here are the changes made:
1. The class `DeviceUtils` has been converted to mixin `MediaQueryMixin`. Now the value of `isMobile` changes when the screen size changes.
2. The text modal now uses the `Blob` class to generate and load the file.
3. HTML templates have been restructured.
4. The `style` attribute has been rewritten in the templates. It was changed to use Vue inline style.
6. Added missing translations.


MHSanaei:
Hi @shishkevichd,  

Thank you for your work and the time you put into it.  
I don’t agree with your change to the index for system status. We don’t need a `div` for it. I used a `div` for the CPU because it included two more info.

shishkevichd:
> I don’t agree with your change to the index for system status

Hello, i reverted this change.

PR #2916, opened 2025-04-17
feat: Added a toggle to use public IPv4 in QR/URI

akilaid:
Added an option to use the server's public IPv4 when generating QR codes and URIs. The necessary translation updates for the new option have been included.
image

Why am I adding this feature/enhancement:
My ISP blocks DNS resolution when used in a certain (let’s say... nonstandard) way I had been relying on, so I had to switch to using the server’s IP address directly (have to manually type the IP in the client app). Also, when installing the panel on cloud providers like AWS, setting the listen IP(for inbounds) to the public IP doesn't work as expected, so this change improves compatibility in those scenarios.

### DIFF
--- a/web/html/modals/qrcode_modal.html
+++ b/web/html/modals/qrcode_modal.html
@@ -1,10 +1,9 @@
 {{define "modals/qrcodeModal"}}
 <a-modal id="qrcode-modal" v-model="qrModal.visible" :title="qrModal.title"
-    :dialog-style="isMobile ? { top: '18px' } : {}"
-    :closable="true"
-    :class="themeSwitcher.currentTheme"
-    :footer="null" width="fit-content">
+  :dialog-style="isMobile ? { top: '18px' } : {}" :closable="true" :class="themeSwitcher.currentTheme" :footer="null"
+  width="fit-content">
   <tr-qr-modal class="qr-modal">
+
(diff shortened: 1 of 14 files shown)


web/html/modals/qrcode_modal.html
```diff
@@ -25,6 +24,18 @@
...
       <tr-qr-box class="qr-box">
+        <div style="display: flex; align-items: center; margin-bottom: 8px;">
+          <span>{{ i18n "useIPv4ForHost" }}: </span>
+          <button
+            type="button"
+            role="switch"
+            :aria-checked="row.useIPv4"
+            :class="['ant-switch', 'ant-switch-small', { 'ant-switch-checked': row.useIPv4 }]"
+            @click="toggleIPv4(index)"
+            style="margin-left: 8px;">
```
sh1shd: Please use Vue inline style

web/html/modals/qrcode_modal.html
```diff
@@ -25,6 +24,18 @@
     </template>
     <template v-for="(row, index) in qrModal.qrcodes">
       <tr-qr-box class="qr-box">
+        <div style="display: flex; align-items: center; margin-bottom: 8px;">
```
sh1shd: Please use Vue inline style

PR #2966, opened 2025-05-02
feat(db)!: hash passwords with bcrypt in database

Columbiysky:
## What is the pull request?

Storing users passwords as a plain text isn't good idea generally. This PR contains changes to hash passwords with SHA256 in DB.
This PR is just an idea, if maintainers will accept it, I'll write better code for this.
I've tested:
 - First time start panel, when .db doesn't exist
 - Changing password for admin account
 - Updating .db to hash password (using runSeeders() function in db.go for this)
 - Of course, logging in panel after hashing password and changing in panel settings

- [ ] Frontend
- [x] Backend

## Type of Changes

- [ ] Bug fix
- [x] New feature
- [x] Refactoring
- [ ] Other

### DIFF
--- a/database/db.go
+++ b/database/db.go
@@ -129,16 +129,13 @@ func InitDB(dbPath string) error {
 	if err := initModels(); err != nil {
 		return err
 	}
-	empty, err := isTableEmpty("users")
-	if empty {
-		if err := initUser(); err != nil {
-			return err
-		}
 
+	isUsersEmpty, err := isTableEmpty("users")
+
+	if err := initUser(); err != nil {
+		return err
 	}
-	return runSeeders(empty)
-
-	// return nil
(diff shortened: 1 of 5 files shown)


shishkevichd:
Hello! Very nice change.

I think it would be better to use `bcrypt` instead of `SHA256` as it is better suited for password hashing.

MHSanaei:
We have a password generator for fresh installs. How do you want to display the password now?  And what happens to users if they want to downgrade to an older version?

shishkevichd:
> How do you want to display the password now?

We can create a variable (in bash) that will store the generated password and display it once.

> And what happens to users if they want to downgrade to an older version?

The password can be reset using the `x-ui` command.

PR #2968, opened 2025-05-03
chore: implement 2fa auth

shishkevichd:
## What is the pull request?

This pull request replaces the `secret token` with `2fa authentication`.

Two-factor authentication differs from secret token in that a new code is generated every 30 seconds, which cannot be selected using the brute force method. 2FA token can be added to any authentication application (Bitwarden, Google Authenticator, and so on.)

- [x] Frontend
- [X] Backend

## Type of Changes

- [ ] Bug fix
- [X] New feature
- [X] Refactoring
- [ ] Other

## Screenshots

image


MHSanaei:
it would be great if you add a qrcode for it

shishkevichd:
> it would be great if you add a qrcode for it

Hello!

I did it.

<details><summary>Screenshot</summary>
<p>

image

</p>
</details>

MHSanaei:
I think we need to add a few more features. For example, we should include an activation step to confirm the user has the correct app then removing the QR code. Also, when disabling two-step verification, we should require the user to enter a token for added security.

ckun52880:
Lack of backend verification: 
The code for "verifying the one-time verification code before saving the secret key" was not found, and there is a hidden danger that the timing of key saving is not controlled. If the front end is directly turned on without verification, the secret of the two-step verification may have been written but not properly verified, or it may not have been written at all, resulting in inconsistent enabled states. Incomplete logic: 
SettingsController does not provide a new interface for enabling/disabling two-step verification (such as /setTwoFactor), and the original updateUserSecret interface has been deleted, which means that the setting of the secret is only based on the front end without back-end confirmation, which is prone to unexpected behavior.

PR #2969, opened 2025-05-03
chore: do not show the current login and password

shishkevichd:
## What is the pull request?

This pull request removes the ability to view usernames and passwords in the x-ui panel managment

This was done for several reasons:
1. Before commit 85cbad3ef420ffdd7fec8657d247fdfe5e03903d, this could have led to a leak and unauthorized access to the panel
2. After commit 85cbad3ef420ffdd7fec8657d247fdfe5e03903d, it began to show a hashed password and mislead the user.

- [ ] Frontend
- [X] Backend

## Type of Changes

- [X] Bug fix
- [ ] New feature
- [X] Refactoring
- [ ] Other

### DIFF
--- a/main.go
+++ b/main.go
@@ -151,9 +151,7 @@ func showSetting(show bool) {
 			fmt.Println("get current user info failed, error info:", err)
 		}
 
-		username := userModel.Username
-		userpasswd := userModel.Password
-		if username == "" || userpasswd == "" {
+		if userModel.Username == "" || userModel.Password == "" {
 			fmt.Println("current username or password is empty")
 		}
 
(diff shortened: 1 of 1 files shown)


MHSanaei:
why did you do that?
we check user pass here
https://github.com/MHSanaei/3x-ui/blob/63edc63ab0c1b2ec679dde0292d5e557da8e350e/install.sh#L85

shishkevichd:
> we check user pass here

I fixed it in 13cf7271d6350a883a124d809b969321934abb7a. 

Now the check for default credentials is done by the `hasDefaultCredential` variable (returns true if the username is admin and the password hash matches the hashed password “admin”)

PR #2988, opened 2025-05-09
fix: issue 2953. Restore from .db file fails

Columbiysky:
## What is the pull request?

In "web\service\server.go" "ImportDB(file multipart.File)" removed defer os.Remove(tempPath) because later in method it will be renamed "err = os.Rename(tempPath, config.GetDBPath())"
Also removed err = database.InitDB(config.GetDBPath()) and error condition because earlier it already has been inited when it was tempPath.

- [ ] Frontend
- [ ] Backend

## Type of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Other

### DIFF
--- a/web/service/server.go
+++ b/web/service/server.go
@@ -524,9 +524,6 @@ func (s *ServerService) ImportDB(file multipart.File) error {
 	}
 	defer tempFile.Close()
 
-	// Remove temp file before returning
-	defer os.Remove(tempPath)
-
 	// Save uploaded file to temporary file
 	_, err = io.Copy(tempFile, file)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


web/service/server.go
```diff
@@ -524,9 +524,6 @@ func (s *ServerService) ImportDB(file multipart.File) error {
 	}
 	defer tempFile.Close()
 
-	// Remove temp file before returning
```
Columbiysky, line 527: Don't need because of line 562
err = os.Rename(tempPath, config.GetDBPath())

web/service/server.go
```diff
@@ -572,14 +569,6 @@ func (s *ServerService) ImportDB(file multipart.File) error {
 	}
 
 	// Migrate DB
-	err = database.InitDB(config.GetDBPath())
```
Columbiysky, line 575: It has been done already in line 534
err = database.InitDB(tempPath)

Columbiysky:
Have you any thoughts about the changes?

web/service/server.go
somebodywashere: We need to remove .temp file if we encounter an error and return before this function.

I added a defer function to solve this

web/service/server.go
```diff
@@ -524,9 +524,6 @@ func (s *ServerService) ImportDB(file multipart.File) error {
 	}
 	defer tempFile.Close()
```
somebodywashere:
```suggestion
	defer func() {
		_, err = os.Stat(tempPath)
		if err == nil {
			tempFile.Close()
			os.Remove(tempPath)
		}
	}()
```

web/service/server.go
somebodywashere: No need to remove this, we just double check all good after renaming a file

PR #3012, opened 2025-05-16
remove password from local logs and tgbot logs

### LINKED ISSUE #3010: Sensitive Information (Password) Logged on Failed Login Attempts
### Describe the bug

When a login attempt fails due to an incorrect username, password, or 2FA code, the application logs the entire login payload, including the plaintext password. This poses a serious security risk by potentially exposing sensitive user credentials in the application logs.
This information is not only stored in local logs but is also sent through the Telegram bot logs, which poses a significant security risk.

### How to repeat the problem?

Attempt to log in with an incorrect password or invalid 2FA code.
Check the application logs.
Observe that the log entry contains both the username and the password used in the attempt.

### Expected action

The application should log only non-sensitive metadata (e.g., username and IP address) when a login fails. Passwords or 2FA codes should never be included in logs.

### Received action

The logs contain entries like the following, which expose the user's password:
WARNING - wrong username: "admin", password: "[password]", IP: "[IP]"

### 3x-ui Version

2.6.0

### Xray-core Version

v25.5.16

### Checklist

- [x] This bug report is written entirely in English.
- [x] This bug report is new and no one has reported it before me.

samssh:
## What is the pull request?
this pull request fix #3010 

- [ ] Frontend
- [x] Backend

## Type of Changes

- [x] Bug fix
- [ ] New feature
- [x] Refactoring
- [ ] Other

### DIFF
--- a/web/controller/index.go
+++ b/web/controller/index.go
@@ -67,17 +67,16 @@ func (a *IndexController) login(c *gin.Context) {
 	user := a.userService.CheckUser(form.Username, form.Password, form.TwoFactorCode)
 	timeStr := time.Now().Format("2006-01-02 15:04:05")
 	safeUser := template.HTMLEscapeString(form.Username)
-	safePass := template.HTMLEscapeString(form.Password)
 
 	if user == nil {
-		logger.Warningf("wrong username: \"%s\", password: \"%s\", IP: \"%s\"", safeUser, safePass, getRemoteIp(c))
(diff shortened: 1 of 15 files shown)


MHSanaei:
Thanks for your report and PR.
I don’t think this is a bug or security issue. I added that log myself to track failed login attempts and see what passwords are being used.
only incorrect usernames and passwords will send to the panel log and the admin Telegram bot, and it’s already been discussed before. Since only you have access to both the panel and the Telegram bot, there’s no real risk, so no need to worry.

PR #3205, opened 2025-07-07
Added list of services for get public IP address (IP v4 and v6)

CHERTS:
## What is the pull request?

Added list of services for get public IP address (IP v4 and v6)

- [ ] Frontend
- [x] Backend

## Type of Changes

- [ ] Bug fix
- [ ] New feature
- [x] Refactoring
- [ ] Other

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -7,6 +7,7 @@ yellow='\033[0;33m'
 plain='\033[0m'
 
 cur_dir=$(pwd)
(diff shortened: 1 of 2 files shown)


MHSanaei:
Your links don’t separate IPv4 and IPv6, which causes some errors.

CHERTS:
You could highlight the problem area and I would fix it. That's the whole point of team development. But I see you don't need it.

PR #3269, opened 2025-07-24
feat: Add multi-server support for Sanai panel

javadtgh:
This commit introduces a multi-server architecture to the Sanai panel, allowing you to manage clients across multiple servers from a central panel.

Key changes include:

- **Database Schema:** Added a `servers` table to store information about slave servers.
- **Server Management:** Implemented a new service and controller (`MultiServerService` and `MultiServerController`) for CRUD operations on servers.
- **Web UI:** Created a new web page for managing servers, accessible from the sidebar.
- **Client Synchronization:** Modified the `InboundService` to synchronize client additions, updates, and deletions across all active slave servers via a REST API.
- **API Security:** Added an API key authentication middleware to secure the communication between the master and slave panels.
- **Multi-Server Subscriptions:** Updated the subscription service to generate links that include configurations for all active servers.
- **Installation Script:** Modified the `install.sh` script to generate a random API key during installation.

**Known Issues:**

- The integration test for client synchronization (`TestInboundServiceSync`) is currently failing. It seems that the API request to the mock slave server is not being sent correctly or the API key is not being included in the request header. Further investigation is needed to resolve this issue.

- [ ] Frontend
- [ ] Backend

## Type of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Other

### DIFF
--- a/database/db.go
+++ b/database/db.go
@@ -35,6 +35,7 @@ func initModels() error {
 		&model.InboundClientIps{},
 		&xray.ClientTraffic{},
 		&model.HistoryOfSeeders{},
+		&model.Server{},
 	}
 	for _, model := range models {
 		if err := db.AutoMigrate(model); err != nil {
--- a/database/model/model.go
+++ b/database/model/model.go
@@ -105,3 +105,12 @@ type Client struct {
 	Comment    string `json:"comment" form:"comment"`
 	Reset      int    `json:"reset" form:"reset"`
 }
+
+type Server struct {
+	Id      int    `json:"id" gorm:"primaryKey;autoIncrement"`
+	Name    string `json:"name" gorm:"unique;not null"`
+	Address string `json:"address" gorm:"not null"`
+	Port    int    `json:"port" gorm:"not null"`
+	APIKey  string `json:"apiKey" gorm:"not null"`
+	Enable  bool   `json:"enable" gorm:"default:true"`
+}
--- a/go.mod
+++ b/go.mod
@@ -15,6 +15,7 @@ require (
 	github.com/pelletier/go-toml/v2 v2.2.4
 	github.com/robfig/cron/v3 v3.0.1
 	github.com/shirou/gopsutil/v4 v4.25.6
+	github.com/stretchr/testify v1.10.0
 	github.com/valyala/fasthttp v1.63.0
 	github.com/xlzd/gotp v0.1.0
 	github.com/xtls/xray-core v1.250608.0
@@ -32,6 +33,7 @@ require (
 	github.com/bytedance/sonic/loader v0.2.4 // indirect
 	github.com/cloudflare/circl v1.6.1 // indirect
 	github.com/cloudwego/base64x v0.1.5 // indirect
+	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/dgryski/go-metro v0.0.0-20250106013310-edb8663e5e33 // indirect
 	github.com/ebitengine/purego v0.8.4 // indirect
 	github.com/fasthttp/router v1.5.4 // indirect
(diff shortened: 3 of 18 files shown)


alireza0:
Thank you for contributing.
Everything is nice so far except our main problem: Sync client/inbound traffics.

For this important problem we should use "sync-sofar" solution to make sure that every usages in all slave servers will count.
Personal prefered solution is to clear all usages from slaves after sync. During traffic sync, all slaves traffics should be added to the master, then evaluate traffic exhaustion. If a client/inbound is exhausted, then disable it on all master/slave.

Other concerns:
Slaves:
1. Disable web interface.
2. Limit api communication only by master source
3. Sync ip limit
4. Sync certificate files
5. Consider Listen IP of each inbound (to be ignored or updatable via master)
6. Consider MultiDomain (to be empty or reserve one config per server)
7. Consider Outbounds specially for tunnels.
8. Consider Reverse Proxy

dev30na (approved):

liamnees:
To make it multi-server, just add traffic equalization and that's it. The problem with the previous scripts was that in addition to equalizing traffic, they also managed time and clients. Please add a script that has no capabilities other than equalizing traffic and that also does this equalization every 2 minutes. I don't think there will be any problems this way.

fusboy:
Сейчас сборка подписок json осуществляем так, может понадобится. на всякий случай оставлю это тут
[parcer.php](https://github.com/user-attachments/files/21891465/parcer.php)

liamnees:
@aliglzr

liamnees:
> To make it multi-server, just add traffic equalization and that's it. The problem with the previous scripts was that in addition to equalizing traffic, they also managed time and clients. Please add a script that has no capabilities other than equalizing traffic and that also does this equalization every 2 minutes. I don't think there will be any problems this way.

@aliglzr

dimasmir03:
@alireza0 @javadtgh can you tell me if I can join and how can I join the development of this feature and speed up its release?

Wasdalt:
Can you tell me where I can write about the API bug?

MHSanaei:
@Wasdalt direct telegram channel
https://t.me/XrayUI

Dreyk-Zer0:
@javadtgh I ask you to consider the scenario where the multi -server panel will be able to work with Proxy Chain settings, where the inbound of one server serves the outbound another. I use such settings in my installations, but now everything has to be steamed by a separate panels.

dimasmir03:
> where the inbound of one server serves the outbound another

1 panel -> 2 panel -> 3 panel chain?

Dreyk-Zer0:
> > where the inbound of one server serves the outbound another
> 
> 1 panel -> 2 panel -> 3 panel chain?

Yes, but i use 2 step chains.

1 panel -> 2 panel
3 panel -> 4 panel

I need to use inbounds via Wireguard in some infrastructures, but it does not go across the country border. It is also convenient for me to have entry points inside the country and the out nodes are changed as it is conveniently depending on the changing situation.

zhaoxueqingxiuyuansoftware:
> @javadtgh @alireza0 @dimasmir03 @MHSanaei
> 
> Hi everyone,
> 
> I wanted to share my thoughts on the multi-server approach. From my experience, full server management or separate slave panels aren’t strictly necessary for scaling. In my setup, I have two panels running on different servers but using a centralized MySQL database. A simple TCP proxy handles requests in a round-robin manner, and it works reliably while keeping the setup simple.
> 
> I would suggest migrating the panel to PostgreSQL first. If your servers are configured identically, there’s no need for an extra “server management” layer. A centralized database plus a proxy load balancer is sufficient. This approach is simpler and already resembles a node-based architecture like n8n.
> 
> Thanks for your work! 😊
> 
> image

I agree with your idea and fully support you.

PR #3340, opened 2025-08-09
Moved DB to local app data folder on Windows

fgsfds:
## What is the pull request?

x-ui.db has been moved to local_app_data/x-ui where it belongs on Windows instead of whatever_drive_binaries_are/etc

- [ ] Frontend
- [x] Backend

## Type of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Refactoring
- [x] Other

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -68,14 +68,14 @@ func GetDBFolderPath() string {
 	defaultFolder := "/etc/x-ui"
 
 	if runtime.GOOS == "windows" {
-		homeDir, err := os.UserHomeDir()
-		if err != nil {
-			logger.Error("Error while getting user folder: %w", err)
+		homeDir := os.Getenv("LOCALAPPDATA")
+		if homeDir == "" {
+			logger.Error("Error while getting local app data folder")
 			return defaultFolder
 		}
 
 		userFolder := filepath.Join(homeDir, "x-ui")
-		err = moveExistingDb(defaultFolder, userFolder)
+		err := moveExistingDb(defaultFolder, userFolder)
 		if err != nil {
 			logger.Error("Error while moving existing DB: %w", err)
 			return defaultFolder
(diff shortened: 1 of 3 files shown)


MK0ltra:
The DB path hasn’t changed for years, and lots of scripts, such as backup services and Telegram bots, expect it to be at /etc/x-ui/x-ui.db

MK0ltra:
> > The DB path hasn’t changed for years, and lots of scripts, such as backup services and Telegram bots, expect it to be at /etc/x-ui/x-ui.db
> 
> Shouldn't be a problem. Existing db will be moved to the new location on first start. Tg bots / backup services get db location from GetDBPath() which now returns new path.

They mostly use this path as a hard-coded value. Just search for /etc/x-ui/x-ui.db on GitHub and you’ll find many repositories. This change will break all of these on windows:

https://github.com/search?q=%2Fetc%2Fx-ui%2Fx-ui.db&type=code&p=1

fgsfds:
> They mostly use this path as a hard-coded value. Just search for /etc/x-ui/x-ui.db on GitHub and you’ll find many repositories. This change will break all of these on windows:

Ah, you mean 3rd party tools. I'm afraid, these won't properly work on windows anyway, since "/etc/x-ui/x-ui.db" could mean "C://etc/x-ui/x-ui.db", "D://etc/x-ui/x-ui.db", "Z://etc/x-ui/x-ui.db" etc., depending on where 3x-ui bins are located.

If anything, moving db to appdata will give these tools a predictable and constant path to look for db at.

MHSanaei:
unfortunately it doesn't work as you said 
my exe file was on `D:\x-ui` but the database create on `C:\Users\xxx\AppData\Local\x-ui`

MHSanaei:
check this one

PR #3381, opened 2025-08-25
AutoFill TOTP

finagin:
## What is the pull request?

Changing the value of the HTML autocomplete attribute of the TOTP field to support Apple Keychain

https://developer.apple.com/documentation/security/enabling-password-autofill-on-an-html-input-element

- [x] Frontend
- [ ] Backend

## Type of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Other


MHSanaei:
thanks

PR #3397, opened 2025-08-30
x-ui.service hardneing

ghost:
Hardening for systemd service,slightly improves security.

Tested on Ubuntu 24.04.3 LTS


Darkcyankitty:
> Thank you, i hope you test it enough

It works stably on my server, but it's worth noting that these changes won't affect security much, I've only made superficial changes that will slightly limit 3x-ui from what it doesn't need, for better security it's worth developing stricter settings, maybe I'll do that later.

Issue #3127, opened 2025-06-20
I would like to ask if this panel has the function of automatically resetting the traffic every month?


Asahi666:
### Is your feature request related to a problem?

As title

### Describe the solution you'd like

I wish there was a monthly reset flow option when adding an inbound, that would automatically reset the flow on a certain date each month.

### Checklist

- [x] This feature report is written entirely in English.

to889:
This function is very useful. x-ui has had this function for many years. I hope 3x-ui can learn from it and reproduce it.

CraftNobody0505:
Also need

Yuis1:
Also need

PR #3407, opened 2025-08-31
feat: Simple periodic traffic reset (for Inbounds) – daily | weekly | monthly 

egregors:
## What is the pull request?

This PR adds a possibility to set periodic traffic reset option in order to reset traffic counter every day | week | month.
The implementation is pretty straightforward, but in general able to be extended for more periods variants.
Traffic reset right now supports only for Inbounds.

My motivation is my own use cases and [this issue](https://github.com/MHSanaei/3x-ui/issues/3127) 

What I have done: 

- add new field into Inbounds model + index
- add new field into UI add | edit Inbounds (check screenshot out), by default – `never`
- add job (every day | week | month) to check if we need to reset traffic counter
- localizations

- [x] Frontend
- [x] Backend

## Type of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Other

## Screenshots

grafik


egregors:
I believe I found better way to do it. Instead of fetch all Unbound each 10 minutes, we could run 3 separated cron jobs:
- every 24h get all Inbounds with `prediodicTrafficReset == "daily"`
- every week get all Inbounds with `prediodicTrafficReset == "weekly"`
- every month get all Inbounds with `prediodicTrafficReset == "monthly"`

and reset it. 

It looks like this approach produces less load to the service. 

@MHSanaei what do u think?

aliglzr:
@egregors Thanks for working on this! This feature will be really useful.
In my opinion, a few improvements could make it stronger:

Scheduler approach: Instead of checking every 10 minutes, I think the separate cron jobs (daily/weekly/monthly) would be more efficient and reduce unnecessary load.

Testing & reliability: Could you add some unit and integration tests to cover daily/weekly/monthly resets, plus edge cases (e.g., timezone, month boundaries)?

Logging & observability: It would help if the reset job logs key events (e.g., which inbound was reset and when) so we can debug if something goes wrong.

UI/UX: It’d be great to show the last reset time or at least make sure the options are clearly labeled and localized.

Commits – Once your PR is ready, please squash or rebase into a few cleaner commits (e.g., model/UI, scheduler, tests/docs) for easier review.

Overall, this is a really good feature, thanks for putting it together!

MHSanaei:
Okay, thank you.

egregors:
Guys, let's cut a scope a bit. It already took a lot of time, so I'd prefer to move iteratively, if you don't mind. So, in this particular PR, I added periodic traffic reset only for inbound. It also shows last reset time in the info tooltip (see screenshots) in the inbound edit modal (only when last_reset_time > 0). 

grafik

I'll add client periodic traffic reset eventually, a bit later.

lilarcor:
> > I understand this topic was closed. Is the clients based traffic reset added into latest version?
> 
> @lilarcor Nop, I did it only for the whole inbound. Currently, I have no plan to add client-based auto reset, sorry. Lack of time 😞

Thanks anyway. I will try to figure it out with calling API.

xiaoxiyao:
After the inbound reset, will the clients under this inbound be reset automatically?

PR #3475, opened 2025-09-15
Improved xray logs display handling

fgsfds:
## What is the pull request?

Improved xray logs parsing

Added support for email field

Displaying logs as a table

- [x] Frontend
- [x] Backend

## Type of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Refactoring
- [x] Other

## Screenshots

image


MHSanaei:
everything looks good except when we download the log

fgsfds:
> everything looks good except when we download the log

Completely forgot about it. Thanks for the fix!

PR #3509, opened 2025-09-18
Correction for cpu history

Drahonn:
Incorrect time designation in CPU history


MHSanaei:
is not incorrect 
2s or 30s ... for between dot point 
we have 60 dot 
that's why you think its 2m - 30m ...

PR #3578, opened 2025-10-26
Feature: add setting certs for subscription while generating for panel

dimasmir03:
## What is the pull request?

_add while generating certs for panel setting this certs for subscription also_

- [ ] Frontend
- [x] Backend

## Type of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Other

### DIFF
--- a/main.go
+++ b/main.go
@@ -321,6 +321,20 @@ func updateCert(publicKey string, privateKey string) {
 		} else {
 			fmt.Println("set certificate private key success")
 		}
+
+		err = settingService.SetSubCertFile(publicKey)
+		if err != nil {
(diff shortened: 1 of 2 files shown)


AerEbi:
I've been using the old version (1.6.1) for a while now. To use this version without any problems, I need to downgrade the xray kernel version to use it without any problems. The required version is 1.7.5. But for a few days now, when I click on the xray kernel in the old panel, it gives the following error:

" Get versionFail: json: cannot unmarshal object into Go value of type []service.Release "

Please look into this and re-enable versioning in older panel versions.

PR #3600, opened 2025-12-05
Set variables for packaging purposes

nebulosa2007:
## What is the pull request?

Added variables for packaging purposes:

1. If XUI_MAIN_FOLDER variable is not set: xui_folder = /usr/local/x-ui. For example, for Arch Linux it will be set to /usr/lib/x-ui 
2. If XUI_SERVICE variable is not set: xui_service = /etc/systemd/system. For example, for Arch Linux it will be set to /usr/lib/systemd/system

- [ ] Frontend
- [X] Backend

## Type of Changes

- [ ] Bug fix
- [ ] New feature
- [X] Refactoring
- [ ] Other

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -672,7 +672,7 @@ install_x-ui() {
         rc-service x-ui start
     else
         if [ -f "x-ui.service" ]; then
-            cp -f x-ui.service /etc/systemd/system/
+            cp -f x-ui.service ${xui_service}/
         else
             case "${release}" in
                 ubuntu | debian | armbian)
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -451,11 +454,11 @@ prompt_and_setup_ssl() {
 }
 
 config_after_install() {
-    local existing_hasDefaultCredential=$(/usr/local/x-ui/x-ui setting -show true | grep -Eo 'hasDefaultCredential: .+' | awk '{print $2}')
-    local existing_webBasePath=$(/usr/local/x-ui/x-ui setting -show true | grep -Eo 'webBasePath: .+' | awk '{print $2}' | sed 's#^/##')
```
MHSanaei, line 455: undo 
' | sed 's#^/##

PR #3610, opened 2025-12-27
vless: use Inbound Listen address in Subscription service

gowa:
## What is the pull request?
When Subscription service is enabled and available, say, via https://example.com:26334/ with all TLS/certificates configured,
but Vless Inbound is configured using IP Address, say, 9.9.9.9, then, currently, the Vless configuration link, generated by subscription service, differs from that generated for manual distribution, particularly, in the end point address:

subscription service generated link:
vless://04497175-5876-41cc-9299-80e47839e675@example.com:46252....

a link for manual distribution:
vless://04497175-5876-41cc-9299-80e47839e675@9.9.9.9:46252....

The changes suggested eliminate this difference in favor of using Inbound Listen address (similar to the way of generating a link for manual distribution): the address where clients need to connect by their VPN client may differ from the address where the subscription service is accessible.

- [ ] Frontend
- [X] Backend

## Type of Changes

- [X] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Other

### DIFF
--- a/sub/subService.go
+++ b/sub/subService.go
@@ -179,9 +179,15 @@ func (s *SubService) genVmessLink(inbound *model.Inbound, email string) string {
 	if inbound.Protocol != model.VMESS {
 		return ""
 	}
+	var address string
+	if inbound.Listen == "" || inbound.Listen == "0.0.0.0" || inbound.Listen == "::" || inbound.Listen == "::0" {
+		address = s.address
+	} else {
+		address = inbound.Listen
+	}
 	obj := map[string]any{
 		"v":    "2",
-		"add":  s.address,
+		"add":  address,
 		"port": inbound.Port,
 		"type": "none",
 	}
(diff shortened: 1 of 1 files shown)


wah2233:
I'm also using a subscription and already have an SSL certificate set up. I'm using VLESS inbound, and everything works perfectly fine for me.

gowa:
@Gothik99 , ok, I think I need to clarify my set up.
1. IP1 and Domain1 are bound (home router). I cannot alter this. Domain name does not belong to me (dyndns alike thing). But for this domain I have a TLS public-private key pair.
2. IP2 - just some other VPS through which I need my vpn work.

I set NAT forward from IP1 to IP2 for subscription service (so that certificate works fine).
But I don't want to do the same for actual VPN ttaffic... there is no point in routing it via that router from one country to another.

serving subscription service requests - I'm ok to NAT/route, but for VPN itself - not ok.

PR #3681, opened 2026-01-18
Refactoring + geodat files auto update

mixa2130:
### Why is this needed?

We live in challenging times, as many governments attempt to restrict internet freedom and monitor users’ online activity. For example, in Russia, Roskomnadzor blocks or throttles around 20 services every day. Xray is a solid protocol, but it is not completely immune to detection. One of the most effective ways to protect a VPN server is to minimize outgoing traffic—something 3xui already does very well.

However, as mentioned earlier, the list of blocked resources changes extremely quickly, and manually updating geodat files through the panel has become quite inconvenient. This pull request addresses that issue by introducing automated updates.

### What this PR includes:
1. Cron container — downloads geodat files on a configurable schedule (via docker env), storing them in a shared volume.
2. Docker proxy — allows the cron container to restart the main 3xui container so the new geodat files are applied automatically.
3. Some additional refactoring that I believe will be useful.

P.S. The main 3xui container can run independently of geodata-cron and docker-proxy. If a user doesn’t need them, they can simply remove these containers.

### What I refactored?

* Split x-ui.sh into a set of Bash scripts located in the lib/ directory. This allowed me to reuse its functions in the main Dockerfile (which is why DockerInit.sh is no longer needed) and in Dockerfile-cron-runner.
* Reduced Docker image size: 349.93 MB → 297.3 MB
* Optimized Docker build time by ~1.5× on a VPS with 1 CPU core and 2 GB of RAM
* Moved README translations from the root directory to `docs/`, making the root directory cleaner
* Fixed several Bash linter issues

### How I tested?

Local Docker tests on macOS, as well as testing on Ubuntu and Debian VPS instances with 1 CPU core and 2 GB of RAM.
An amd64 test release for manual installation is available at https://github.com/mixa2130/3x-ui/releases/tag/v0.0.1

- [ ] Frontend
- [x] Backend

## Type of Changes

- [ ] Bug fix
- [x] New feature
- [x] Refactoring
- [ ] Other

## Screenshots

Screenshot 2026-01-18 at 22 39 08

## What would I like to update in 3xui?

* I would like to add support for custom geodat files.
* I would like to have an official Docker image on Docker Hub to save time and VPS resources.
* Built-in Nginx support / examples

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -44,7 +44,7 @@ services:
       TZ: "UTC"
       DOCKER_PROXY_URL: "http://docker-proxy:2375"
       TARGET_CONTAINER_NAME: "3xui_app"
-      CRON_SCHEDULE: "*/5 * * * *" #"0 */6 * * *"
+      CRON_SCHEDULE: "0 */6 * * *"
       SHARED_VOLUME_PATH: "/app/bin"
     volumes:
       - $PWD/geodata/:/app/bin/


mixa2130:
@MHSanaei take a look please 🙂‍↕️

PR #3748, opened 2026-02-04
Add x-ui-updategeo.timer and .service files

nebulosa2007:
## What is the pull request?
- Refactoring/fix: removed some obsolete code — checking for the existing of the `x-ui.service` file — will never be true [even for legacy versions](https://github.com/nebulosa2007/3x-ui/blob/d8fb09faaeeebe34b57f1c26aa1ab8b529327989/x-ui.sh#L169).
- Refactoring: for the Alpine `x-ui.rc` file, loading starts right away with ipv4, like for other service files.
- New: Added timer and systemd service files for loading the `dat` files:
```
 [Timer]
AccuracySec=5min              
# For delay activation by up to 5 minutes to batch timer executions efficiently.
OnBootSec=5min                
# Runs the timer 5 minutes after system boot
OnCalendar=*-*-* 02,14:00:00 UTC  
# Schedules the timer to trigger every day at 02:00 and 14:00 UTC. (see below)
FixedRandomDelay=true         
# Applies the randomized delay deterministically.
RandomizedDelaySec=10min      
# Adds a random delay of up to 10 minutes to the scheduled activation time (useful to avoid simultaneous triggers on multiple hosts).
Persistent=true               
# Ensures missed runs (e.g., if the system was off) are triggered immediately after boot.

Unit=x-ui-updategeo.service   
# Specifies which service unit to start when the timer fires.
```

It is set to run service at 02:00 and 14:00 UTC:
1. Loyalsoldier (geoip.dat, geosite.dat) - [updates at 10PM UTC ](https://github.com/Loyalsoldier/v2ray-rules-dat/blob/b74d235f909125256d8b6d3bfbb8cf55872b8f10/.github/workflows/run.yml#L5)
2. chocolate4u (geoip_IR.dat, geosite_IR.dat) - [updates at 04:40 UTC](https://github.com/Chocolate4U/Iran-v2ray-rules/blob/676695ea3b4c95d5cf48a7c4e2e718bac5b8a099/.github/workflows/release.yml#L12)
3. runetfreedom (geoip_RU.dat, geosite_RU.dat) -  [updates at 03:30, 09:30, 15:30, and 21:30 UTC](https://github.com/runetfreedom/russia-v2ray-rules-dat/blob/bc6ee7f63025745b338d9b2391f2f4615b13cfb5/.github/workflows/build.yml#L5C5-L5C33)

This is a compromise to receive updates twice a day.

After installing the files, only a message on how to activate the timer will be displayed.
It works for version 2.8.8 and later. I have been testing this timer for a couple of weeks already.

- [ ] Frontend
- [x] Backend

## Type of Changes

- [ ] Bug fix
- [x] New feature
- [x] Refactoring
- [ ] Other


nebulosa2007:
@MHSanaei FYI https://xtls.github.io/en/config/geodata.html

Issue #3623, opened 2026-01-03
2.8.6 update forces to enable SSL


wildrun0:
### Describe the bug

I'm using 3x-ui behind the nginx server where is SSL cert is already issued but updating from 2.8.5 forces SSL option for internal web server

<img width="937" height="193" alt="Image" src="https://github.com/user-attachments/assets/943cad26-a0e8-4ded-acc6-06e59dbf297d" />

### How to repeat the problem?

1. Try to update 2.8.5 -> 2.8.6
2. Try to not issue a SSL cert when prompted

### 3x-ui Version

2.8.6

### Checklist

- [x] This bug report is written entirely in English.
- [x] This bug report is new and no one has reported it before me.

wildrun0:
weird

Anderhar:
> 我们的标准是监听 0.0.0.0/:: 等公网 IP 段时需要可信的 TLS 证书，对于 127.0.0.1 等内网 IP 段则无需强制 TLS，毕竟还有 SSH 端口转发

In fact, the way this requirement was implemented ended up to worse for 3X-UI. 
In my case, the built-n acme.sh never works as expected, and now just doesn't allow to complete the installation without any errors (by the way, 3X-UI runs acme.sh as root, which is [not intended](https://github.com/acmesh-official/acme.sh/wiki/sudo) by its developers). 
It's very disappointing if you've always used the panel on 127.0.0.1 without a subscription service. Seems like someone just added unwanted problems for me with no right to avoid them.  
Don't be like Apple, please.

**Suggested solution:** provide at least an environment variable for the installer to skip this SSL step. (However, just asking me directly would have been better.)

PR #3754, opened 2026-02-06
add external SSL ( e.g., Nginx reverse proxy)

appotry:
fix: https://github.com/MHSanaei/3x-ui/issues/3623


MHSanaei:
Hi 
Thanks for your PR but you just made a button to skip the SSL 
we don't want to do that

PR #3814, opened 2026-02-16
Add configurable host overrides for exported client links

MrTeeett:
## What is the pull request?
This pull request introduces configurable host overrides for exported client links and JSON subscriptions.

Previously, the exported client links derived the host from inbound.listen or the HTTP request host header. In setups using wildcard listen addresses (0.0.0.0 / ::), reverse proxies, CDNs, or NAT, this could result in incorrect or inconsistent host values in generated links.

This change adds a deterministic host resolution mechanism with the following priority:

1) client.subHost (per-client override)
2) inbound.settings.subHost (per-inbound override)
3) Global default host (SubDefaultHost)
4) inbound.listen (if not wildcard)
5) Request host header (fallback)

All link generators (vmess, vless, trojan, shadowsocks) and JSON subscription exports now use this unified resolution logic.

The new field is optional and backward compatible.

- [ ] Frontend
- [x] Backend

## Type of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Other

## Screenshots

изображение


ckun52880:
Hi, thanks for the great work on this PR — the `subHost` override feature looks very useful.

I have a deployment scenario where Xray is placed behind a reverse proxy, and I think some additional optional overrides might also be helpful.

For example, my actual inbound configuration is:

* Address: `127.0.0.1`
* Port: `10001`
* TLS: disabled

because Nginx handles the external TLS and forwards traffic internally.

But the real client connection parameters should be:

* Domain: `example.com`
* Port: `443`
* TLS: enabled

In this case, overriding only the host may not fully solve the exported link issue, because the generated subscription may still contain the internal port or `security=none`.

Would it be possible to consider adding optional overrides for exported:

* port
* TLS/security type

For example:

```json id="q8m1zp"
{
  "subHost": "example.com",
  "subPort": 443,
  "subTLS": "tls"
}
```

I think this could be useful for users deploying behind:

* Nginx/Caddy reverse proxy
* Docker internal networking
* CDN / TLS termination
* Reality fallback / stream proxy setups

Just a suggestion for consideration, and thank you again for your work on this project.

PR #3877, opened 2026-03-02
Dns changer

d3l1f3r:
## What is the pull request?

Added the ability to change the DNS resolver for the server 

- [ ] Frontend
- [x] Backend

## Type of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Other

### DIFF
--- a/x-ui.sh
+++ b/x-ui.sh
@@ -1671,234 +1671,11 @@ run_speedtest() {
     speedtest
 }
 
-run_librespeed() {
-    echo -e "${yellow}What do you want to do?\n1) Use\n2) Install/Update\n3) Delete"
-    read -p ": " action
-
-    if [ "$action" = 1 ]; then
-        if command -v librespeed-cli &>/dev/null; then
-            librespeed-cli
-        else
-            echo -e "${red}LibreSpeed is not installed!"
-        fi
-    
-    elif [ "$action" = 2 ]; then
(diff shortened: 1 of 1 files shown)


x-ui.sh
```diff
@@ -1678,6 +1678,36 @@ ip_validation() {
     ipv4_regex="^((25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9][0-9]?|0)\.){3}(25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9][0-9]?|0)$"
 }
 
+change_dns() {
+    echo -e "${yellow}Changing DNS resolver"
+    echo -e "${plain}Enter resolver (default: quad9): "
+    read resolver
+
+    IPv6_on=$(cat /proc/sys/net/ipv6/conf/all/disable_ipv6)
```
cherts: IPv6 can be enabled, but it won't be used.
Multiple checks must be used, and each must be positive.
The second check must be:
`ip -6 addr show scope global`
If the server interface has an IPv6 address, the command will return a set of data. If the output is empty, then there is no IPv6 address, effectively indicating that the server isn't using IPv6.
Next, if the` ip -6` check is successful, we should try to retrieve data via the IPv6 protocol.
For example, `curl -6 https://ifconfig.co`
Only if all three steps are positive can we conclude that the server is using IPv6.

PR #3887, opened 2026-03-05
perf: replace /dev/urandom | tr with openssl rand to fix CPU spike in gen_random_string

### LINKED ISSUE #3886: perf: gen_random_string causes high CPU usage due to 75% byte rejection in tr

During installation, gen_random_string pipes /dev/urandom directly into 
tr -dc 'a-zA-Z0-9'. Since only ~24% of random bytes pass the filter, 
tr discards ~75% of all input and busy-loops at high CPU until the pipe closes.

This causes a noticeable CPU spike on every fresh install or credential regeneration, especially noticeable on low-spec VPS ( less than 1 vCPU  / 512MB RAM), where tr can saturate the single core for several seconds during installation. An example is on Google Cloud Platform's e2-Micro Instance.

Replace the current /dev/urandom | tr pipeline with openssl rand -base64, 
which outputs A-Za-z0-9+/= only, reducing tr's rejection rate from ~75% to ~2%:
```
gen_random_string() {
    local length="$1"
    openssl rand -base64 $(( length * 2 )) \
        | tr -dc 'a-zA-Z0-9' \
        | head -c "$length"
}
```
> Update: after reviewing the source, openssl is not included in install_base for any distro. The PR therefore also adds openssl to all package manager install lists as a new explicit dependency.

### Describe alternatives you've considered

Here is a python alternative, cleanest but requires python pre-installed:
```
gen_random_string() {
    python3 -c "
import secrets, string
chars = string.ascii_letters + string.digits
print(''.join(secrets.choice(chars) for _ in range($1)), end='')
"
}
```
No CPU idle time issues, with guaranteed length precision. Recommended for use when scripts can rely on Python. 

I didn't apply this method because this is a shell installation script and the goal is to keep it self-contained. Adding Python3 as a dependency would pull in a full runtime or a single use case, which is disproportionate for a lightweight install script.

So, OpenSSL is the pragmatic choice: one explicit new dependency across all distros, but a minimal and justified one — and far safer than relying on Python being present. :)

### Checklist

- [x] This feature report is written entirely in English.

Kookiejarz:
closes #3886

Replace /dev/urandom | tr pipeline with openssl rand -base64 to reduce 
tr's byte rejection rate from ~75% to ~2%, fixing CPU spikes on low-spec VPS.

See issue #3886 for detailed analysis.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -76,37 +76,38 @@ is_port_in_use() {
 install_base() {
     case "${release}" in
         ubuntu | debian | armbian)
-            apt-get update && apt-get install -y -q cron curl tar tzdata socat ca-certificates
+            apt-get update && apt-get install -y -q cron curl tar tzdata socat ca-certificates openssl
         ;;
         fedora | amzn | virtuozzo | rhel | almalinux | rocky | ol)
-            dnf -y update && dnf install -y -q curl tar tzdata socat ca-certificates
+            dnf -y update && dnf install -y -q curl tar tzdata socat ca-certificates openssl
         ;;
         centos)
             if [[ "${VERSION_ID}" =~ ^7 ]]; then
-                yum -y update && yum install -y curl tar tzdata socat ca-certificates
+                yum -y update && yum install -y curl tar tzdata socat ca-certificates openssl
             else
-                dnf -y update && dnf install -y -q curl tar tzdata socat ca-certificates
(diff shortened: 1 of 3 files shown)


MHSanaei:
Thanks for your PR 
we have two more gen_random_string on x-ui.sh and update.sh

PR #4065, opened 2026-04-21
Fix xhttp xPadding settings missing from generated links (panel + subs)

pwnnex:
## Summary

Admins configuring an xhttp inbound with a non-default `xPaddingBytes`, or enabling obfs mode with a custom `xPaddingKey` / `xPaddingHeader` / `xPaddingPlacement` / `xPaddingMethod`, end up with working **server** settings but a **client link** that doesn't mention any of them. The client keeps xray / sing-box's internal defaults, every request it sends fails the server's padding validation and the inbound log fills with

```
invalid padding (queryInHeader=Referer, key=x_padding) length: 0
```

With the custom-key case, xray ends up looking for a header/cookie that the client never sets, so the key-side of the error looks like `key=""` to the admin. The client-visible symptom is *"no matter what I put in Padding Key, only the default `x_padding` + Referer actually work"* — user reports on OpenWRT / Podkop / sing-box.

This PR closes the gap in **every** place where 3x-ui generates a link from an xhttp stream.

## What was dropped before

All four link generators only copied `path`, `host`, `mode` out of `xhttpSettings` and silently discarded:

- `xPaddingBytes`
- `xPaddingObfsMode`
- `xPaddingKey`
- `xPaddingHeader`
- `xPaddingPlacement`
- `xPaddingMethod`

This happened both in:

* `sub/subService.go` — the subscription service that serves `/sub/<id>` to external clients,
* `web/assets/js/model/inbound.js` — the panel's own "Copy URL" / QR / Info buttons in the inbound details modal.

So regardless of how the admin obtained the link (fetching a subscription or copying from the panel UI), the padding settings never reached them.

## Fix

Three files, three mirrored changes:

### `sub/subService.go` (server-side)

New helper `applyXhttpPaddingParams(xhttp, params)` writes two encodings side-by-side:

- `x_padding_bytes=<range>` — flat query param, read by sing-box-family clients (Podkop, OpenWRT sing-box, Karing, NekoBox, …).
- `extra=<url-encoded JSON>` — full xhttp settings blob, read by xray-core-family clients (v2rayNG, Happ, Furious, Exclave, …), including the obfs-mode key / header / placement / method.

Called from `genVlessLink`, `genTrojanLink`, `genShadowsocksLink`. `genVmessLink` copies the same fields directly into the VMess base64 JSON body.

### `web/assets/js/model/inbound.js` (panel UI)

Add two static helpers on `Inbound`:

- `Inbound.applyXhttpPaddingToParams(xhttp, params)` — for `vless://` / `trojan://` / `ss://` URLs. Same two-encoding scheme as the Go helper.
- `Inbound.applyXhttpPaddingToObj(xhttp, obj)` — for the VMess JSON blob.

Called from `genVmessLink`, `genVLESSLink`, `genTrojanLink`, `genSSLink`.

### `web/assets/js/model/outbound.js` (latent positional-arg bug)

Both the VMess-JSON path (L933) and `fromParamLink` (L975) were calling:

```js
new xHTTPStreamSettings(json.path, json.host, json.mode)
```

The constructor signature is `(path, host, headers, scMaxBufferedPosts, scMaxEachPostBytes, scStreamUpServerSecs, noSSEHeader, xPaddingBytes, mode, …)`, so `mode` was silently landing in the `headers` slot and the real `mode` field kept its `"auto"` default.

Fixed by constructing the object with only `path`/`host` positionally and setting `mode` by name. While touching the VLESS URL parser, also pick up `x_padding_bytes` and the `extra` JSON blob from the imported URL so the symmetric *import* of a padded link works too.

## Zero-impact on simple inbounds

All additions guard on *"is the field actually set?"* — empty strings and `xPaddingObfsMode: false` are skipped. An inbound that uses the UI defaults produces the exact same URL it produced before this patch, so existing subscriptions / deep-links are unaffected.

## Test plan

- [x] `go build ./...` + `go vet ./sub/...` clean
- [x] Inbound with no padding override → URL unchanged vs. pre-patch
- [x] Inbound with `xPaddingBytes="80-600"` only → URL gains `x_padding_bytes=80-600` and `extra={"xPaddingBytes":"80-600"}`; obfs-only fields not emitted
- [x] Inbound with obfs enabled + `xPaddingKey="remixsid"` + `xPaddingHeader="Cookie"` + placement/method → all set values land in `extra`; empty ones are omitted so xray falls back to its own defaults
- [x] Same result whether the admin pulls the link from the panel (`genVmessLink` / `genVLESSLink` / `genTrojanLink` / `genSSLink`) or from `/sub/<id>`
- [x] Importing a VLESS URL that carries `x_padding_bytes=...` and/or `extra=...` into an outbound now round-trips mode and the padding settings, instead of dropping mode into the headers slot

### DIFF
--- a/sub/subService.go
+++ b/sub/subService.go
@@ -250,6 +250,21 @@ func (s *SubService) genVmessLink(inbound *model.Inbound, email string) string {
 			obj["host"] = searchHost(headers)
 		}
 		obj["mode"], _ = xhttp["mode"].(string)
+		// VMess base64 JSON supports arbitrary keys; copy the padding
+		// settings through so clients can match the server's xhttp
+		// xPaddingBytes range and, when the admin opted into obfs
+		// mode, the custom key / header / placement / method.
+		if xpb, ok := xhttp["xPaddingBytes"].(string); ok && len(xpb) > 0 {
+			obj["x_padding_bytes"] = xpb
+		}
+		if obfs, ok := xhttp["xPaddingObfsMode"].(bool); ok && obfs {
+			obj["xPaddingObfsMode"] = true
+			for _, field := range []string{"xPaddingKey", "xPaddingHeader", "xPaddingPlacement", "xPaddingMethod"} {
+				if v, ok := xhttp[field].(string); ok && len(v) > 0 {
+					obj[field] = v
+				}
+			}
+		}
 	}
 	security, _ := stream["security"].(string)
 	obj["tls"] = security
@@ -408,6 +423,7 @@ func (s *SubService) genVlessLink(inbound *model.Inbound, email string) string {
 			params["host"] = searchHost(headers)
 		}
 		params["mode"], _ = xhttp["mode"].(string)
+		applyXhttpPaddingParams(xhttp, params)
 	}
 	security, _ := stream["security"].(string)
 	if security == "tls" {
@@ -604,6 +620,7 @@ func (s *SubService) genTrojanLink(inbound *model.Inbound, email string) string 
 			params["host"] = searchHost(headers)
 		}
 		params["mode"], _ = xhttp["mode"].(string)
+		applyXhttpPaddingParams(xhttp, params)
 	}
 	security, _ := stream["security"].(string)
 	if security == "tls" {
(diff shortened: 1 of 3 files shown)


pwnnex:
Closing — was addressing link generation, but the real bug is in inbound storage. Re-investigating.

MHSanaei:
Well done

pwnnex:
> Well done

release please new hotfix, brother
900RAM usage :(

MHSanaei:
@pwnnex tomorrow

PR #4083, opened 2026-04-22
Fix IP Limit continuous ban loop after a legitimate ban expires (#4077)

pwnnex:
Closes #4077.

## What users see

- Two IPs connect to the same config → fail2ban bans the excess one (correct).
- The ban duration expires (e.g. 1800s).
- The single legitimate IP tries to reconnect → gets banned again **immediately**.
- Loop continues forever. `/var/log/x-ui/3xipl.log` keeps filling with `[LIMIT_IP] … Disconnecting OLD IP =` lines for the same honest IP.

Works fine on v2.8.11, starts looping on v2.9.1.

## Root cause

`web/job/check_client_ip_job.go → updateInboundClientIps` merges two lists:

1. The persisted `inbound_client_ips.ips` blob (what we've seen historically).
2. Freshly observed lines from xray's access log.

Then it sorts **ascending by timestamp** and keeps the first `limitIp` entries. The intent (from 60abeaa) is *"protect the original/current connections, ban new excess ones"*.

The hole: nothing ever evicts an IP that simply stopped connecting. Its timestamp from the last time it was logged stays there forever. On the next run that stale IP is still "the oldest", so it wins the protected slot, and whichever IP is **actually** active right now is classified as "new excess" and gets re-banned. Every. Single. Run.

## Fix

Drop entries whose last-seen timestamp is older than `ipStaleAfterSeconds` (30 minutes) from the merge. An IP that is actually still active refreshes its timestamp any time xray emits a new `accepted` log line for a fresh TCP, which is well inside 30 minutes even for idle streaming. A client that has really gone away falls out of the table in bounded time and frees its slot.

The merge logic was extracted into a pure `mergeClientIps(old, new, staleCutoff) → map` helper so the policy can be exercised without the DB-backed job.

## Tests

5 subtests in `check_client_ip_job_test.go`, all pass. `go build ./...` + `go vet ./web/job/...` + `gofmt -l .` clean.

PR #4085, opened 2026-04-22
x-ui.sh: bundle nftables when installing fail2ban

pwnnex:
Fresh Debian 12+, Ubuntu 24+ and recent RHEL-family minimal server images ship `fail2ban` with `banaction = nftables-multiport` as the default in `/etc/fail2ban/jail.conf`, but do not pull in the `nftables` package. The first SSH brute-force attempt reaches the default `sshd` jail and fail2ban logs:

```
stderr: /bin/sh: 1: nft: not found
returned 127 -- HINT on 127: "Command not found"
```

over and over until the attacker gives up. Noisy, looks like a 3x-ui regression on first install — see the discussion on #4083 where a user updated, checked `Service Status` / `Ban Logs` from the x-ui menu, and assumed the update caused it.

The `3x-ipl` jail itself is fine because `create_iplimit_jails` writes an iptables-based action (`action.d/3x-ipl.conf` inherits from `iptables-allports.conf`). But the default sshd jail is inherited from the distro fail2ban config and the noise is real.

Add `nftables` to the package list in every branch of `install_iplimit` so fresh installs end up with a working default sshd jail. Installs that already have `nftables` present are a no-op.

No effect on existing deployments — `install_iplimit` only runs when the admin explicitly picks "Install IP Limit" from the x-ui menu, and only when `fail2ban-client` is not already on the system. 

### Test plan

- [x] bash syntax check on x-ui.sh (no errors)
- [x] diff review — only the package arg changed per distro, nothing else
- [x] behaviourally equivalent on systems where `nftables` is already installed (apt / dnf / pacman / apk all no-op an already-installed package)

### DIFF
--- a/x-ui.sh
+++ b/x-ui.sh
@@ -1802,7 +1802,14 @@ install_iplimit() {
     if ! command -v fail2ban-client &>/dev/null; then
         echo -e "${green}Fail2ban is not installed. Installing now...!${plain}\n"
 
-        # Check the OS and install necessary packages
+        # Install fail2ban together with nftables. Recent fail2ban packages
+        # default to `banaction = nftables-multiport` in /etc/fail2ban/jail.conf,
+        # but the `nftables` package isn't pulled in as a dependency on most
+        # minimal server images (Debian 12+, Ubuntu 24+, fresh RHEL-family).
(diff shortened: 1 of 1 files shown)


pwnnex:
oops, srry for ping

PR #4095, opened 2026-04-23
Feature/share quota

amirhosseinds:
## What is the pull request?

Added a new option when creating a new client that allow client to have a same traffic usage and expirytime with other clients in different inbounds with same subId. its tested and working.

- [x] Frontend
- [x] Backend

## Type of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Other


unspike:
so no shared quota in next releases ? 
@MHSanaei

MHSanaei:
@unspike i did better than this

PR #4213, opened 2026-05-10
fix(arch): correct x-ui service path

odrawq:
Fixes the Arch Linux systemd service path for x-ui.

The service file currently points to `/usr/lib/x-ui/x-ui`, but on Arch Linux the binary is installed to `/usr/local/x-ui/x-ui`. This causes the service to fail with `status=203/EXEC`.


nebulosa2007:
@odrawq In Arch, you should use the [package from the AUR](https://aur.archlinux.org/packages/3x-ui) instead of running the `install.sh` file.

According to [Arch package guidelines](https://wiki.archlinux.org/title/Arch_package_guidelines#Package_etiquette), installing packages in the `/usr/local/` directory is not permitted.



@MHSanaei Please revert the changes; the paths in the service file are correct.

MHSanaei:
@nebulosa2007 do it I'm not at home

PR #4299, opened 2026-05-12
fix(multi-node): scope tag and email unique constraints per node

abdalrahmanx9:
## Problem

When running a central panel with multiple remote nodes, the traffic sync silently drops inbounds from some nodes. The server log shows:

```
WARNING - setRemoteTraffic: create central inbound for tag inbound-48055 failed: UNIQUE constraint failed: inbounds.tag
WARNING - node traffic sync: merge for $NODE_NAME failed: UNIQUE constraint failed: client_traffics.email
```

**Two global UNIQUE constraints** are responsible:

1. **`inbounds.tag`** (`gorm:"unique"`) -- if Node A has an inbound on port 48055 (tag = `inbound-48055`), Node B cannot sync its inbound on the same port. The tag must be globally unique, even though each node runs a separate xray instance.

2. **`client_traffics.email`** (`gorm:"unique"`) -- if Node A has a client with email `foo@bar`, and Node B also has a client with that email, the client stats CREATE fails and **rolls back the entire transaction** -- including all inbounds that were successfully created in the same cycle. The node stays at 0 inbounds permanently, retrying every 10s and failing each time.

## Root cause

Tags and client emails only need to be unique **within the same xray instance**. A tag like `inbound-48055` on Node A is a completely different xray socket from `inbound-48055` on Node B -- they are on different machines. The same applies to client emails.

## Changes

4 files, +67/-9

- **`database/model/model.go`**: Changed `Tag` from `gorm:"unique"` (global) to `uniqueIndex:idx_node_tag` composited with `NodeID`.
- **`xray/client_traffic.go`**: Changed `Email` from `gorm:"unique"` (global) to `uniqueIndex:idx_ct_inbound_email` composited with `InboundId`.
- **`web/service/port_conflict.go`**: `tagExists()` scopes by `node_id` so the UI does not see remote tags as conflicts.
- **`database/db.go`**: Added `migrateLegacyUniqueConstraints()` that removes old UNIQUE constraints from existing databases on startup and creates the new composite indexes.


abdalrahmanx9:
this commit created an issue when deleting a client

abdalrahmanx9:
@MHSanaei please review this one carefully

pwnnex:
i will check it

PR #4316, opened 2026-05-12
fix(hysteria2): restore missing masquerade config in inbound form

### LINKED ISSUE #4314: Bug report: Hysteria 2 Masquerade option not avaible in v3.0.1
### Describe the bug

Hello.

I don't know if this issue occurs with a clean installation of the panel. My panel has consistently received every released update, including 3.0.0.
The problem is this: I can't find the Masquerade option in either previously created Hysteria 2 inbound installations or new ones.
Apparently, they forgot to add the new option when creating v3.0.0+, as Masquerade was added in previous releases, and I configured it myself.

### How to repeat the problem?

Check existing inbounds or create a new one

### Expected action

_No response_

### Received action

_No response_

### 3x-ui Version

3.0.1

### Xray-core Version

26.4.25

### Checklist

- [x] This bug report is written entirely in English.
- [x] This bug report is new and no one has reported it before me.

abdalrahmanx9:
## Summary

The Hysteria2 Masquerade option was missing from the Stream settings tab in the inbound form after the v3.0.0 rewrite, even though the data model (`HysteriaMasquerade` class in `inbound.js`) already supported it as an orphaned field.

Fixes #4314

## Changes

- `frontend/src/pages/inbounds/InboundFormModal.vue`: Added full Masquerade configuration UI in the Stream tab (visible only for Hysteria protocol). Includes enable toggle, type selector (Proxy/HTTP/File/String/404), and conditional fields per type (directory, URL, headers, status code, content).

- `sub/subJsonService.go`: Added `masquerade` passthrough in `genHy()` so the Xray JSON subscription config preserves the masquerade block.

**Note:** Masquerade is a server-side Xray feature already handled natively by Xray-core when the config is passed through — no additional server-side changes are needed.

### DIFF
--- a/frontend/src/models/inbound.js
+++ b/frontend/src/models/inbound.js
@@ -687,8 +687,9 @@ export class HysteriaMasquerade extends XrayCommonClass {
     }
 
     static fromJson(json = {}) {
+        const type = ['proxy', 'file', 'string'].includes(json.type) ? json.type : 'proxy';
         return new HysteriaMasquerade(
-            json.type,
+            type,
             json.dir,
             json.url,
             json.rewriteHost,
--- a/frontend/src/pages/inbounds/InboundFormModal.vue
+++ b/frontend/src/pages/inbounds/InboundFormModal.vue
@@ -1671,6 +1671,74 @@ watch(
               </a-select>
             </a-form-item>
           </template>
+
+          <!-- ====== Hysteria Masquerade ====== -->
+          <template v-if="protocol === Protocols.HYSTERIA">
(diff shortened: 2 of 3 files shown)


nebulosa2007:
Fixes https://github.com/MHSanaei/3x-ui/issues/4314 not 4303

Ramazik:
Thank you for your prompt response to the report. However, you did indeed provide the issue number incorrectly. The correct number is #4314, as noted in the comment above.

abdalrahmanx9:
edited thanks

Issue #4077, opened 2026-04-22
IP Limit (Fail2ban) continuous ban loop and false bans in v2.9.1


xAlokyx:
### Describe the bug

In version 2.9.1, the IP Limit feature (Fail2ban) has a bug where it creates an endless ban loop. 

When 2 different IPs connect to the same config, it correctly bans both. However, after the specified ban duration expires, if only ONE IP tries to connect, it gets banned again immediately. This loop continues indefinitely, making the config completely unusable even for a single, valid IP connection. It seems the connection history or state is not clearing properly after the ban is lifted.

### How to repeat the problem?

1. Upgrade from version v2.8.11 to 3x-ui v2.9.1 and ensure the IP Limit (Fail2ban) feature is enabled.
2. Connect to a single configuration using 2 different IPs simultaneously to trigger the initial ban.
3. Wait for the ban duration (e.g., 1800 seconds) to completely expire.
4. Attempt to connect to the same configuration using only ONE valid IP.
5. Check the Fail2ban logs (e.g., /var/log/x-ui/3xipl.log). The single IP will be banned immediately again, repeating the cycle.

*Note: I have attached screenshots of the terminal logs showing the ban entries and the Fail2ban actions*

<img width="5120" height="2272" alt="Image" src="https://github.com/user-attachments/assets/319eeb15-35b7-4869-aecb-564a428fffbf" />

<img width="5120" height="1256" alt="Image" src="https://github.com/user-attachments/assets/68e12357-53c5-464a-a718-396136263406" />

### 3x-ui Version

2.9.1

### Xray-core Version

26.2.6

### Checklist

- [x] This bug report is written entirely in English.
- [x] This bug report is new and no one has reported it before me.

DParadoxus:
Same ploblem. You can set the ban time to 1 minute and see the magic of banning the same ip every minute

Awixs777:
Same ploblem. I’ve temporarily disabled fail2ban. Hopefully this can be fixed quickly so I don’t have to roll back to version 2.8.11.
Version: 2.9.2

xAlokyx:
> Same ploblem. I’ve temporarily disabled fail2ban. Hopefully this can be fixed quickly so I don’t have to roll back to version 2.8.11. Version: 2.9.2
same issues in version 2.9.2 no fix

pwnnex:
fixed

PR #4319, opened 2026-05-12
Preserve historical client IPs

vahellame:
## What is the pull request?

Preserves historical client IP records in the IP log while keeping IP Limit enforcement based only on IPs observed in the current scan.

Previously, stale client IP entries were evicted after 30 minutes. This caused the panel IP log to lose historical IP addresses even though they are useful for auditing and visibility.

This change keeps historical IPs in `inbound_client_ips`, while `partitionLiveIps()` ensures that only currently observed IPs are counted toward `LimitIP` and sent to fail2ban.

* [ ] Frontend
* [x] Backend

## Type of Changes

* [x] Bug fix
* [ ] New feature
* [ ] Refactoring
* [ ] Other

## Screenshots

Not applicable.

### DIFF
--- a/web/job/check_client_ip_job.go
+++ b/web/job/check_client_ip_job.go
@@ -35,25 +35,6 @@ var job *CheckClientIpJob
 
 const defaultXrayAPIPort = 62789
 
-// ipStaleAfterSeconds controls how long a client IP kept in the
-// per-client tracking table (model.InboundClientIps.Ips) is considered
-// still "active" before it's evicted during the next scan.
-//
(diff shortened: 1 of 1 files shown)


vahellame:
I remember that the stale IP eviction was introduced to fix the continuous ban loop described in https://github.com/MHSanaei/3x-ui/issues/4077.

This PR keeps that fix intact by continuing to enforce `LimitIP` only against IPs observed in the current scan. At the same time, it allows the panel to preserve clients' historical IP addresses without clearing them from the database.

PR #4321, opened 2026-05-12
Preserve clients' IP history

vahellame:
## What is the pull request?

This PR preserves historical client IP records in `inbound_client_ips` while keeping IP Limit enforcement based only on IPs observed in the current access-log scan.

The previous stale IP eviction was introduced to fix the continuous Fail2Ban loop described in https://github.com/MHSanaei/3x-ui/issues/4077. A related regression was later tracked in https://github.com/MHSanaei/3x-ui/issues/4091, where db-only IPs could still affect ban decisions before `partitionLiveIps()` separated live IPs from historical ones.

This PR keeps that protection intact:

- `partitionLiveIps()` still ensures that only IPs observed in the current scan are counted toward `LimitIP`
- historical db-only IPs remain excluded from Fail2Ban decisions
- all live IPs, including IPs queued for Fail2Ban, are preserved in the database for visibility/history

The only behavior changed here is retention: client IP history is no longer removed from `inbound_client_ips` after 30 minutes.

- [ ] Frontend
- [x] Backend

## Type of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Other

## Screenshots

Not applicable.

### DIFF
--- a/web/job/check_client_ip_job.go
+++ b/web/job/check_client_ip_job.go
@@ -35,25 +35,6 @@ var job *CheckClientIpJob
 
 const defaultXrayAPIPort = 62789
 
-// ipStaleAfterSeconds controls how long a client IP kept in the
-// per-client tracking table (model.InboundClientIps.Ips) is considered
-// still "active" before it's evicted during the next scan.
-//
-// Without this eviction, an IP that connected once and then went away
-// keeps sitting in the table with its old timestamp. Because the
-// excess-IP selector sorts ascending ("oldest wins, newest loses") to
-// protect the original/current connections, that stale entry keeps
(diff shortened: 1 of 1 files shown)


MHSanaei:
the 30-minute eviction is intentional; without it the per-client IP blob grows unbounded. If you'd like an IP-history view, a separate audit table with its own retention would be the right shape. Thanks for the PR

vahellame:
No, it's not intentional. This is a dirty hack for Fail2Ban behavior introduced in 2.8.10

Also, after 2.8.10, the IP list is displayed upside down: newer IPs are shown below older ones

PR #4361, opened 2026-05-14
feat(qr): integrate QR code generation and update size parameter

farhadh:
## What is the pull request?

The QR is now a real SVG rendered by the `qrcode` library. It will stay crisp at any zoom level, and since it scales by CSS (width: 100%; height: auto), the modules sharpen further as you zoom into the page rather than blurring as the canvas version did.

- [x] Frontend
- [ ] Backend

## Type of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Other


MHSanaei:
we don't need this, super unnecessary, ant design already have qrcode

farhadh:
@MHSanaei ant design doesn't display the client's QR code - only the subscription one is displayed.
image

MHSanaei:
Screenshot 2026-05-14 112140

i don't see any problem

farhadh:
@MHSanaei for simple ones, yes, but if you add more to the config such that the link becomes longer, it wouldn't show. Try this on your side: VLESS + X25519 auth + reality + mldsa65.

MHSanaei:
This configuration is extremely large for a qrcodes and may exceed the practical limits of many QR scanners or clients.
For heavy configurations like this, it is recommended to use a subscription link instead of embedding the entire config into a qrcodes
large qrcodes can become too dense to scan reliably, especially on mobile devices.

MHSanaei:
i'll add a limit when they use mldsa65 qrcode for client doesn't show

farhadh:
fair :)

PR #4491, opened 2026-05-22
feat(inbound): Advanced XHTTP and external TLS proxy settings

beehunt9r:
## What is the pull request?

Pass XHTTP settings into VLESS import links and add ability to configure external proxy SNI, Fingerprint and ALPN. Important feature when using proxy behind external CDN with CNAME.

- [x] Frontend
- [x] Backend

## Type of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Other

## Screenshots

image


SPIRIT-00:
There is a similar need for grpc, ws, HTTPUpgrade, they are still used and can be placed behind reverse proxies.

PR #5020, opened 2026-06-06
fix(subscription): include XHTTP padding options in Clash output

w3struk:
## What changed

Map the six Xray XHTTP padding fields from `xhttpSettings` into Mihomo's `xhttp-opts` keys when generating Clash subscriptions:

- `xPaddingObfsMode` -> `x-padding-obfs-mode`
- `xPaddingBytes` -> `x-padding-bytes`
- `xPaddingKey` -> `x-padding-key`
- `xPaddingHeader` -> `x-padding-header`
- `xPaddingPlacement` -> `x-padding-placement`
- `xPaddingMethod` -> `x-padding-method`

Empty string values are omitted, while an explicitly present boolean value is preserved. Existing path, host, and mode behavior is unchanged.

## Why

The regular VLESS subscription retains these fields in `extra`, but the Clash/Mihomo YAML generator previously emitted only path, host, and mode. Mihomo clients therefore imported an incomplete XHTTP configuration.

The key names and value types match Mihomo v1.19.27's `XHTTPOptions` and VLESS-extra conversion.

## Validation

- `go test ./sub`
- `go test ./...`

### DIFF
--- a/sub/subClashService.go
+++ b/sub/subClashService.go
@@ -428,6 +428,21 @@ func (s *SubClashService) applyTransport(proxy map[string]any, network string, s
 			if mode, ok := xhttp["mode"].(string); ok && mode != "" {
 				opts["mode"] = mode
 			}
+			if value, ok := xhttp["xPaddingObfsMode"].(bool); ok {
+				opts["x-padding-obfs-mode"] = value
+			}
+			paddingFields := map[string]string{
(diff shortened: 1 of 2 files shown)


w3struk:
Closing because this change was prepared for a Mihomo-specific workflow that is not the target client for the deployment. We are reverting the integration and will use the existing Xray/VLESS subscription path instead.

PR #5076, opened 2026-06-08
feat(mtproto): add MTProto (FakeTLS) protocol via managed mtg sidecar

MHSanaei:
Xray-core has no mtproto proxy, so mtproto inbounds run as standalone mtg (9seconds/mtg) sidecar processes managed by the panel — one per inbound — and are excluded from the generated Xray config entirely.

- model: MTProto protocol constant, validator, and FakeTLS secret helpers (GenerateFakeTLSSecret/HealMtprotoSecret)
- mtproto package: per-inbound mtg process manager with reconcile, graceful stop, and best-effort Prometheus traffic scraping
- runtime: delegate mtproto inbounds to the mtg manager instead of the Xray gRPC API; skip mtproto when building the Xray config
- web: boot reconcile + StopAll wiring, periodic reconcile/traffic job, port-conflict transport, secret healing on inbound add/update
- sub: tg:// proxy share-link generation
- frontend: protocol option, Zod schema, Protocol tab (FakeTLS domain + regenerable secret), info-modal link, and i18n
- provisioning: fetch mtg v2.2.8 in install.sh, DockerInit.sh, and the Linux + Windows release workflows


trvkn:
@MHSanaei, native MTProto landing is great, but one thing worth reconsidering: mtg is by its author's own design a single-secret, no-management throwaway proxy, so a multi-user panel ends up spawning one process per inbound and scraping Prometheus best-effort - fighting the tool rather than using it. mtg's own README now **points users to telemt** ([telemt/telemt](https://github.com/telemt/telemt)) for **adtag support** (promoted-channel sponsorship), so panel operators who want to monetize/sponsor channels could use that option, as well as managed/multi-secret use: telemt is built around a management API, native multi-secret config, first-class Prometheus, and a heavily-debugged FakeTLS/TCP-splice masking path - under active development (67 releases, Rust + Tokio). 

It's also **proven effective in real DPI environments** (RU TSPU, China's GFW, Iran), where keyless probes get a transparent splice to a real upstream with valid chain-of-trust (no MITM, no fake certs). The main cost is per-arch Rust artifacts vs mtg's tiny static Go binaries, so it could land as an alternative backend behind the same manager abstraction rather than a hard replacement.

Issue #5074, opened 2026-06-08
[Feature]: Add custom vless:// to subscription page


maxbranson64-eng:
### Is your feature request related to a problem?

There are lots of github repos that provide txt lists of public vless servers but you cannot add them to your subscription page

### Describe the solution you'd like

Make it possible to add third-party vless:// links to Subscription page or even txt lists. The ability to http test will be a great option too. The idea is:
1. Add link to a txt list with vless:// in it
2. Parse and test each link with HTTP url test to make sure that tunnel works fine
3. Add link to subscription page
4. Retest each 1-24 hours

That will make possible to expand connection options with third-party servers. Since lists are updated automatically you will get fresh and checked lists on each subscription update.

### Are you willing to help?

- [ ] I'd like to implement this feature myself and open a PR.
- [ ] I can help test once a PR is open.

### Before submitting

- [x] I searched [existing issues](../issues?q=is%3Aissue) and this feature has not been requested.
- [x] This feature request is written in English.

PR #5227, opened 2026-06-12
Add client outbound attachment support as external config

saeedhasani:
## Summary

This PR adds support for attaching selected outbound configurations to individual clients and exposing them through client subscriptions as external configurations.

A new **"Attached Outbounds (External Config)"** field has been added to the client create/edit form, supporting multi-select, select-all, and clear-all actions similar to attached inbounds.

Only outbounds explicitly marked with **"Use as External Config"** are available for client assignment and subscription generation.

## Why

Many operators use 3x-ui alongside third-party or provider-managed nodes and would like to distribute those configurations through the same client subscription endpoint.

Before this change, subscriptions could only be generated from attached inbounds. There was no built-in way to include externally imported outbound configurations in a user's subscription feed.

This PR enables:

* Attaching external outbound configurations directly to clients.
* Delivering third-party/proxy-provider configurations through standard client subscription links.
* Supporting multiple protocol types as long as they are represented as supported outbounds (e.g. VLESS, VMess, Trojan, Shadowsocks, and other outbound-based configurations).
* Supporting clients that rely entirely on external configurations without requiring subscription generation from local inbounds.

### Backend

* Added a new `client_outbounds` model/table for client-to-outbound associations.
* Added APIs:

  * `GET /panel/api/clients/outboundOptions`
  * `POST /panel/api/clients/:email/attachOutbounds`
  * `POST /panel/api/clients/:email/detachOutbounds`
* Extended client create/edit flows to persist and hydrate outbound assignments.
* Updated subscription generation to include attached external outbounds.
* Updated JSON subscription generation to export selected outbounds as clean client-facing configurations.
* Internal metadata fields are removed from generated subscription outputs.

### Frontend

* Added **Attached Outbounds (External Config)** to the client form.
* Added **Use as External Config** switch to outbound settings.
* Reused existing multi-select UX patterns (select all / clear all).

### Subscription Output

* External outbound links now preserve full transport and stream settings, including advanced TLS/Reality parameters when present.
* Generated subscriptions retain the original outbound credentials and imported configuration details.
* Metadata used internally by the panel is excluded from exported client configurations.

## Verification

Successfully verified with:

```bash
go test ./internal/sub ./internal/web/service ./internal/web/controller ./internal/database/model
```

and

```bash
npm run build
```

`git diff --check` completed without issues.


saeedhasani:
Resolves #5074

MHSanaei:
i made a better version 
image

PR #5241, opened 2026-06-12
perf(db): index group_name and client_traffics hot columns

n0ctal:
## Summary

This PR adds the missing hot-path database indexes for:

- `client_records.group_name`
- `client_traffics.inbound_id`
- `client_traffics.last_online`

These columns are used frequently in filtering, grouping, and traffic-related reads, but were not indexed.

## Why

On larger datasets, these lookups become unnecessarily expensive and increase query latency for common panel operations.

This change keeps the schema behavior the same while improving read performance on existing and future databases.

## Scope

- adds index tags for the three hot columns
- relies on the existing migration / AutoMigrate flow
- no behavioral change

## Validation

- added/updated tests to assert the indexes exist after migration
- validated with database tests

## Risk

Low. This is a schema/index improvement only and does not change application logic.

### DIFF
--- a/internal/database/index_tags_test.go
+++ b/internal/database/index_tags_test.go
@@ -1,0 +1,41 @@
+package database
+
+import (
+	"testing"
+
+	"gorm.io/driver/sqlite"
+	"gorm.io/gorm"
+	"gorm.io/gorm/logger"
+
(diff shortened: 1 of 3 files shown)


MHSanaei:
hi, thanks for your works, all of your PRs have conflicts please update your fork

PR #5270, opened 2026-06-13
fix(subscriptions): avoid shared mutable state during generation

n0ctal:
## Summary

This PR removes request-to-request shared mutable subscription-generation state by making request preparation explicitly request-scoped.

## Why

Subscription generation can be called concurrently for different hosts / formats. Shared mutable service state makes that concurrency harder to reason about and can lead to cross-request contamination.

This change keeps the same output behavior while making the generation flow safer under concurrent access.

## Scope

- isolates request-specific preparation state
- keeps existing subscription generation behavior
- improves correctness under concurrent requests

## Validation

- added/updated focused subscription tests
- validated against the subscription package test suite

## Risk

Medium. This changes internal state handling in a concurrency-sensitive area, but the patch is intentionally narrow and test-backed.

### DIFF
--- a/internal/sub/build_urls_test.go
+++ b/internal/sub/build_urls_test.go
@@ -70,6 +70,27 @@ func TestBuildURLs_EmptySubId(t *testing.T) {
 	}
 }
 
+func TestForRequestDoesNotMutateSharedService(t *testing.T) {
+	initSubDB(t)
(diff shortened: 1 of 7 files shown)


MHSanaei:
please fix conflicts

PR #5271, opened 2026-06-13
feat(web): cap request body size on state-changing routes

n0ctal:
## Summary

This PR adds a request body size limit for state-changing routes.

It uses `http.MaxBytesReader` to reject oversized request bodies before they cause unnecessary memory allocation or long processing on mutation endpoints.

## Why

Even with authentication in place, large request bodies on create/update/import-style routes are an avoidable resource risk. A reasonable cap improves resilience against buggy clients and accidental oversized submissions.

## Scope

- adds body-size limiting middleware
- applies it to state-changing requests
- does not affect normal read-only requests

## Validation

- added tests for:
  - under-limit request bodies
  - over-limit request bodies
  - routes without request bodies

## Risk

Low. This is defensive hardening and does not alter normal successful request flows.


MHSanaei:
please fix conflicts

PR #5323, opened 2026-06-14
feat: release-driven golden-image & unattended-install deployment pipeline

MHSanaei:
## Summary

Adds a release-driven "golden image" + unattended-install deployment pipeline so the panel can be shipped as a cloud image or one-click cloud-init install, with **unique per-instance credentials generated on first boot**.

Motivation: today `install.sh` is fully interactive and the panel seeds a hardcoded `admin/admin` + generates its session secret/panel GUID on first start. That made any pre-baked image a security landmine (every clone sharing the same admin creds + secret). This PR makes unattended install first-class and guarantees per-instance secrets.

## What's added (`deploy/`)

- **Non-interactive `install.sh`** — triggers on `XUI_NONINTERACTIVE=1` or a non-TTY stdin; every prompt becomes an env var/default, interactive UX unchanged. Writes `/etc/x-ui/install-result.env` (mode 600). Honors `XUI_USERNAME/PASSWORD/PANEL_PORT/WEB_BASE_PATH/SSL_MODE/DOMAIN/ACME_EMAIL/DB_TYPE/DB_DSN`.
- **`deploy/firstboot/`** — oneshot unit (ordered `Before=x-ui.service`, sentinel-guarded) that replaces the seeded `admin/admin` with random creds on a random high port, regenerates the secret/GUID, mints an API token, and writes them to `/etc/x-ui/credentials.txt` + MOTD. Idempotent.
- **`deploy/packer/`** — one build, two sources (`amazon-ebs` AMI + `qemu` qcow2/raw) for **amd64 and arm64**; `provision.sh` (installs, **no DB**) → `harden.sh` (key-only SSH, no default passwords) → `cleanup.sh` (wipes any DB/creds/keys, fails the build if a secret survives).
- **`deploy/cloud-init/cloud-init.yaml`** — generic user-data for Hetzner/AWS/DO/Vultr/GCP/Azure/Oracle.
- **`deploy/lightsail/`** — Lightsail launch script (self-service) + AWS-CLI snapshot builder (Lightsail can't use an EC2 AMI).
- **`deploy/marketplace/`** — AWS Marketplace (EC2 AMI) + Hetzner submission notes.
- **`.github/workflows/image.yml`** — on `release: published`, builds amd64/arm64 qcow2 (attached to the release) and AMIs (gated on AWS secrets so forks skip cleanly). Default region `eu-central-1`.
- **`.github/workflows/smoke.yml`** + `deploy/test/` — container smoke tests on amd64 and arm64 runners.

## Security guarantee (verified)

Two fresh instances from the same setup produce **different** username, bcrypt password hash, 32-char session secret, API token, port and base path — nothing shared. The image contains no `x-ui.db`; `cleanup.sh` enforces it.

## Testing

- Non-interactive install + first-boot smoke tests pass in Docker on **amd64 and arm64** (real released binaries; `ELF aarch64` confirmed, `hasDefaultCredential: false`).
- `packer fmt -check` + `packer validate` pass for both arches; `actionlint` + `shellcheck` clean on all new shell.
- Not executed locally (need a real cloud account): the actual AMI/qcow2 cloud builds and the Lightsail snapshot pipeline — these run in CI / on AWS.

## Notes

- The interactive install experience is unchanged; new behavior only triggers in non-interactive mode.
- `cloud-init.yaml` and the Lightsail scripts fetch `install.sh`/firstboot from `main`, so they go fully live once this merges.
- No secrets, keys, AMI IDs, or account numbers are committed. The AWS AMI build is optional and gated on `AWS_ROLE_ARN`/`AWS_ACCESS_KEY_ID`.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,3 +4,9 @@ DockerEntrypoint.sh text eol=lf
 frontend/src/generated/** text eol=lf
 frontend/public/openapi.json text eol=lf
 frontend/src/test/__snapshots__/** text eol=lf
+
+# Cloud-image deploy assets are consumed on Linux — force LF regardless of host.
+*.service text eol=lf
+deploy/**/*.service text eol=lf
+deploy/**/*.hcl text eol=lf
+deploy/**/*.yaml text eol=lf
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -115,7 +115,7 @@ jobs:
           sudo apt-get install -y --no-install-recommends ${{ matrix.qemu_pkgs }}
 
       - name: Setup Packer
-        uses: hashicorp/setup-packer@main
+        uses: hashicorp/setup-packer@v3
         with:
           version: latest
 
@@ -203,7 +203,7 @@ jobs:
         uses: actions/checkout@v6
 
       - name: Setup Packer
-        uses: hashicorp/setup-packer@main
+        uses: hashicorp/setup-packer@v3
         with:
           version: latest
 
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -1,7 +1,7 @@
 name: Build Cloud Images
 
-# Build golden cloud images from a published release:
-#   * qemu      -> qcow2 attached to the GitHub release (always)
+# Build golden cloud images from a published release, for amd64 and arm64:
+#   * qemu       -> qcow2 attached to the GitHub release (always)
 #   * amazon-ebs -> AWS AMI (only when AWS credentials are configured)
 #
 # Images contain NO database and NO baked credentials; first boot generates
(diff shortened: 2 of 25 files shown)


MHSanaei:
Thanks for the review — all 9 findings addressed in 7e723bc3. Each was independently verified (and the uncertain Lightsail-key one researched) before fixing; both container smoke tests still pass and shellcheck/actionlint are clean.

| Finding | Fix |
| --- | --- |
| `install-result.env` written unquoted (sourceable → injection) | `write_install_result` now uses `printf %q` (no-op for the alphanumeric random defaults; safe for pinned values) |
| firstboot: empty/unparsable `-show` → silent skip could leave `admin/admin` | Fail-closed: require exactly `true`/`false`, else `exit 1` **without** writing the sentinel so the next boot retries |
| `credentials.txt` written unescaped | `printf %q` |
| Password/API token in world-readable `/etc/motd` (firstboot) | MOTD now shows only URL + username; secrets via `sudo cat /etc/x-ui/credentials.txt` (mode 600) |
| `deploy/README.md` env-var list incomplete | Added `XUI_ACME_HTTP_PORT`, `XUI_SSL_IPV6`, `XUI_SERVER_IP` |
| Lightsail key saved without base64 decode | Detect `-----BEGIN` (PEM) vs base64 and handle both, then assert a valid PEM before SSH |
| `setup-packer@main` unpinned (×2) | Pinned to `@v3` |
| Full creds appended to `/etc/motd` (cloud-init) | Full creds → serial console only; `/etc/motd` gets URL + username + pointer |

Also fixed the same `/etc/motd` secret-exposure pattern in `deploy/lightsail/launch-script.sh` (not flagged, but the same class of issue).

PR #5341, opened 2026-06-15
fix(ui): correct selector tag spacing in balancer inbounds

jimself218-ops:
Fix missing space between selector tags in the balancer row on the inbounds page.
Tags were rendered without proper spacing due to incorrect inline style syntax.
Before:
image
After:
image


MHSanaei:
fixed in #5343

PR #5362, opened 2026-06-15
perf(xray): compile log/traffic regexps once at package scope

n0ctal:
## Summary

Compiles the log and traffic regexps once at package scope instead of on every call.

## Why

`GetTraffic` recompiled two stats regexps on every traffic-stats tick, and `LogWriter.Write` recompiled two more on every log line — wasted work on hot paths.

## Scope

- hoists `trafficRegex` / `clientTrafficRegex` and `crashRegex` / `logLineRegex` to package-level vars
- patterns are identical; no behavior change

## Validation

- `xray` package tests pass
- `go build` / `go vet` clean

## Risk

Low. Pure refactor — the regex patterns are unchanged.


n0ctal:
Heads-up: the `codegen` check here is red due to stale committed frontend types on `main` — the `staticEgressResolver` type added in c1fdcd98 wasn't regenerated, so `npm run gen` produces a diff on every PR. This PR only touches Go (regex hoisting) and is unrelated; #5365 regenerates the types and clears the gate (this PR goes green on rebase once it lands).

Issue #5141, opened 2026-06-09
[Bug]: scMinPostsIntervalMs: 30 is automatically added in 3.x.x and triggers TSPU blocking on mobile networks in Russia


ttrroyy:
### Describe the bug

After updating to the 3.x.x branch, many users started experiencing mass connection blocks.
The panel now automatically injects the following parameters into the Xray configuration:
JSON"scMaxEachPostBytes": "1000000",
"scMinPostsIntervalMs": "30"
The value scMinPostsIntervalMs: 30 is quickly detected by Russia’s TSPU (traffic filtering system) on mobile networks, causing immediate blocking.
These parameters are being added to:

The Advanced Xray template
The Extra block in client links and subscriptions

### How to reproduce the problem

Workaround:
Manually remove both lines from the Advanced template and from the Extra block in client configs.
Request:
Please make scMinPostsIntervalMs optional or remove the automatic injection of this aggressive default value, as it breaks usability on Russian mobile networks.

### Relevant logs

```shell

```

### 3x-ui version

3.x.x

### Xray-core version

25.x.x

### How did you install 3x-ui?

install.sh script

### Operating system

Ubuntu 24.04

### Is the panel behind a reverse proxy or CDN?

None

### Before submitting

- [x] I searched [existing issues](../issues?q=is%3Aissue) and this bug has not been reported.
- [x] I am running the latest released version of 3x-ui (or have verified the bug still exists on it).
- [x] This bug report is written in English.
- [x] I have redacted any sensitive data (tokens, real domains, client UUIDs).

SpeedeMiner:
Is this the only parameter that triggers blocking?
"scMinPostsIntervalMs": "30".
I only use this parameter, scMaxEachPostBytes": "2000000". Should I remove it as well?

MHSanaei:
Fixed in d94c3fc5.

Root cause confirmed: the panel seeded every XHTTP config with `scMaxEachPostBytes=1000000` and `scMinPostsIntervalMs=30` — which are xray-core''s own internal defaults — and then emitted them into every generated config and share link. The literal `scMinPostsIntervalMs=30` is exactly the stable fingerprint TSPU keys on.

The fix has two parts so it helps both new and existing setups:

- **New configs** no longer seed these values at all (empty defaults in the inbound/outbound schemas and templates), so xray-core applies its own defaults silently with nothing on the wire to fingerprint.
[...]

PR #5393, opened 2026-06-16
fix(sub): preserve non-default scMinPostsIntervalMs and use per-inbound xmux in JSON subscriptions

w3struk:
## Problem

JSON subscriptions generated by 3x-ui had two bugs that prevented XHTTP client tuning knobs from reaching subscribers, and the inbound/outbound forms had mode-conditional visibility issues that hid important fields.

### 1. `scMinPostsIntervalMs` never reaches the DB

The frontend wire normalizer (`normalizeXhttpForWire`) unconditionally deleted `scMinPostsIntervalMs` from inbound configs before persisting to the database:

```typescript
if (side === 'inbound') {
    delete out.scMinPostsIntervalMs;  // always deleted, even non-default values
}
```

This meant that even when an admin set a custom value like `"50-150"`, it was silently dropped on save. JSON subscriptions could never include it because it was never stored.

The xray-core default `"30"` is a known DPI fingerprint (#5141) and must still be stripped, but non-default tuning values must survive the round-trip so that `buildXhttpExtra` and the JSON subscription generator can propagate them to clients.

### 2. JSON subscriptions ignore per-inbound `xmux`

The JSON subscription generator always used the global `subJsonMux` panel setting for `outbound.Mux`, even when the inbound carried per-inbound `xmux` inside `xhttpSettings`. This meant:

- XHTTP outbounds with `xmux` still got the legacy `mux.cool` block injected
- The inbound's own `xmux` was silently ignored in JSON subscriptions
- Share links (`vless://`) already read per-inbound `xmux` via `buildXhttpExtra` — JSON subscriptions were the only format that didn't

### 3. Inbound form hides server-side fields behind mode gates

`scMaxEachPostBytes` is a server-side field used by xray-core in **every** mode (both `handlePacketUp` and `handleStreamUp` validate it). It was hidden behind the `packet-up/auto` conditional, making it invisible in `stream-up` and `stream-one` modes.

`scMaxBufferedPosts` is only used by `handlePacketUp` and correctly belongs behind the `packet-up/auto` gate.

`scMinPostsIntervalMs` is client-only and packet-up/auto-specific — it was missing from the inbound form entirely.

### 4. Outbound form hides `scMinPostsIntervalMs` for `mode: "auto"`

Because xray-core resolves client `mode: "auto"` with TLS to `packet-up`, the `scMinPostsIntervalMs` field must be visible for `auto` mode too — otherwise users cannot configure it for the most common TLS+XHTTP setup.

### Commit 1: Preserve non-default `scMinPostsIntervalMs` in inbound wire payload

**File:** `frontend/src/lib/xray/stream-wire-normalize.ts`

Instead of unconditionally deleting `scMinPostsIntervalMs`, only strip the xray-core default (`"30"`) and empty values. Custom tuning values like `"50-150"` now survive the round-trip.

### Commit 2: Use per-inbound `xmux` instead of global `subJsonMux` in JSON subscriptions

**File:** `internal/sub/json_service.go`

`getConfig()` now checks whether `xmux` is present in the inbound's `xhttpSettings`. When it is, the per-inbound `xmux` handles multiplexing and the legacy `outbound.Mux` is suppressed. When `xmux` is absent, the global `subJsonMux` is used as before (backward compatible).

### Commit 3: Add `scMinPostsIntervalMs` to inbound XHTTP form

**File:** `frontend/src/pages/inbounds/form/transport/xhttp.tsx`

Adds the missing UI field for `scMinPostsIntervalMs` in the inbound form's packet-up section, so admins can set anti-DPI intervals directly from the panel.

### Commit 4: Show packet-up fields for `auto` mode in inbound XHTTP form

**File:** `frontend/src/pages/inbounds/form/transport/xhttp.tsx`

Because xray-core resolves client `mode: "auto"` with TLS to `packet-up`, the packet-up fields (including `scMinPostsIntervalMs`) are now also shown when the inbound mode is `auto`.

### Commit 5: Show `scMinPostsIntervalMs` for `auto` mode in outbound form, update placeholder

**File:** `frontend/src/pages/xray/outbounds/transport/xhttp.tsx`

Same consistency fix for the outbound form: `scMinPostsIntervalMs` is now visible for `mode: "auto"`. Placeholder changed from `"30"` (DPI fingerprint) to `"e.g. 50-150"`.

### Commit 6: Show `scMaxEachPostBytes` for all modes in inbound form

**File:** `frontend/src/pages/inbounds/form/transport/xhttp.tsx`

`scMaxEachPostBytes` is used by xray-core in every mode (both `handlePacketUp` and `handleStreamUp` validate it). It is now always visible regardless of mode. `scMaxBufferedPosts` remains gated behind `packet-up/auto` since it is only used by `handlePacketUp`.

### Commit 7: Update XhttpForm snapshot

**File:** `frontend/src/test/__snapshots__/inbound-form-blocks.test.tsx.snap`

Snapshot updated to reflect `scMaxEachPostBytes` being unconditionally visible.

## Why these changes matter

- **Anti-DPI:** The default `scMinPostsIntervalMs: 30` is a known fingerprint. Allowing users to set ranges like `"50-150"` and propagating them to subscriptions reduces fingerprinting.
- **XHTTP xmux correctness:** `mux.cool` and XHTTP `xmux` are different multiplexers. Using the global `subJsonMux` (mux.cool) for XHTTP outbounds caused many short-lived TCP connections, which is also fingerprintable. Per-inbound `xmux` fixes this.
- **Mode consistency:** We now default to explicit `stream-up` in our deployment template because `auto` with TLS resolves to `packet-up` in xray-core, contrary to what the public docs claim.
- **Form completeness:** Server-side fields used in all modes should not be hidden behind client-mode gates. Client-only fields should only appear for relevant modes.

## Backward Compatibility

| Scenario | Before | After |
|----------|--------|-------|
| Inbound without `xmux`, global `subJsonMux` set | `outbound.Mux` from global | Same |
| Inbound with `xmux`, global `subJsonMux` set | `outbound.Mux` from global (double-mux) | `outbound.Mux` suppressed, `xmux` in `xhttpSettings` |
| `scMinPostsIntervalMs = "30"` (default) | Deleted | Deleted |
| `scMinPostsIntervalMs = "50-150"` (custom) | Deleted (bug) | Preserved |
| `scMinPostsIntervalMs = ""` (empty) | Deleted | Deleted |

## Testing

Deployed to a test server and verified:
- `scMinPostsIntervalMs: "50-150"` appears in JSON subscription
- `xmux: {"maxConcurrency": "16-32", "hMaxRequestTimes": "600-900"}` appears inside `xhttpSettings`
- `outbound.Mux` is suppressed when `xmux` is present
- Without `xmux`, global `subJsonMux` is used as `outbound.Mux`
- `scMaxEachPostBytes` is visible in all modes in the inbound form
- `scMaxBufferedPosts` is visible only for `packet-up` and `auto` modes
- `scMinPostsIntervalMs` is visible for `packet-up` and `auto` modes in both inbound and outbound forms


ilya-practicum:
btw, currently ui missing scMinPostsIntervalMs parameter at all

ilya-practicum:
Thx

PR #5396, opened 2026-06-16
fix(xray): verify the release archive checksum before installing

n0ctal:
## Summary

Verify the downloaded Xray-core release archive against its published `SHA2-256` before installing, instead of installing it after only a TLS fetch + HTTP-200 + a size cap.

## Why

`UpdateXray` → `downloadXRay` fetched the release zip and installed the binary from it with no integrity check on the artifact itself. TLS protects the transport, not the bytes: a corrupted or tampered release asset (a bad mirror/CDN, or a proxy that terminates TLS) would be extracted and run as the panel's long-lived xray process.

## Scope

- after download, fetch the `.dgst` sidecar XTLS publishes next to each asset, parse its `SHA2-256` (the label is `SHA2-256=`, not `SHA256=`), compute the archive's SHA-256, and abort on mismatch
- abort the update when the `.dgst` is unreachable or has no/short `SHA2-256` entry — fail closed, never install an unverified archive. *(Maintainer note: this hard-fails updating to any release without a published `.dgst`; happy to switch to warn-and-proceed if you'd prefer to keep installing those.)*
- the `.dgst` read is capped at 64 KiB

## Validation

- unit tests for the digest parser (the real v26.3.27 `.dgst` line format, plus missing/malformed entries) and the fetch (httptest 200 + 404)
- `internal/web/service` package tests pass

## Risk

Low. Adds a verification step inside the existing download path; no API or behaviour change beyond rejecting an unverifiable archive.

### DIFF
--- a/internal/web/service/server.go
+++ b/internal/web/service/server.go
@@ -846,7 +846,11 @@ func (s *ServerService) downloadXRay(version string) (string, error) {
 		return "", err
 	}
 	if got := hex.EncodeToString(hasher.Sum(nil)); !strings.EqualFold(got, want) {
-		return "", fmt.Errorf("download xray: checksum mismatch (want %s, got %s)", want, got)
+		// User-facing warning: the archive's SHA-256 does not match the official
+		// release checksum, so the download is corrupted or has been tampered
+		// with. Abort the install so a bad binary is never run, and tell the user
+		// to retry/re-download rather than proceed with a mismatched image.
+		return "", fmt.Errorf("Xray update aborted: the downloaded archive does not match the official SHA-256 checksum, so the image is corrupted or differs from the official release. Please exit and re-download the official image, then try again (expected %s, got %s)", want, got)
 	}
 
 	ok = true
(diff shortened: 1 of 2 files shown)


ckun52880:
Thank you very much for your contribution. I completely understand why you added this pull request—a faulty image/CDN, or a proxy that terminates TLS, leading to the prolonged use of an incorrect kernel. However, could you please add a warning if the SHA256 comparison is inconsistent? This warning would allow users to choose whether to proceed. This would be necessary for some users. The warning could be supplemented with something like: "When ordinary users see this message, it indicates that the image is corrupted or inconsistent with the official image; it is recommended to exit the installation and re-pull the image," or something similar. Please forgive any ambiguity in my wording.

n0ctal:
Thanks for the review! I've added a plain-language warning on a SHA-256 mismatch, using your suggested wording: it now tells the user the downloaded archive is corrupted or differs from the official release and to exit and re-download the official image before retrying.

I kept the update **aborting** on a mismatch rather than offering a "proceed anyway", since auto-proceeding past a failed checksum would install a binary that may be corrupted or tampered with — exactly what the check is meant to prevent. The new message points the user to the safe next step (re-download). If you'd still prefer a user-overridable "proceed anyway" path, I'm happy to add it behind an explicit force flag — just let me know.

PR #5491, opened 2026-06-22
fix(online): gate local inbounds correctly when xray is idle

younesvatan78:
## Summary

Multi-inbound clients appeared online on every inbound they were attached to, even when they had only used one of them. The bug surfaced whenever the local xray had no recent traffic (grace window expired between polls).

## Root cause

\`GetActiveInboundsByGuid\` returned an empty map \`{}\` when \`GetLocalActiveInbounds\` returned nil (no active inbounds). The frontend uses the presence/absence of a GUID key to decide whether to gate the online display:

- **Key absent (\`undefined\`)** → "unknown node, don't gate" — correct fallback for remote nodes that haven't reported yet
- **Key present, empty slice** → "node known, no active inbounds" — correct for the local panel when xray is idle

Because the local panel's GUID was missing from the map when idle, the frontend's \`activeByGuidRef.current.get(guid)\` returned \`undefined\`, and the \`undefined === undefined\` fallback in \`rollupClients\` ungated all local inbounds, making every client appear online on all their inbounds.

## Fix

Move the \`panelGuid\` lookup before the active-inbounds check and always emit \`{guid: []}\` instead of \`{}\` when the local panel has no active inbounds. A missing key still means "unknown remote node — don't gate", which is the right semantic for nodes that haven't reported.

## Test plan

- [ ] Attach one client to two local inbounds
- [ ] Generate traffic on inbound A only
- [ ] Confirm client shows online on inbound A but not inbound B
- [ ] Wait for the xray grace window to expire (or restart xray); confirm client no longer shows online on either inbound

### DIFF
--- a/internal/web/service/client_traffic.go
+++ b/internal/web/service/client_traffic.go
@@ -121,22 +121,29 @@ func (s *ClientService) resetAllClientTrafficsLocked(id int) error {
 	now := time.Now().Unix() * 1000
 
 	if err := db.Transaction(func(tx *gorm.DB) error {
-		whereText := "inbound_id "
-		if id == -1 {
-			whereText += " > ?"
-		} else {
-			whereText += " = ?"
-		}
-
+		// client_traffics.inbound_id is stale: it reflects the inbound the row was
+		// first inserted under and is never refreshed. Use the client_inbounds join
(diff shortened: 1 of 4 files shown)


younesvatan78:
Closing — reopening from a clean single-commit branch rebased on current main.

PR #5507, opened 2026-06-23
feat(traffic): inbound Traffic Multiplier (per-attachment billing)

nima1024m:
## Summary

Adds a per-inbound **Traffic Multiplier** that scales how fast a client's real traffic draws down their quota — so operators can price premium tunnels (multiplier > 1) and economy tunnels (< 1) differently, with the rate made visible to the end user. Default `1` is fully behaviour-preserving.

Billing is **exact and fair across inbounds**: each `(client, inbound)` pair is metered separately, each slice is multiplied by its inbound's multiplier, and the slices are summed. Worked example — a client on a 2× and a 0.5× inbound who moves 50 GB and 30 GB is billed `50×2 + 30×0.5 = 115 GB` against quota while **Real** stays 80 GB.

## How it works

- **Real Traffic** = bytes actually moved (unchanged metering). **Billed Traffic** = Real after each inbound's multiplier, summed — the number quota is enforced against.
- Xray meters usage per client-email (summed across inbounds), so exact per-inbound billing needs per-attachment metering. Each `(client, inbound)` is given its own Xray accounting identity `"<inboundId>::<email>"`; the suffix is an internal detail and is stripped back to the logical email everywhere else (IP-limit grouping, online detection, subscriptions, UI).
- **Non-retroactive**: Billed is accrued at the multiplier in force at each poll, so changing an inbound's multiplier never re-bills past usage.
- **Multi-node**: the multiplier rides to nodes as an inbound field; each hosting panel bills its own slice; the master sums Real + Billed across the fleet (the node→master fold and the master→node enforcement push both carry Billed). Old nodes degrade safely to 1×.

## Feature commits (one per phase)

1. **Phase 0** — schema: `Inbound.multiplier` (0.1–100, default 1), new `client_inbound_traffics` per-attachment table, `billed_up/billed_down` on client/node/global traffic, idempotent backfill (Billed = Real for existing rows).
2. **Phase 1** — per-attachment metering: encode identity at config build + live AddUser/RemoveUser; decode on the stats path; accrue Real per attachment + Billed into the per-client aggregate.
3. **Phase 2** — enforce quota on Billed (client depletion predicates, incl. the cross-panel push); zero Billed + clear the per-tunnel breakdown on every reset/renew/delete. Detach retains Billed (no refund).
4. **Phase 3** — decode the per-attachment identity in the IP-limit and online-detection paths so grouping stays per-client.
5. **Phase 4** — propagate Billed + the multiplier across the node fleet.
6. **Phase 5** — UI + `{MULTIPLIER}` subscription token: inbound form field, client modal Real-vs-Billed, an inbounds-list multiplier filter (range buckets + exact values), full localization across all 13 locales, regenerated codegen.

## Code review — two max-effort rounds

The change was put through two rounds of multi-agent, recall-oriented review (many independent finder angles → adversarial verification → gap sweep), each landing as one additive fix commit. No findings were left unaddressed.

### Round 1 — `fix(traffic): address ultra-review findings`

- **Spurious full Xray restarts (HIGH regression).** `RemoveUser` now sends the encoded `"<inboundId>::<email>"`, but the "User not found" already-removed idempotency checks still compared the *logical* email, so the benign case forced a full core restart (dropping every live connection) on each poll. Now matches the bare `"not found"` substring at all 5 sites.
- **Multi-node version skew.** An old-build node reports no Billed (decodes to 0); the node→master fold now falls back to Real (a neutral 1×) so a node-only client's Billed enforcement still fires.
- **Real-vs-Billed display consistency.** Subscription tokens, clients-page exhausted buckets, per-node depleted/active counts, and the `/get/:email` `usedTraffic` field all switched to Billed, so what's shown agrees with when service is actually cut.
- **Node multiplier mirrored on update** (not just first create) — keeps `{MULTIPLIER}` and the list filter fresh.
- **External-inform webhook + dashboard WS** aggregate the per-attachment slice back to one row per email (preserves the existing per-client shape for consumers).
- **`DelInbound`** drops the inbound's `client_inbound_traffics` rows — no orphan leak / id-reuse double-count.

### Round 2 — `fix(traffic): close ultra-review gaps (upgrade double-bill, Real-vs-Billed)`

- **Multi-node upgrade double-bill (HIGH).** `backfillBilledTraffic` seeded Billed on `client_traffics` but not the per-node delta baseline `node_client_traffics`. AutoMigrate adds `billed_*` defaulting to 0 while `up/down` keep their pre-upgrade totals, so the first node-sync folded `deltaBilled = canon.Billed − 0` (the node's entire history) on top of the already-backfilled client row — one-time doubling Billed for every node client and instantly depleting any whose quota lay below it. Now seeds the node baseline too. (`client_global_traffics` needs none: overwritten per push and folded with MAX, so a stale 0 can't lower anything.)
- **Clients page read Real where enforcement reads Billed.** The status buckets (depleted/active/expiring) and the "remaining" sort used `Up+Down`, contradicting the summary badge and the client's actual cut-off state on any non-1× inbound. Switched to Billed. (The usage-range filter and "traffic" sort stay Real — they mean bytes moved; inbound-level depletion stays Real — inbounds have no multiplier.)
- **Subscription surfaces reported Real against a Billed quota.** The Subscription-Userinfo header (raw + clash + json), the public info page, and the aggregate that feeds them now carry/report Billed, so `total − used` equals what enforcement cuts on and they agree with the already-Billed remark tokens.
- **Telegram "deplete soon" detection** used `Total − (Up+Down)`; the two client checks switched to Billed.
- **IP-limit live disconnect** only kicked the one inbound resolved by email; for a client on several inbounds the over-limit connection survived on the others (fail2ban still banned the IP fleet-wide). Now force-drops the client on **every** inbound it is attached to.
- **Inbounds-list multiplier filter** could blank the list with no way to clear once the last non-1× inbound became 1×; the predicate now only applies while its control is shown.

## Testing

- **Go:** per-attachment accrual (the 115 GB example), non-retroactive billing on a mid-cycle multiplier change, Billed-based depletion, reset clearing Billed + attachment rows, node→master Billed fold, the upgrade node-baseline backfill (seed / idempotent / reset), Billed-based clients-page buckets + remaining sort, per-email sub aggregation carrying Billed, `getInboundsByEmail` returning every attachment, identity encode/decode round-trip (incl. an embedded `::` and legacy passthrough), `{MULTIPLIER}` token rendering. `go build ./...` and the full `go test -race -shuffle=on ./internal/...` suite pass.
- **Frontend:** `npm ci`, `gen` (no codegen diff), `typecheck`, `lint`, `test` (572 pass), `build` — all green.
- **Locales:** all 13 translation files validated as JSON with full key parity (multiplier + billed labels present everywhere).

## Upgrade notes

- One-time Xray stats rebaseline on first boot (the metered email string changes); no Real is lost and aggregates persist. Billed backfills to Real on both `client_traffics` and the per-node baseline since every existing inbound was effectively 1×.
- Nodes should be updated to a build with this change to bill their inbounds at the configured multiplier; until then a node's inbounds bill at 1× (and old nodes degrade safely).

## Still open (follow-up, not a bug)

- A per-tunnel breakdown **drill-down UI**: `client_inbound_traffics` already persists per-attachment Real + Billed, but the client modal currently surfaces only per-client totals. A read endpoint + modal table would expose the split.


MHSanaei:
no

nima1024m:
Hi @MHSanaei, no problem at all — it's your call as maintainer, and I respect that. 🙏

Would you mind sharing a little of the reasoning, though? I put a lot of care into this one — exact per-attachment metering, multi-node propagation, full `-race` test coverage, localization across all 13 locales, and a couple of self-review passes — so I'd genuinely like to understand the "no" rather than guess at it.

A few things would help me learn from it:
- Is it the **feature direction** (a traffic multiplier just isn't something you want in the panel)?
- A **scope/complexity** concern (the surface area, or the per-attachment metering change)?
- Or something about the **approach** I could rework?

I'm happy to trim it down, split it into smaller PRs, or adjust the design if any form of it would be welcome — and if it's simply not a direction you want to take, that's completely fine too. Either way, thanks for the time you put into the project. 🙂

PR #5508, opened 2026-06-23
Remove force-restart on client disable — sync stored config instead

SKITTLE6969:
```text
refactor(job): replace Xray force-restart with config sync to preserve active connections

After `disableInvalidClients` removes over-quota or expired users via
the gRPC RemoveUser API, the runtime state already matches the desired
state. Calling `RestartXray(true)` forced an unnecessary restart that
dropped active connections.

This replaces the force-restart in both the traffic timer and multi-node
sync paths with a lightweight memory config sync:

```go
if proc := service.XrayProcess(); proc != nil {
    if newCfg, cfgErr := j.xrayService.GetXrayConfig(); cfgErr == nil {
        proc.SetConfig(newCfg)
    }
}

```

By syncing the stored config in-memory without calling the Xray API,
the running instance remains untouched, and future hot diffs
(`tryHotApply`/`ComputeHotDiff`) can accurately track state changes.

---

## Summary

Replaces an aggressive Xray force-restart with a lightweight in-memory configuration sync after invalid clients are removed at runtime.

## Why

Calling `RestartXray(true)` after runtime user removal dropped all active connections unnecessarily, as the running instance was already in the desired state. This preserves active connections and ensures future hot-diff checks are accurate.

## Type of change

* [x] Refactoring (no behavior change)

## Areas affected

* [x] Xray config generation
* [x] Statistics / traffic counters
* [x] Multi-node (sub-nodes)

## Breaking changes

None

## Files changed

* internal/web/job/xray_traffic_job.go
* internal/web/job/node_traffic_sync_job.go

```

```


MHSanaei:
Test result — the client is NOT disconnected when it ends

PR #5524, opened 2026-06-23
feat(clients): bulk-set XTLS flow from the Adjust dialog

rqzbeh:
## Summary

Add a "Set flow" dropdown to the bulk **Adjust** dialog so an admin can set or clear the XTLS flow on all selected clients at once, alongside the existing days/traffic adjustments.

## Why

Flow can already be set per client in the credential tab, but there was no way to change it for many clients at once. Admins migrating inbounds to XTLS Vision (or away from it) had to edit clients one by one. This reuses the existing bulk-adjust path to make it a single action.

The dropdown is empty by default (saving has no effect on flow). Options:
- **No change** (default)
- **Disable** — clears flow on every selected client
- **xtls-rprx-vision**
- **xtls-rprx-vision-udp443**

## Type of change

- [x] New feature

## Areas affected

- [x] Xray config generation

## How was this tested?

- New backend tests in `internal/web/service/client_bulk_flow_test.go`:
  - vision flow set on an eligible VLESS inbound, then cleared via "Disable"; both request a restart; re-setting the same value is a no-op (no restart).
  - vision flow refused on an ineligible inbound (ws transport) — reported as skipped, flow left untouched.
  - empty / unknown directive with no days/traffic errors as "no adjustment".
- `go test ./internal/web/service ./internal/web/controller` — all pass (363).
- `npm run lint`, `npm run typecheck` pass; `npm run gen` produces no diff (codegen gate clean).
- Manual: Clients page → select clients → More → Adjust → pick a flow → Apply; flow column updates, ineligible clients reported skipped.

## Breaking changes

None. Flow rides the existing inbound-settings → `SyncInbound` path (`ClientRecord.Flow` + `client_inbounds.flow_override`); no new endpoint, DB column, or migration. The `/panel/api/clients/bulkAdjust` request gains an optional `flow` field (omitted = unchanged). Setting a vision flow is gated by `inboundCanEnableTlsFlow`: ineligible inbounds are left untouched and reported as skipped; clearing is always allowed. A real flow change requests an xray restart (local) or a node reconcile (remote).

## Checklist

- [x] For frontend changes: `npm run lint`, `npm run typecheck` pass.


rqzbeh:
Thanks for the review — addressed in 99e9978.

**Blocking (store divergence):** flow ineligibility is now tracked in a dedicated `bulkInboundAdjustResult.flowIneligible` map instead of `perEmailSkipped`. Since it no longer enters `skippedReasons`, it can't suppress the `ClientTraffic` write — so a client adjusted with both a days/traffic delta and a flow directive on an ineligible inbound now persists expiry/total to all three stores (inbound JSON, `ClientRecord`, `ClientTraffic`) and is counted as adjusted, while the unhonored flow is appended to `Skipped` separately. Added `TestBulkAdjust_DaysApplyDespiteIneligibleFlow` which asserts `ClientRecord` and `ClientTraffic` advance in lockstep.

**Broad delete:** removed. Flow reasons never enter `skippedReasons` anymore, so honoring a flow can no longer erase an unrelated reason (unlimited expiry, a real persistence error on another inbound). The flow-skip line is added at the end only when the flow was honored nowhere *and* there's no other reason for that client.

**Inline comment:** dropped from `ClientBulkAdjustModal.tsx`; the whitelist-sync note now lives next to `bulkFlowAllowed` in `client_bulk.go`.

**API docs (non-blocking):** added the optional `flow` field to the `bulkAdjust` example in `endpoints.ts` and regenerated `openapi.json`.

i18n keys remain en-US only (other locales fall back via `fallbackLng`), matching the existing pattern.

PR #5599, opened 2026-06-27
fix(settings): normalize API token timestamps

Tomilla:
## Summary

Fix incorrect API-token creation dates in Panel Settings → Authentication → API Token. API-token timestamps now consistently use Unix seconds while remaining compatible with legacy millisecond values.

## Why

The API contract and frontend expected `ApiToken.createdAt` in Unix seconds, but the database model used `autoCreateTime:milli`.

For example, the backend returned:

`1782485394270`

The frontend multiplied this value by 1,000 again, producing a date thousands of years in the future instead of:

`06/26/2026, 22:49:54`

This PR restores the documented seconds-based contract and safely handles data created by affected versions.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Xray config generation
- [x] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot

### Automated validation

The following commands passed locally:

- `go build ./...`
- `go test ./...`
- `npm run lint`
- `npm run typecheck`
- `npm run test` — 40 test files and 589 tests passed
- `npm run build`

### Added regression coverage

- Added a frontend test covering both:
  - Unix seconds: `1782485394`
  - Legacy Unix milliseconds: `1782485394270`
- Added a backend unit test for API response normalization.
- Added a database migration test confirming that:
  - Legacy millisecond values are converted to seconds.
  - Existing seconds values remain unchanged.
  - Running the migration multiple times is safe.

### Real database test

Tested locally with a copy of an existing `x-ui.db` containing:

`created_at = 1782485394270`

After backend startup, the copied database contained:

`created_at = 1782485394`

The authenticated endpoint:

`GET /panel/api/setting/apiTokens`

returned:

```json
{
  "success": true,
  "msg": "",
  "obj": [
    {
      "id": 2,
      "name": "central-panel",
      "enabled": true,
      "createdAt": 1782485394
    }
  ]
}
```

The frontend was tested through the Vite development server in a Chromium-based browser.

On Panel Settings → Authentication → API Token, the token rendered as:

`central-panel — 06/26/2026, 22:49:54`

The original database was not modified.

## Screenshots / recordings

Screenshot of  Authentication → API Token Datetime Bug:
image

image

Fixed:
image

## Breaking changes

None expected.

- Newly created API-token timestamps are stored in Unix seconds.
- Existing millisecond values are automatically migrated to seconds during startup.
- The migration is idempotent.
- The backend also normalizes remaining legacy millisecond values before returning them.
- The frontend accepts both seconds and milliseconds for compatibility with older panel versions.
- The existing API documentation already defines `createdAt` using Unix seconds.

## Checklist

- [x] I verified the Wiki / README / API docs; the existing API contract already documents seconds.

### DIFF
--- a/frontend/src/pages/settings/SecurityTab.tsx
+++ b/frontend/src/pages/settings/SecurityTab.tsx
@@ -13,7 +13,7 @@ import {
   message,
 } from 'antd';
 import { ApiOutlined, SafetyOutlined, UserOutlined } from '@ant-design/icons';
-import { ClipboardManager, HttpUtil, RandomUtil } from '@/utils';
+import { ClipboardManager, HttpUtil, IntlUtil, RandomUtil } from '@/utils';
 import type { AllSetting } from '@/models/setting';
 import { SettingListItem } from '@/components/ui';
 import { useMediaQuery } from '@/hooks/useMediaQuery';
@@ -37,6 +37,12 @@ interface ApiTokenRow {
 interface SecurityTabProps {
   allSetting: AllSetting;
   updateSetting: (patch: Partial<AllSetting>) => void;
+}
+
+const UNIX_MILLISECONDS_THRESHOLD = 100_000_000_000;
+
+function apiTokenCreatedAtMilliseconds(createdAt: number): number {
+  return createdAt < UNIX_MILLISECONDS_THRESHOLD ? createdAt * 1000 : createdAt;
 }
 
 type TfaType = 'set' | 'confirm';
@@ -194,7 +200,7 @@ export default function SecurityTab({ allSetting, updateSetting }: SecurityTabPr
 
   function formatTokenDate(ts: number): string {
     if (!ts) return '';
-    return new Date(ts * 1000).toLocaleString();
+    return IntlUtil.formatDate(apiTokenCreatedAtMilliseconds(ts));
   }
 
   function toggleTwoFactor() {
--- a/frontend/src/test/api-token-date.test.tsx
+++ b/frontend/src/test/api-token-date.test.tsx
@@ -1,0 +1,36 @@
+import { fireEvent, render, screen } from '@testing-library/react';
+import { describe, expect, it, vi } from 'vitest';
+
+import type { AllSetting } from '@/models/setting';
+import SecurityTab from '@/pages/settings/SecurityTab';
+import { HttpUtil } from '@/utils';
+
+describe('API token creation date', () => {
+  it('renders both API seconds and legacy millisecond timestamps', async () => {
+    vi.spyOn(HttpUtil, 'get').mockResolvedValueOnce({
+      success: true,
+      msg: '',
+      obj: [
+        {
+          id: 2,
+          name: 'seconds-token',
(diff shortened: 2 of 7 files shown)


Tomilla:
> **Review summary**
> 
> The fix is correct, focused, and well covered by tests. The root cause is accurately diagnosed: `ApiToken.CreatedAt` used `autoCreateTime:milli` while the API contract, the `example:"1736000000"` tag on `ApiTokenView`, and the frontend all assumed Unix seconds. Switching the model to `autoCreateTime`, normalizing legacy millisecond rows at startup, and defensively normalizing in `toView` together restore the seconds-based contract. The `100_000_000_000` threshold cleanly separates seconds (below it until year ~5138) from milliseconds (above it since ~1973), so the conversion is unambiguous for any realistic timestamp.
> 
> **Confirmed in the source**
> 
> * `toView` is the only consumer of `ApiToken.CreatedAt` (`internal/web/service/panel/api_token.go:30`), so no other caller breaks.
> * `normalizeApiTokenCreatedAtSeconds` runs inside `initModels`, which `InitDB` always calls (`internal/database/db.go:1073`), after `ApiToken` is auto-migrated. It is idempotent: after one pass values fall below the threshold and the `WHERE created_at >= threshold` clause matches nothing.
> * Integer division in `gorm.Expr("created_at / ?", 1000)` truncates identically on SQLite and PostgreSQL, so the migration is portable.
> * `IntlUtil.formatDate` passes its numeric argument to `new Date(...)`, which expects milliseconds, so `apiTokenCreatedAtMilliseconds` is the correct adapter (`frontend/src/utils/index.ts:926`).
> * `internal/web/dist/` is gitignored (no tracked files), so the source-only frontend change is consistent with the build-at-release convention, and no OpenAPI source change is required since no route was added.
> 
> No blocking issues. The findings below are optional.
> 
> **Suggestions**
> 
> * The `100_000_000_000` threshold and the seconds/milliseconds adapter are now duplicated in three places: `internal/database/db.go` (line 1095), `internal/web/service/panel/api_token.go` (line 22), and `frontend/src/pages/settings/SecurityTab.tsx`. Acceptable across the Go/TS boundary, but the two Go copies could share one exported constant to avoid drift.
> 
> **For the maintainer to verify (out of scope for this PR)**
> 
> * Every other `CreatedAt` in `internal/database/model/model.go` still uses `autoCreateTime:milli` (lines 559, 629, 638, 679, 698, 771, 855). Notably `Node.CreatedAt` at line 559 carries `example:"1700000000"` (Unix seconds) while being stored in milliseconds, the same class of mismatch this PR fixes for `ApiToken`. It may be worth checking whether the node UI has an analogous display bug. This PR scope is correctly limited to `ApiToken`.
> 
> _This review was generated automatically; a maintainer may follow up._

fixed. The `100_000_000_000` threshold already defined as one exported constant and shared by other go source files.

PR #5609, opened 2026-06-27
fix(web): don't let an empty skip-suffix disable the request body cap

n0ctal:
## Summary
`MaxBodyBytes(limit, skipSuffixes...)` exempts a request when its path ends in one of `skipSuffixes`. An empty string in that list makes `strings.HasSuffix(path, "")` always true, so a single empty suffix silently disables the body-size cap for **every** route. This adds a `suffix != ""` guard and tightens the matcher (`hasAnySuffix`).

## Why
The body cap protects state-changing routes from unbounded request bodies (memory/DoS). An accidental empty entry in the skip list (e.g. from a config/format slip) removes that protection globally and silently.

## Scope
- `internal/web/middleware/bodylimit.go`: skip a path only when a **non-empty** configured suffix matches.
- `internal/web/middleware/bodylimit_test.go`: assert a sub-path of an exempt route (`/importDB/other`) is **not** exempt and that over-limit bodies on non-exempt paths are still capped.
- `internal/web/web.go`: minor call-site cleanup.

## Validation
`go test ./internal/web/middleware` passes (incl. the new sub-path/empty-suffix cases).

## Risk
Minimal. Exempt routes still work; only the unintended global bypass is closed.


n0ctal:

PR #5610, opened 2026-06-27
fix(settings): require re-2FA confirmation for sensitive setting changes

n0ctal:
## Summary
When two-factor auth is enabled, changing sensitive settings did not require a fresh 2FA confirmation. This adds a server-side re-2FA check on sensitive setting mutations, with the corresponding Security-tab UX.

## Why
A hijacked session (or an unattended logged-in panel) could alter security-relevant settings without re-asserting the second factor. Gating sensitive changes on a fresh 2FA code restores the second-factor guarantee for the actions that matter most.

## Scope
- `internal/web/service/setting.go` + `internal/web/controller/setting.go`: server-side 2FA requirement on sensitive setting changes (server-side enforced, not just UI).
- `frontend/src/pages/settings/SecurityTab.tsx` (+ SettingsPage / useAllSettings): prompt for the 2FA code on those changes.
- `internal/web/service/setting_security_test.go`: coverage.

## Validation
`go test ./internal/web/service` passes.

## Risk
Low. Only adds a confirmation gate; non-sensitive settings and 2FA-disabled setups are unaffected.


n0ctal:

PR #5611, opened 2026-06-27
fix(sync): mark node dirty inside the mutation transaction (atomic ConfigDirty)

n0ctal:
## Summary
`NodeService.MarkNodeDirty` is currently called AFTER a node-affecting mutation, on a fresh `database.GetDB()` handle outside the mutation's transaction. This adds `MarkNodeDirtyTx(tx, id)` and calls it inside each mutation's transaction so the `config_dirty` mark commits atomically with the change.

## Why
The dirty flag drives `ReconcileNode`. Because the mark is a separate post-commit write, a crash/error between the committed mutation and `MarkNodeDirty` leaves a committed config change that is never marked dirty → never reconciled → **silent node drift**. Committing the mark in the same transaction closes that window.

## Behavior change (please read — not a pure no-op)
For `AddInbound`/`DelInbound` the mark was already conditional on a pre-commit decision and is moved unchanged into the tx. **But** at the sites where the live runtime push happens *after* the transaction (`client_bulk.go`, `client_inbound_apply.go`, `SetInboundEnable`, `UpdateInbound`, `resetClientTrafficLocked`), the dirty mark changes from **conditional** (previously set only when the incremental push could not converge the node) to **unconditional** (set whenever `NodeID != nil`).

This is inherent to durable-first: the mark must commit *with* the change, before the post-commit push outcome is known, so it cannot depend on that outcome. Effect: each node-bound mutation now guarantees a full `ReconcileNode` on the next sync tick even when the incremental gRPC push already converged. Because dirty is a boolean it coalesces to **≤1 extra reconcile per node per sync interval**. The tradeoff is intentional: guaranteed drift-freedom over the incremental fast-path.

Optional follow-up (not in this PR): clear the dirty flag after a confirmed successful post-commit push to restore the fast path while keeping the in-tx mark as the durability backstop.

## Scope
- `node.go`: `MarkNodeDirtyTx(tx,id)` (nil-tx guarded); `MarkNodeDirty` delegates on `GetDB()`.
- `client_bulk.go`/`client_inbound_apply.go`/`inbound.go`/`inbound_traffic.go`: `MarkNodeDirtyTx(tx,...)` inside the mutation tx.
- `node_dirty_test.go`: dirty flag does not escape a rolled-back tx; persists on commit; `ClearNodeDirty` CAS unchanged.

## Validation
`go test -race ./internal/web/service` (rollback-atomicity test included). Post-commit reconcile/`ClearNodeDirty` path untouched.

## Risk
No correctness/data-loss change — the system still converges and atomicity improves. The only behavior delta is the added (coalesced) reconcile load described above.


n0ctal:

PR #5613, opened 2026-06-27
fix(nodes): make node API tokens write-only

n0ctal:
## Summary

- Make node API tokens write-only in node API responses.
- Preserve backwards-compatible update behavior: omitted or blank `apiToken` keeps the stored token.
- Add explicit `clearApiToken` for intentional token removal.
- Regenerate API schemas/types and the OpenAPI document for the new `NodeView` / `NodeMutationRequest` contract.

## Why

Returning a node's API token in list/get/add responses leaks a credential that can control the remote node API. The previous draft fixed the leak but made blank `apiToken` an error on update, which broke the existing UI pattern where an empty token field is submitted for an existing node.

This revision keeps the security invariant while preserving the old edit/test flow.

## Scope

- `model.Node.ApiToken` is excluded from JSON/form serialization.
- Node read endpoints return `NodeView`, which exposes only `hasApiToken`.
- Node write/probe endpoints accept `NodeMutationRequest`.
- On update, `apiToken` omitted or blank means keep the stored token.
- `clearApiToken=true` explicitly clears the token.
- Enabled non-mTLS nodes still require a stored or supplied token.
- OpenAPI/generated frontend types now include `NodeView`, `NodeMutationRequest`, `hasApiToken`, and `clearApiToken`.

## Validation

- `go test ./internal/web/service ./internal/web/controller -run 'NodeCredential|NodeController.*Token|NodeUpdateRequiresToken|NodeCredentialsNeverMarshal|NodeViewExposesOnlyCredentialPresence|CheckAPIAuth' -count=1 -v`
- `go vet ./internal/web/service ./internal/web/controller ./tools/openapigen`
- `go build ./...`
- `npm run gen`
- `npm run build`
- `npm run typecheck`
- `git diff --check`

`make gen` / `make verify` are not present in this upstream tree; the equivalent project scripts above were used.

## Risk

Low. Existing API clients that send an empty token while editing a node continue to work and keep the stored token. Clients that intentionally need to clear the token must use `clearApiToken=true`; clearing is rejected for enabled non-mTLS nodes unless another authentication method is configured.

The response shape changes from `apiToken` to `hasApiToken`. That is the intended security boundary: credentials are accepted only on write and are never returned on read.


n0ctal:

PR #5614, opened 2026-06-27
fix(runtime): refresh cached node remotes on identity change

n0ctal:
## Summary

`Manager.RemoteFor` caches one `*Remote` per node id and returns the cached instance on every later call, keyed only on `node.Id`. If a node's connection identity changes (address, port, base path, scheme, API token, TLS verify mode, pinned cert, egress/outbound tag, allow-private), the cached `Remote` keeps using the **stale** identity until the process restarts or the entry is explicitly invalidated. This PR makes the cache converge: a cached remote whose identity no longer matches the requested node is transparently replaced.

## Why

A node's identity can change at runtime (token rotation, address/port move, switching TLS verify mode, repinning a cert). With the id-only cache, the master can keep talking to a worker with an old token or old address after the operator has already changed it — silently, with no error, until a restart. That is both a correctness bug (config changes don't take effect) and a security gap (a rotated-out token stays live in the cached path). Convergence on identity change closes both.

## Scope

`internal/web/runtime/manager.go`:

- `RemoteFor` now compares the cached remote's identity against the requested node via `sameRemoteIdentity`. On a match it returns the cached instance (fast path under `RLock`); on a mismatch it rebuilds the remote under the write lock and replaces the map entry. Standard double-checked locking — no network I/O is performed under the lock.
- `sameRemoteIdentity` compares exactly the fields that affect how we connect/authenticate: `Id, Scheme, Address, Port, BasePath, ApiToken, AllowPrivateAddress, TlsVerifyMode, PinnedCertSha256, OutboundTag`. Display/metrics fields are intentionally excluded so cosmetic edits don't churn the cache.
- `cloneRemoteNode` stores a defensive copy of the node in the cache (deep-copying `InboundTags`) so later mutation of the caller's `*Node` cannot retroactively change the cached identity and mask a real divergence.

Tests: `manager_convergence_test.go` covers same-identity cache hit (no rebuild), each identity-field change forcing a rebuild, and the no-churn case for display fields.

## Validation

- `go test ./internal/web/runtime -run 'Convergence|RemoteFor|Identity|Refresh'` — green
- `go test -race ./internal/web/runtime -run 'Convergence|RemoteFor|Identity|Refresh'` — green
- `go vet ./internal/web/runtime` — clean

## Risk

Low. Behavior is unchanged when a node's identity is unchanged (same cached instance returned). The only new behavior is that an identity change now produces a fresh remote instead of reusing a stale one; any in-flight request on the old instance completes normally and the old remote is GC'd. No public API or schema change.


n0ctal:

PR #5624, opened 2026-06-27
Fix WireGuard client binding in client manager

fastnas2023:
## Summary
- Fix WireGuard client attachment so client bindings map to `settings.peers[]` instead of unsupported `settings.clients[]`.
- Generate/reuse WireGuard peers for bound clients and filter inactive/expired/depleted/detached managed peers from runtime Xray config.
- Allow WireGuard in client inbound pickers and improve selected inbound tag layout.
- Update OpenAPI docs and add focused service tests for WireGuard client binding behavior.

## Verification
- `go test ./internal/web/service`
- `go test ./...`
- `npm --prefix frontend run build`
- `git diff --check`

### DIFF
--- a/frontend/public/openapi.json
+++ b/frontend/public/openapi.json
@@ -2301,7 +2301,7 @@
     },
     {
       "name": "Inbounds",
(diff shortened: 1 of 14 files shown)


MHSanaei:
no need 9c8cd08f90205d449427705e4483e3eb3b790223

PR #5623, opened 2026-06-27
feat(wireguard): support Xray dynamic peer runtime

mrAntonD:
## Summary

This updates WireGuard client handling to align with Xray-core dynamic peer management introduced in Xray-core v26.6.27.

Related context:
- Previous PR: #5607
- Xray-core change: https://github.com/XTLS/Xray-core/commit/345c76f9a8ec438e24bbbc5a246bd0bfeab480b1
- Related PR: #5620

## What Changed

- Update the xray-core dependency to the commit that includes WireGuard dynamic peer management.
- Add WireGuard support to runtime AddUser by building a WireGuard peer account from publicKey, allowedIPs, preSharedKey, keepAlive, and email.
- For local Xray runtime, WireGuard client add/update/delete now first tries dynamic AddUser/RemoveUser.
- If dynamic runtime update fails, the code falls back to the existing full inbound update path.
- Persist email on WireGuard peers while keeping comment for backward compatibility.
- Keep old peer migration behavior: existing comment values still work.

## Why

Xray-core v26.6.27 changed WireGuard server-mode peers into users internally and added dynamic peer management. This PR keeps the existing 3x-ui WireGuard client model, while allowing new Xray versions to apply peer changes without rebuilding the whole inbound.

## Compatibility

- Non-WireGuard protocols should not change behavior.
- New Xray versions use dynamic peer operations where possible.
- Older or unsupported runtimes fall back to the previous inbound update behavior.
- Existing peer configs using comment remain supported.

## Validation

Automated checks:

- `go test ./internal/...`
- `go test ./...`
- `git diff --check`

Runtime smoke test:

- Built frontend with Node 22: `cd frontend && npm run build`
- Built panel binary: `go build -o x-ui-build main.go`
- Deployed to a Linux amd64 systemd test deployment running Xray `v26.6.27`
- Restarted `x-ui` successfully
- Verified `x-ui` service is active
- Verified Xray process starts successfully
- Verified Xray reports version `26.6.27`
- Verified panel and subscription HTTP ports are listening
- Verified no startup errors after restart

PR #5631, opened 2026-06-28
fix(wireguard): address review feedback

mrAntonD:
Follow-up to MHSanaei/3x-ui#5623.

This PR addresses the review feedback from the closed WireGuard clients PR and keeps the same feature direction: WireGuard peers are managed as first-class clients while `settings.peers[]` remains the runtime projection for Xray.

## What changed

- Preserve WG client comments across panel restarts by storing `email` separately from the real peer `comment`.
- Keep disabled WG clients attached to their inbound during startup reconciliation, so disabling stays reversible after restart.
- Rebuild WG `settings.peers[]` from linked clients on inbound updates, runtime API updates, and cold-start config generation instead of trusting stale inbound form snapshots.
- Merge `wgSettings` during client record merge, with redacted conflict reporting.
- Complete the new WG i18n keys across all 13 locale files and route WG UI strings/tab labels through `t()`.
- Derive the next WG peer IP from existing peer addresses instead of assuming only `10.0.0.0/24`.
- Fix the inbound list WG active-client counter by reading `peer.email` with legacy fallback to `peer.comment`.

## Validation

- `go test ./... -timeout=300s`
- `npm run typecheck`
- `npm run build` with Node `22.23.0`
- `git diff --check`

## Notes

The previous PR review threads are still unresolved on GitHub because MHSanaei/3x-ui#5623 is closed, but each actionable item from that review is covered here.


MHSanaei:
Thanks a lot for the work here, @mrAntonD — and for the careful review-feedback pass.

Closing this as **superseded**. WireGuard multi-client support landed on `main` in 9c8cd08f via a different architecture than #5623 (which this PR builds on), so most of these fixes are now either already handled or no longer applicable:

- **Canonical storage**: clients live in `settings.clients` (like every other protocol) and are projected to `peers[]` only when emitting the Xray config, at level 0. So `settings.peers[]` is rebuilt from clients on every config generation — no stale snapshot to reconcile.
- **`ReconcileWgPeers` (disabled peers across restart)**: not needed. A disabled client stays in `settings.clients` and is simply omitted from the projected peers, so enable/disable is fully reversible after a restart.
- **Partial-update preservation**: `UpdateInboundClient` already carries forward stored keys/allowedIPs/PSK/keepalive when the payload omits them (keys are never rotated on edit).
- **Comment preservation**: `comment` is a first-class client field now, not stored in the peer object.
- **Node-sync merge**: WireGuard fields are merged as first-class columns in `MergeClientRecord`.
- **i18n / counters**: the 13 locales were added with the feature, and the active-client counter is fully client-centric.

The one genuinely useful idea — **deriving the next peer IP from the existing peer subnet instead of assuming `10.0.0.0/24`** — has been ported to `main` in commit 79069d2b (with tests), so a migrated inbound whose peers live in a non-default subnet now allocates new clients in the right range.

Closing in favor of the native-users implementation. Thanks again! 🙏

PR #5639, opened 2026-06-28
Fix WireGuard client binding and external link toggles

fastnas2023:
## Summary
- Fix WireGuard client attachment so client bindings map to `settings.peers[]` instead of unsupported `settings.clients[]`.
- Generate/reuse WireGuard peers for bound clients and filter inactive/expired/depleted/detached managed peers from runtime Xray config.
- Allow WireGuard in client inbound pickers and improve selected inbound tag layout.
- Add per-row enable toggles for client external links/subscriptions; disabled rows remain saved but are omitted from generated subscriptions.
- Update OpenAPI docs and add focused service/subscription tests.

## Verification
- `go test ./internal/web/service ./internal/sub`
- `go test ./...`
- `npm --prefix frontend run build`
- `git diff --check`


MHSanaei:
Thanks for the work here! WireGuard multi-client is already supported on `main` via the canonical `settings.clients[]` store, which is projected to `settings.peers[]` at Xray config-generation time — so the client-binding rework in this PR conflicts with the current architecture (and the PR itself is now showing merge conflicts). Closing this one.

The per-row enable/disable toggle for external links/subscriptions is a nice idea though — if you'd like to open a small standalone PR just for that, it would be very welcome. 🙏

PR #5607, opened 2026-06-27
feat(wireguard): manage peers as clients

mrAntonD:
Replaces #5466.

Updated on top of latest main. Merge conflicts resolved.

Checks:
- `npm run typecheck`
- `go test ./...`

PR #5466, opened 2026-06-21
feat(wireguard): manage WireGuard peers as clients on the Clients tab

mrAntonD:
## Why

WireGuard peers were second-class citizens — managed only inside the Inbounds form alongside transport settings. This meant no per-peer traffic limits, no expiry, no sub IDs, and no per-client QR/config download. Moving peers to the Clients tab gives WireGuard the same UX as any other protocol while keeping the Inbounds form clean.

## Screenshot

WireGuard clients on the Clients tab

## Summary

- **WireGuard peer management moved to Clients tab**: peers are now created/edited/deleted from the Clients page like any other protocol; the Inbounds form retains WG settings (key, MTU, etc.) but the peer list is removed from it
- **Auto-migration on upgrade**: existing `settings.peers[]` are migrated to the `clients` table on first startup — no data loss
- **Peer count on Inbounds tab**: WireGuard inbounds now show active peer count
- **Peers stay in sync on every restart**: `ReconcileWgPeers` runs on each panel start, so peers added directly to `settings.peers[]` always appear in the Clients tab
- **Duplicate IP prevention**: entering an already-used `allowedIPs` in the client form auto-assigns the first available IP instead; existing clients are not affected when editing

### Backend
- `model.go` — `WgPeerSettings` struct + `wg_settings` column on `ClientRecord`; `MarshalJSON`/`UnmarshalJSON` expose it as `wgPeer` in API responses
- `client_wg_apply.go` — `AddWgClient`, `UpdateWgClient`, `DelWgClient`: full CRUD; each mutation patches `settings.peers[]` in-place (no full DB rebuild)
- `client_wg_peers.go` — `syncWgPeersFromClients`, `wgPeerToRecord`, `SyncWgInbound`
- `inbound_migration.go` — `ReconcileWgPeers` called on every panel start to keep `settings.peers[]` and `clients` table in sync
- `inbound.go` — `UpdateInbound` preserves `settings.peers[]` when the form omits them; `inboundWgPublicKey` derives server public key from `secretKey` via curve25519; `GetInboundOptions` exposes `wgPublicKey`
- `client_crud.go` — `Delete()` routes WG inbounds through `DelWgClient`
- `client_paging.go` — `ClientSlim` includes `wgPeer` so next-IP suggestion works from the paged list
- `inbound.go` controller — WG CRUD endpoints under `/panel/api/inbounds/:id/`

### Frontend
[...]

PR #5642, opened 2026-06-28
fix(wireguard): restore client config UX

mrAntonD:
## Summary

This PR restores several WireGuard client UX pieces on top of the upstream multi-client implementation:

- show an always-visible active client count beside the total client count for inbounds, including `0` active clients;
- add a direct `settings.clients[]` fallback in the inbound list so WireGuard counts render even when the async rollup has not populated `clientCount` yet;
- show the WireGuard inbound server public key in the inbound info modal by deriving it from `secretKey` when needed;
- expose WireGuard `.conf` client config in the client create/edit form, client info modal, and client QR modal, with copy/download/QR actions;
- keep `wireguard://` peer links copy-only instead of rendering QR codes that are not useful for standard WireGuard clients;
- include lightweight WireGuard hints (`wgPublicKey`, `wgMtu`) in inbound options so client views can render complete configs.

## Why

After testing the current upstream WireGuard multi-client flow, the data model works, but several UX pieces from the earlier implementation were missing. This made it hard to inspect/copy client configs and caused peer-link QR codes to appear even though they do not lead to a usable WireGuard import.

This is a focused follow-up to the earlier WireGuard client work and review iterations:

- #5631
- #5623
- #5607
- #5466

## Validation

- `npm run build` from `frontend` using Node 22
- `go test ./internal/web/service`
- deployed and smoke-tested on a test NL server: panel/sub ports active, Xray started, WireGuard client UX checked manually

### DIFF
--- a/frontend/public/openapi.json
+++ b/frontend/public/openapi.json
@@ -1854,6 +1854,12 @@
           "tlsFlowCapable": {
             "example": true,
             "type": "boolean"
+          },
+          "wgMtu": {
+            "type": "integer"
+          },
+          "wgPublicKey": {
+            "type": "string"
           }
         },
         "required": [
@@ -2751,7 +2757,9 @@
                       "remark": "VLESS-443",
                       "ssMethod": "",
                       "tag": "in-443-tcp",
-                      "tlsFlowCapable": true
+                      "tlsFlowCapable": true,
+                      "wgMtu": 0,
+                      "wgPublicKey": ""
                     }
                   ]
                 }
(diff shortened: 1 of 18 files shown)


mrAntonD:
Addressed the review feedback in `5051cbea`:

- regenerated OpenAPI/Zod/frontend generated artifacts for `wgPublicKey` and `wgMtu`;
- fixed the frontend typecheck issue in the WireGuard client-count fallback;
- made client WireGuard `.conf` generation use the configured public host consistently with inbound QR/config generation;
- moved the WireGuard config labels through i18n.

All CI checks are now passing, including `codegen` and `frontend`.

MHSanaei:
give me access to edit this PR

MHSanaei:
don't open any pr without edit access

Issue #5486, opened 2026-06-22
[Feature]: Suggestion: Improving TalkBack (Accessibility) Compatibility for Visually Impaired Users


alinasrollahzadeh405-sudo:
### Is your feature request related to a problem?

I am a visually impaired user who relies on Android's TalkBack screen reader to manage my servers. Currently, navigating the X-UI web interface is very difficult because many buttons, input fields, and UI elements lack proper accessibility labels and semantic HTML, making it nearly impossible for TalkBack to read them correctly.

### Describe the solution you'd like

I would appreciate it if the UI could be made more accessible by adding 'aria-label' attributes to buttons and icons, ensuring input fields have proper 'label' tags, and improving the overall keyboard/screen-reader navigation flow. This would allow visually impaired admins to manage their servers effectively and independently.

### Are you willing to help?

- [ ] I'd like to implement this feature myself and open a PR.
- [ ] I can help test once a PR is open.

### Before submitting

- [x] I searched [existing issues](../issues?q=is%3Aissue) and this feature has not been requested.
- [x] This feature request is written in English.

PR #5652, opened 2026-06-29
feat(a11y): screen-reader & keyboard accessibility across the panel (#5486)

nima1024m:
Implements panel-wide screen-reader (Android TalkBack) and keyboard accessibility for upstream issue MHSanaei/3x-ui#5486.

## What changed
Icon-only buttons across the panel previously announced only their untranslated icon name (e.g. "edit", "ellipsis") or nothing, and several clickable elements (mobile card actions, dashboard cards, drag handles, QR codes, list remove-icons) were not keyboard-operable. This adds proper accessible names and keyboard support across every management surface, **without changing any visual or mouse behavior**.

Built from a panel-wide audit, split into reviewable commits:
- **Lists, toolbars & dashboard** — aria-labels on row-action/toolbar icon buttons (inbounds, clients, groups, hosts, nodes, xray outbounds/routing/dns/balancers) and dashboard cards; sidebar hamburger mislabel fixed.
- **Modals, QR & copy panels** — info modals, QR panel, sub-links, log/backup/version modals, 2FA; click-to-copy QR codes made keyboard-operable.
- **Forms & shared components** — `SettingListItem` now labels the ~90 settings-tab inputs via `aria-labelledby`; `InputAddon` remove controls are keyboard-operable; form add/remove/regenerate buttons and bare inputs labeled; decorative icons `aria-hidden`; charts exposed as labeled images.
- **Shared xray forms** — `FinalMaskForm`, `CustomSockoptList`, `SniffingFields`, and the WireGuard `ConfigBlock` (used by the host/inbound/client forms).
- **Lint harness (dev-only)** — `eslint-plugin-jsx-a11y` (flat-config recommended ruleset) so accessibility regressions fail `npm run lint`. This is a lint-time tool only: it is not bundled and does not affect the built panel — the shipped accessibility is the `aria-*`/`role`/`tabIndex` attributes added to the JSX.

A small shared `activateOnKey` helper (`src/utils/a11y.ts`) provides Enter/Space activation for non-button clickable elements.

> Rebased onto current `main` (includes the WireGuard client-config feature); its restructured `ClientInfoModal` and new `ConfigBlock` are labelled too.

## Internationalization
12 new i18n keys added to **all 13 locale files** with real translations (not English fallbacks): `sort`, `regenerate`, `jsonEditor`, `downloadImage`, `menu.openMenu`, `pages.xray.outbound.moveToTop`, `pages.xray.balancer.{costMatch,costValue,costRegexp}`, `pages.settings.notifyTime.{interval,unit}`, `pages.inbounds.sniffingDestOverride`. Everything else reuses existing keys. Three previously-hardcoded English outbound menu labels are now translated.

## Dependency note: jsx-a11y on ESLint 10
`eslint-plugin-jsx-a11y@6.10.2` (its latest release) declares a peer range that stops at **ESLint 9**, while this panel is on **ESLint 10**. The plugin itself runs fine on ESLint 10 through flat config (`flatConfigs.recommended`) — the peer range is simply stale — but the version mismatch makes a clean `npm ci` abort with `ERESOLVE`. This is resolved with a single npm `overrides` entry so the plugin accepts the project's ESLint version:

```json
"overrides": {
  "eslint-plugin-jsx-a11y": { "eslint": "$eslint" }
}
```

This is surgical: it only relaxes jsx-a11y's `eslint` peer and leaves normal peer resolution intact, so recharts' `react-is` peer still auto-installs. No `legacy-peer-deps`, no other dependency changes.

## Verification
- `npm run typecheck`, `npm run lint` (jsx-a11y enforced — 0 errors), `npm run build`, `npm run test` (**616/616**), `make gen` (no diff).
- CI on the fork is **green across all jobs** (frontend, multi-arch builds, CodeQL analyze, go-test, golangci, govulncheck, race, codegen).
- Multi-angle code review found **no correctness regressions to normal (non-accessibility) functionality**: the stop-propagation removals, QR onClick relocations, dropdown-trigger conversions and shared-component changes were each traced safe across their call sites.
- Locale completeness verified by script: every referenced `t()` key resolves in en-US; all 12 new keys present as strings in all 13 locales.

## Notes
- All changes are additive (`aria-*` / `role` / `tabIndex` / `onKeyDown`) and do not alter mouse/visual behavior.
- `jsx-a11y/no-autofocus` is disabled deliberately — autofocus on the login field and modal primary inputs is intentional focus management that helps screen-reader and keyboard users land on the right control.
- Possible follow-up (intentionally not done here, to avoid large late churn): extract the repeated `role/tabIndex/aria-label/onClick/onKeyDown` icon pattern (~34 sites) into a small `ClickableIcon` component.


nima1024m:
Thanks for the careful trace — appreciated. Both suggestions are addressed in 7aeca55.

**1. Mobile-card `more` triggers missing `size="small"`** — Fixed in `InboundList`, `ClientsPage`, and `NodeList`. These three were the only `MoreOutlined` dropdown triggers in the panel without it; they now match the adjacent `size="small"` `Switch` and the established desktop `RowActions` pattern, so the converted buttons no longer carry antd's default height/hover box.

**2. `NodeList` `.card-head` `role="button"` wrapping interactive controls** — Fixed rather than deferred, since it's the PR's own code. The disclosure semantics (`role="button"` / `tabIndex` / `aria-expanded` / `aria-label` / keyboard activation) moved off the header and onto the chevron (`RightOutlined`) as a dedicated expand affordance. The header `<div>` keeps its plain mouse `onClick`, so click-anywhere-to-expand is unchanged for mouse users, while the `Switch`, info button, and dropdown are no longer nested inside a `role="button"`. The remaining `<div onClick>` lint flag is suppressed with the same scoped `eslint-disable` convention already used in `ConfigBlock.tsx`, plus an inline justification.

On the two maintainer-check notes:
- **Non-English translations** — verified all 12 new keys resolve as non-empty strings across all 13 locales. The only value identical to English is id-ID `notifyTime.interval` = "Interval", which is a legitimate Indonesian cognate (its sibling `unit` is localized), not a fallback.
- **Mobile `more` button hover/height** — addressed by the `size="small"` change above.

Verification after the change: `npm run typecheck`, `npm run lint` (0 errors), `npm run test` (616/616), and `npm run build` all green.

Issue #5689, opened 2026-06-30
[Feature]: Allow opting out of auto-applied XTLS Vision (xtls-rprx-vision) per inbound/client


FZ1010:
### Is your feature request related to a problem?

There's no way to opt a specific inbound (or a client on a specific inbound) **out** of the auto-applied `xtls-rprx-vision` flow. Recent 3x-ui correctly reports VLESS-over-XHTTP with VLESS-encryption (ML-KEM / "vlessenc") as flow-capable and injects `xtls-rprx-vision` for it, consistent with Xray's migration away from flow-less VLESS. **That default is reasonable and should stay** — the gap is only that there's no per-inbound/per-client way to opt out, so a client attached to two flow-capable inbounds is all-or-nothing.

Whether the flow lands in a generated config is decided by two mirrored gates:

- `inboundCanEnableTlsFlow(protocol, streamSettings, settings)` — `internal/web/service/inbound_protocol.go` (powers `tlsFlowCapable` in `/inbounds/options` and the write-path clamp in `client_crud.go`).
- `vlessFlowAllowed(network, security, settings)` — `internal/sub/service.go`, applied at `genVlessLink` (`if len(flow) > 0 && vlessFlowAllowed(...) { params["flow"] = flow }`).

Both return `true` for `xhttp` + vlessenc, and there's no inbound- or membership-level way to say "not here". The per-`(client, inbound)` value already exists as `client_inbounds.flow_override` and is resolved by `EffectiveFlow` / `EffectiveFlowsByEmails` (`internal/web/service/client_lookup.go`, added for #4792 / #4834) — but it's only ever **derived** from the client's flow clamped by `inboundCanEnableTlsFlow`; callers can't set it independently.

### Describe the solution you'd like

Additive and non-breaking — keep Vision-on-by-default and only add an opt-out. Either of:

1. **Make per-`(client, inbound)` `flow_override` settable** via the client add/update API (the column + `EffectiveFlow` resolver already exist), so a client can carry Vision on some inbounds and not others. Would need a writable path (e.g. `POST /panel/api/clients/:email/flow` with `{inboundId, flow}`) plus a panel control, and the value must survive the existing flow-restore pass.
[...]

PR #5690, opened 2026-06-30
feat(inbound): DisableFlow — opt an inbound out of auto XTLS Vision (#5689)

FZ1010:
Implements **approach (2)** from #5689 — an inbound-level opt-out from automatic XTLS Vision (`xtls-rprx-vision`) injection.

### Why

VLESS-over-XHTTP with vlessenc is (correctly) reported flow-capable, so a client on both a Reality inbound and a tunneled/CDN-fronted XHTTP+vlessenc inbound gets `xtls-rprx-vision` on **both**, with no way to opt the tunnel inbound out (single client-level flow; the panel clamps per its own capability). Some deployments specifically don't want Vision on the tunneled inbound. See #5689 for the full write-up.

### What this does

Adds a panel-only `DisableFlow` flag to `model.Inbound` (`disable_flow` column, GORM AutoMigrate). When set, the inbound behaves as not-flow-capable **without** changing its transport/encryption:

- `GetInboundOptions` reports `tlsFlowCapable=false`.
- The write-path clamp (`clientWithInboundFlow`) stores an empty flow for clients attached to it (so `client_inbounds.flow_override` is `""`).
- `genVlessLink` never emits `flow=` for it (sub/share links stay clean).
- The client move/copy path, the post-edit Vision restore (`restoreVisionFlowForEligibleInbound` callers), the migration restore, and bulk flow-set all honor it.

The flag is never sent to xray (panel metadata only). The default (`false`) preserves all existing behavior — this is purely additive.

### Verification

- `go build ./internal/...` (affected packages) — clean.
- `go test ./internal/sub/...` flow suite passes, including a new `TestGenVlessLink_DisableFlowSuppressesFlow` (a capable xhttp+reality+vlessenc inbound with `DisableFlow=true` emits no `flow=`), alongside the existing positive cases.
- The DB-integration tests under `internal/web/service` weren't run in my local env (`CGO_ENABLED=0`, so `go-sqlite3` can't init) — they'll run in CI.

### Status — draft

- [ ] **Frontend toggle** (React, `frontend/src/pages/inbounds/…`) — follow-up commit; the field is already settable via the inbound add/update API in the meantime.
- [ ] Runtime verification on a live panel (I implemented from source; haven't run a built panel).

Marking draft pending the frontend toggle + a maintainer's nod on placement/naming (`DisableFlow` vs e.g. `NoFlow`). Happy to adjust to fit your conventions. Closes part of #5689.


FZ1010:
Thanks for the thorough review — addressed in a38a1bb:

**Blocking 1 (Clash/JSON still emit the flow).** Added the negated `DisableFlow` check to both sibling emitters — `clash_service.go` (`buildProxy`) and `json_service.go` (`genVless`) — so all three paths agree. Added `TestBuildProxy_VLESSFlowSuppressedByDisableFlow` and `TestSubJsonServiceVlessFlowSuppressedByDisableFlow` next to the existing raw-link test.

**Blocking 2 (stale generated artifacts).** Ran `make gen` — `disableFlow` is now in all four `frontend/src/generated/*` files and `frontend/public/openapi.json`, and added `example:"false"` on the field. `make gen-check` is clean.

**(a) toggle authority.** Good catch — took your suggestion: `UpdateInbound` now strips any flow already stored on a `DisableFlow` inbound's clients (settings `clients[].flow` + `client_inbounds.flow_override`), so xray and the subscription agree after a toggle rather than xray expecting a flow the client no longer sends. The explicit emitter guards stay as defense-in-depth.

**(b) node sync.** Confirmed it works for node-hosted inbounds without propagating the column: node sync carries `settings`/`stream_settings` (`inbound_node.go` ~449/488), the master strips the flow there, and the master generates the subscriptions — so the effect reaches node inbounds via the already-stripped settings. Happy to also propagate the `disable_flow` column if you'd prefer node-side authority.

**(c)** Frontend toggle is included; live-panel verification is the remaining open item.

FZ1010:
Pushed c3a07f6 after a deeper self-review — the first cut had two blocking gaps that made the toggle a no-op on the path that matters for #5689 (editing an existing multi-inbound client's inbound):

**Blocking — DisableFlow was never persisted.** `UpdateInbound` read `inbound.DisableFlow` to branch strip-vs-restore but never copied it onto the `oldInbound` row before `tx.Save`, so the column stayed false. The flow got stripped for that one request, then `MigrationRestoreVisionFlow` (and the next edit) re-injected Vision. Fixed by copying the field in the update block.

**Blocking — the edit Switch couldn't round-trip.** The frontend `DBInbound` model never declared `disableFlow`, and its constructor uses `ObjectUtil.cloneProps`, which only copies keys the instance already owns — so the API value was dropped and the Switch always read false, reverting the opt-out on the next save. Fixed by declaring + initializing the field.

**Hardening:**
- `AddInbound` now strips client flow (settings + parsed clients) when DisableFlow is set, so a created-disabled inbound can't persist a flow xray would expect.
- `GetXrayConfig` forces `flow=""` for DisableFlow inbounds (VLESS + Trojan) as defense-in-depth — the live config and the subscription stay in agreement no matter how a flow_override got stored.
- `genTrojanLink` honors DisableFlow too; dropped a dead explicit `flow_override` clear (SyncInbound rebuilds it); clear `disableFlow` in the form when switching off VLESS; added the two i18n keys to the remaining 12 locales.

**Tests (run locally with CGO/sqlite):** `stripClientFlows` unit cases; a DB-backed `AddInbound` clamp test; and a `UpdateInbound` persist+strip+resist-restore regression that fails without the persist fix. Frontend: a `DBInbound`/adapter round-trip test that fails without the model field. Full `go test ./internal/web/service ./internal/sub` and the frontend vitest suite (621) are green; `gen-check` clean.

FZ1010:
Also pushed e16acf4: removed the `//` line comments I'd added across the Go / frontend / test files to match `CLAUDE.md` (this also drops the generated OpenAPI field descriptions, since those were sourced from the Go doc comments). No behavior change. Verified with `golangci-lint` (rebuilt with go1.26.4 to match the repo's target), full `go test ./internal/web/service ./internal/sub` (CGO sqlite), frontend `typecheck`/`vitest`, and `make gen-check` — all clean.

FZ1010:
Closing to re-open as a fresh PR with a cleaned-up writeup and full `make verify` results. Same branch/commits; superseding PR linked shortly.

FZ1010:
Superseded by #5698 (same branch + commits, fresh writeup with full `make verify` results).

Issue #4792, opened 2026-06-02
[Bug]: VLESS Reality `flow` is not saved or cleared on multi-inbound client (VLESS + Hysteria) still present in v3.2.6


VMMaster:
### Describe the bug

After upgrading to version 3.2.5, a bug appeared when saving a client that is attached to multiple inbounds at the same time: **Hysteria** and **VLESS Reality**. I have just updated to version 3.2.6, but unfortunately, the issue still persists.

### How to reproduce the problem

1. Create or edit a client and assign them to two inbounds simultaneously: one Hysteria and one VLESS Reality.
2. On the main node, select `xtls-rprx-vision` in the **flow** setting.
3. Save the client.
4. Open the client modal/card again.

### Expected behavior

The `flow` parameter should be correctly stored and rendered per-inbound, regardless of whether the client shares other non-VLESS inbounds (like Hysteria), just like it used to work flawlessly in version 3.2.0.

### Actual behavior

- If the client has **only** VLESS Reality, the `flow` setting is preserved correctly upon reopening.
- If the client has **both** Hysteria + VLESS Reality:
  - Opening the client card shows an empty `flow` field (it fails to hydrate the form/UI dropdown).
  - If you have multiple nodes (Main + Additional): the `flow` value is saved on the main node during the initial setup, but it completely fails to save on the additional node. Re-opening the card wipes the flow setting, forcing the admin to re-select it every single time.

### Relevant logs

```shell

```

### Screenshots

When adding only vless reality, everything will be saved well even if the main and additional node are selected

<img width="701" height="929" alt="Image" src="https://github.com/user-attachments/assets/7de869ce-946c-45bc-937d-2376595c3226" />

<img width="702" height="1012" alt="Image" src="https://github.com/user-attachments/assets/b890bdc6-0823-4af3-a5f1-3474e74e6b0f" />

I added all the Vless + Hysteria connections, the primary and secondary nodes, selected flow xtls-rprx-vision, saved it, and went back to the card, but it was empty.

<img width="706" height="1018" alt="Image" src="https://github.com/user-attachments/assets/88da1b6a-1647-48c8-85f5-00aa3825d394" />

<img width="701" height="1004" alt="Image" src="https://github.com/user-attachments/assets/22c6eb0f-858f-4aa5-9d35-f4767722c264" />

<img width="331" height="254" alt="Image" src="https://github.com/user-attachments/assets/3bcc145c-8f8c-459e-810c-d688487acaee" />

[...]

PR #5691, opened 2026-06-30
feat(clients): writable per-(client,inbound) XTLS flow override (#5689)

FZ1010:
Implements **approach (1)** from #5689 — a writable per-`(client, inbound)` XTLS flow, so a client can keep Vision on some flow-capable inbounds and an empty flow on others in the same subscription.

### Why

A client delivered on both a Reality inbound (Vision wanted) and a tunneled XHTTP+vlessenc inbound (Vision **not** wanted) currently gets `xtls-rprx-vision` on both — the single client-level flow is clamped per inbound by capability, but there's no way to clear it on one *capable* inbound while keeping it on another. See #5689.

### What this does

The relational per-membership store already exists (`client_inbounds.flow_override`) and is resolved by `EffectiveFlow` / `EffectiveFlowsByEmails`. This PR makes it **settable**:

- New service method `ClientService.SetInboundClientFlow(inboundSvc, inboundId, email, flow)` — sets the flow for one client on **one** inbound, bypassing the `clientWithInboundFlow` capability clamp. It writes the inbound `settings.clients[].flow` (what subscription generation reads) via the existing `UpdateInboundClient`, then mirrors the value onto `client_inbounds.flow_override`.
- New endpoint `POST …/clients/:email/flow` with body `{ "inboundId": <id>, "flow": "xtls-rprx-vision" | "" }`.
- A pure helper `clientWithFlow` (returns a copy with the flow applied) so the core is unit-testable without a DB.

### Open question for maintainers (the one design decision)

An explicit **empty** flow set here can be re-populated by `restoreVisionFlowForEligibleInbound` (#4792) if the inbound is later edited into a newly flow-eligible state, since that path treats an empty settings flow as "restore the intended Vision from a sibling". Making an explicit clear durable across such edits needs a small marker — e.g. a sentinel `flow_override` value (`"none"`) meaning *explicitly off*, or a per-membership lock flag. I left that out deliberately because it's a model/semantics choice that's yours to make; happy to implement whichever you prefer.

### Verification

- `go build ./internal/web/service/... ./internal/web/controller/...` — clean.
- `go test ./internal/web/service -run TestClientWithFlow` — passes (covers the override-clears-on-one-inbound and missing-email cases).
- Couldn't run the DB-integration paths locally (`CGO_ENABLED=0`) or boot a panel — so this is **draft** pending runtime verification.

### Status — draft

- [ ] Frontend: a per-inbound flow control in the client form (the API endpoint works in the meantime).
- [ ] Maintainer decision on the explicit-clear durability (above).
- [ ] Runtime verification on a live panel.

Alternative to #5690 (the inbound-level `DisableFlow` toggle) — both close part of #5689; pick whichever fits the design.

### DIFF
--- a/frontend/public/openapi.json
+++ b/frontend/public/openapi.json
@@ -5634,6 +5634,64 @@
         }
       }
     },
+    "/panel/api/clients/{email}/flow": {
+      "post": {
+        "tags": [
+          "Clients"
+        ],
+        "summary": "Override a client's XTLS flow on a single inbound, bypassing the per-inbound capability clamp — so a client can keep xtls-rprx-vision on one flow-capable inbound and clear it on another in the same subscription. Clearing (\"\" or \"none\") is always allowed; a non-empty flow must be a recognized value and is only accepted on a flow-capable inbound.",
+        "operationId": "post_panel_api_clients_email_flow",
+        "parameters": [
+          {
+            "name": "email",
+            "in": "path",
+            "required": true,
+            "description": "Client email (unique identifier).",
+            "schema": {
+              "type": "string"
+            }
+          }
+        ],
+        "requestBody": {
+          "required": true,
+          "content": {
+            "application/json": {
+              "schema": {
+                "type": "object"
+              },
+              "example": {
+                "inboundId": 7,
+                "flow": ""
+              }
+            }
+          }
+        },
+        "responses": {
+          "200": {
+            "description": "Successful response",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "object",
+                  "properties": {
+                    "success": {
+                      "type": "boolean"
+                    },
+                    "msg": {
+                      "type": "string"
+                    },
+                    "obj": {}
+                  }
+                },
+                "example": {
+                  "success": true
+                }
+              }
+            }
+          }
+        }
+      }
+    },
     "/panel/api/clients/resetAllTraffics": {
       "post": {
         "tags": [
--- a/frontend/public/openapi.json
+++ b/frontend/public/openapi.json
@@ -1126,6 +1126,9 @@
             "description": "Flow control (XTLS)",
             "type": "string"
           },
+          "flowLock": {
+            "type": "boolean"
+          },
           "group": {
             "description": "Logical grouping label",
             "type": "string"
--- a/frontend/src/generated/examples.ts
+++ b/frontend/src/generated/examples.ts
@@ -224,6 +224,7 @@ export const EXAMPLES: Record<string, unknown> = {
     "enable": false,
     "expiryTime": 0,
     "flow": "",
+    "flowLock": false,
     "group": "",
     "id": "",
     "keepAlive": 0,
--- a/frontend/src/generated/schemas.ts
+++ b/frontend/src/generated/schemas.ts
@@ -1100,6 +1100,9 @@ export const SCHEMAS: Record<string, unknown> = {
         "description": "Flow control (XTLS)",
         "type": "string"
       },
+      "flowLock": {
+        "type": "boolean"
+      },
       "group": {
         "description": "Logical grouping label",
         "type": "string"
--- a/frontend/src/generated/types.ts
+++ b/frontend/src/generated/types.ts
@@ -232,6 +232,7 @@ export interface Client {
   enable: boolean;
   expiryTime: number;
   flow?: string;
+  flowLock?: boolean;
   group?: string;
   id?: string;
   keepAlive?: number;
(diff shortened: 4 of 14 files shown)


FZ1010:
Thanks — addressed in 87dfe06:

**Blocking 1 (route not in API docs).** Added the `POST /clients/:email/flow` entry to `frontend/src/pages/api-docs/endpoints.ts` and ran `make gen` — `openapi.json` now lists it (165 paths) and `make gen-check` is clean.

**Blocking 2 (clamp removed both ways).** A non-empty flow is now gated behind `inboundCanEnableTlsFlow` **and** the recognized-value allowlist (`bulkFlowAllowed`, the same set bulk-adjust uses); an unknown value or a flow on a non-capable inbound is rejected. Clearing (`""`/`"none"`) stays unconditional, since that's the safe intended use.

**Suggestion 3 (redundant mirror).** Removed — relying on `UpdateInboundClient` → `SyncInbound` recreating the `client_inbounds.flow_override` row from the written settings flow (client_link.go 152-156), as you noted. Dropped the now-unused `database`/`logger` imports.

**Style note.** Trimmed the long in-code NOTE. Moving its content here, since it's the one genuine semantics gap: an explicit *cleared* flow on a capable inbound can be re-populated by `restoreVisionFlowForEligibleInbound` if that inbound is later edited into a newly-eligible state (it restores the intended Vision from a sibling's `flow_override`). Making an explicit clear durable would need a marker — a sentinel `flow_override` value or a per-membership lock. Left out as a model decision; happy to implement whichever you'd prefer.

Still draft pending the frontend per-inbound control and live-panel verification.

FZ1010:
Pushed 2a95c1e — this closes the core gap that the override wasn't **durable**.

**Problem.** `SetInboundClientFlow`'s write was correct, but two pre-existing passes re-homogenize a client's flow across its inbounds and would silently undo a deliberate clear:
- `restoreVisionFlowForEligibleInbound` (runs on every boot **and** every inbound edit) re-injects Vision onto an empty-flow, flow-capable inbound whenever a sibling's `flow_override` resolves to Vision (`EffectiveFlowsByEmails`).
- `ClientService.Update` re-clamps the single edited flow onto every flow-capable inbound on any ordinary client edit (traffic/expiry/…).

So clearing Vision on a tunneled inbound while a Reality sibling kept it would revert on the next reboot or edit — exactly the #5689 scenario.

**Fix — a per-(client,inbound) `flowLock` marker:**
- `Client.FlowLock`, stored in the inbound's `settings` JSON. It survives `SyncInbound` (which rebuilds `flow_override` from settings) and never reaches Xray (`GetXrayConfig` rebuilds clean client entries, and Xray ignores unknown client fields regardless).
- `SetInboundClientFlow` pins the written client.
- `restoreVisionFlowForEligibleInbound` skips flow-locked clients.
- `ClientService.Update` preserves a flow-locked inbound's per-inbound flow (`clientForInboundEdit`) instead of clamping.
- Regenerated OpenAPI/schemas for the new field.

**Tests (DB-backed, sqlite/CGO):** a cleared override survives `MigrationRestoreVisionFlow` **and** a whole-client edit; restore skips a flow-locked client (with an un-pinned control proving the test catches a regression); `SetInboundClientFlow` validation paths (unknown flow / non-capable inbound / unknown client / valid pin).

Also removed the `//` line comments I'd added, per `CLAUDE.md`. Verified locally with `golangci-lint` + `go test ./internal/web/service ./internal/sub` + `make gen-check` (all green).

Still draft: this delivers the durable backend primitive; a panel UI + a per-inbound read endpoint are not in scope here (the feature stays API-only for now, as noted above).

FZ1010:
Pushed c5c9431 — **no longer API-only; added the panel UI.**

The client info modal already displayed a client's flow per inbound, gated on the inbound being flow-capable (`isVlessTlsFlow` = `canEnableTlsFlow`). Replaced that read-only display with a `Select` (`xtls-rprx-vision` / `none`) that calls `POST /clients/:email/flow` for that (client, inbound) and invalidates the inbounds query so the change reflects immediately.

Because it reuses the existing capability gate, it surfaces on every flow-capable inbound — **including a tunneled xhttp+vlessenc one (the #5689 scenario)** — so an operator can keep Vision on a Reality inbound and clear it on the tunneled one entirely from the panel. The `flowLock` from the previous commit keeps that durable against the restore pass and the standard client editor.

No new i18n keys (reuses the existing flow label + `none`; the handler returns the standard client-update toast). Verified: `typecheck`, `eslint`, full `vitest` (616), and the production `build` are green; `gen-check` clean.

FZ1010:
Closing to re-open as a fresh PR with a cleaned-up writeup and full `make verify` results. Same branch/commits; superseding PR linked shortly.

FZ1010:
Superseded by #5699 (same branch + commits, fresh writeup with the panel UI and full `make verify` results).

PR #5699, opened 2026-06-30
feat(clients): per-(client,inbound) XTLS flow override + panel UI (#5689)

### LINKED ISSUE #5689: [Feature]: Allow opting out of auto-applied XTLS Vision (xtls-rprx-vision) per inbound/client

FZ1010:
Closes #5689.

> Two alternative approaches to #5689 — this one (per-(client,inbound) override) and the inbound-scoped `DisableFlow` toggle in #5698. Merge whichever fits; the other can be closed.

## Problem

A VLESS client attached to multiple inbounds carries a single client-level `flow`. The panel clamps that flow onto **every** flow-capable inbound the client is on, so you cannot keep `xtls-rprx-vision` on a Reality inbound while turning it off on a tunneled XHTTP + VLESS-encryption inbound for the same client. Vision over a CDN/tunneled transport is frequently undesirable, but today it is all-or-nothing.

## Approach — writable per-(client,inbound) flow override

A new endpoint sets or clears the XTLS flow for **one client on one inbound**, independently of its flow on sibling inbounds, with a panel control to drive it.

### Backend
- `POST /panel/api/clients/:email/flow` → `SetInboundClientFlow(inboundId, email, flow)` writes the chosen flow into just that inbound's settings. It validates against the recognized-flow allowlist and the inbound's capability (clearing — `""`/`none` — is always allowed; an unknown value or a flow on a non-capable transport is rejected).
- **Durability via a `flowLock` marker.** Without it, `restoreVisionFlowForEligibleInbound` (runs on every boot and inbound edit) and the whole-client editor (`ClientService.Update`) both re-homogenize a client's flow across its inbounds and would undo a deliberate clear. `flowLock` is stored in the inbound's settings JSON (so it survives `SyncInbound`, which rebuilds `flow_override` from settings) and is never sent to Xray (`GetXrayConfig` rebuilds clean client entries, and Xray ignores unknown client fields). `SetInboundClientFlow` pins the written client; the restore pass and `Update` skip/preserve flow-locked entries.

### Frontend
- A new **Flow** tab on the inbound edit modal (**Modify Inbound** — alongside Basics/Protocol/Stream/Security/Sniffing/Advanced), so the override lives where an operator editing an inbound actually looks for inbound-scoped controls, the same place #5698's `DisableFlow` switch lives. It only renders when editing an existing inbound that is currently flow-capable (`canEnableTlsFlow`) and has at least one attached client. Each row lists a client's email with a `Select` (`xtls-rprx-vision` / `none`) that posts to the endpoint above and refreshes the inbound list in place — no page reload, modal stays open so multiple clients can be adjusted in one sitting.
- Deployed and exercised against a live panel (real VLESS Reality + XHTTP/TLS+vlessenc inbound pair, one client attached to both): the Flow tab appears only on the flow-capable inbound, the dropdown round-trips through the real endpoint, and the saved value persists correctly across modal reopen.

### Tests
- DB-backed: a cleared override survives both `MigrationRestoreVisionFlow` and a whole-client edit (a regression that fails without the `flowLock` + `Update` changes); `restoreVisionFlowForEligibleInbound` skips a flow-locked client (with an un-pinned control proving the test catches a regression); `SetInboundClientFlow` validation paths (unknown flow / non-capable inbound / unknown client / valid pin).

## `make verify` — ✅ `verify: OK`

Ran the full CI-mirroring gate locally (CGO sqlite via MinGW; golangci-lint built for go 1.26.4):

| stage | result |
|---|---|
| `gen-check` | clean — no codegen drift |
| `lint` (golangci-lint + eslint) | **0 issues** |
| `typecheck` (`tsc --noEmit`) | clean |
| `test` — `go test -shuffle=on ./...` | all packages pass, **0 failures** |
| `test` — frontend vitest | **634 passing** |
| `build` (Vite bundles + `go build ./...`) | OK |

No `//` line comments in committed Go/TS, per the repo guidelines.


FZ1010:
Marked ready — pushed 310ad41, the panel control moved.

The flow Select used to live in the client info popup (`InboundInfoModal`), which is a fairly buried screen. Moved it into a new **Flow** tab on the inbound edit modal (**Modify Inbound**, next to Basics/Protocol/Stream/Security/Sniffing/Advanced) — the same place #5698's `DisableFlow` switch lives, since that's where an operator editing an inbound actually looks for an inbound-scoped control. It only shows when editing a currently flow-capable inbound that has clients attached, and lists each client's email with the `xtls-rprx-vision`/`none` dropdown; changing it posts to the existing `/clients/:email/flow` endpoint and refreshes the list in place (modal stays open).

Deployed this to a real panel and exercised it end-to-end against a live VLESS Reality + XHTTP/TLS+vlessenc inbound pair sharing one client: the Flow tab appears only on the flow-capable inbound, the dropdown round-trips through the real endpoint, and the saved value persists correctly across modal reopen and a node restart.

PR description updated to match. `make verify` green (`gen-check`/`lint`/`typecheck`/`go test`/vitest 634/`build`).

PR #5711, opened 2026-07-01
fix: make all self-managed file downloads/installs atomic, with real completion status

nima1024m:
## Root cause

Every self-update path wrote downloaded/copied content directly onto a live path via `curl -o`/`cp -f` — no temp file, no exit-status check, no non-empty check. A truncated write corrupts the live file, causing a downstream failure later (bash syntax error, service parse failure, Xray routing failure) with no error at the time it actually happened.

**Fix pattern applied throughout:** download/copy → PID-suffixed temp file → verify curl/cp exit status → verify non-empty → `mv -f` → verify `mv` exit status → fail loudly on any step instead of continuing silently.

## Commits 1–3 — `/usr/bin/x-ui` itself

- `x-ui.sh`: `update_menu()`/`update_shell()` refactored onto a shared `replace_xui_script(url, use_if_modified_since)` helper.
- `update.sh`: `update_x-ui()` — same pattern.
- `install.sh`: identical overwrite bug, fixed the same way.
- Audited the Go dashboard updater (`internal/web/service/panel/panel.go`) — already used `os.CreateTemp` + `defer`/`EXIT` trap, safe by construction. Added the one thing missing: a zero-byte-download check.

## Commit 4 — completion-status signal (`feat`)

- `update.sh` writes `{runId,state,exitCode,finishedAt}` JSON to `/etc/x-ui/update-status.json` via `trap ... EXIT` (covers every exit path, including two bare `exit 1` sites that bypass `_fail()`).
- `panel.go` generates `runID := time.Now().UnixNano()`, passes it via `XUI_UPDATE_RUN_ID`/`XUI_UPDATE_STATUS_FILE` env vars.
- New `GET /panel/api/server/getUpdateStatus` endpoint; frontend (`PanelUpdateModal.tsx`) polls it instead of trusting bare HTTP reachability.
- In-memory concurrency guard (`acquireUpdateSlot`/`releaseUpdateSlot`) so two `/updatePanel` calls can't launch racing `update.sh` runs.
- `XUI_UPDATE_RUN_ID` validated `^[0-9]+$` before JSON interpolation.
- `runId` typed as a decimal **string** everywhere (Go struct, status file, generated TS type) — an `int64` UnixNano value exceeds `Number.MAX_SAFE_INTEGER` and would lose precision as a raw JSON number.

## Commit 5 — 12-angle whole-branch review fixes

- Concurrency guard wasn't releasing on `state:"failed"`, only on a 5-minute timeout — fixed to track `updateRunID` and check terminal status via `GetUpdateStatus()`.
- Added a genuine concurrent test (`TestAcquireUpdateSlotConcurrency`, 200 goroutines, `-race`) — prior tests were sequential-only, proven to give zero signal by temporarily breaking the mutex and confirming they still passed.
- Removed the redundant upfront `writeUpdateStatus` pending-write and the now-dead function.
- Added HTTP tests for `getUpdateStatus`, including a regression test that `runId` is a JSON string on the wire.

## Commit 6 — systemd unit atomicity

`_install_xui_service_unit()` helper in both `update.sh`/`install.sh`, covering `cp`-from-tarball (4 sites) and the `curl`-from-GitHub fallback (3-way distro `case`).

## Commit 7 — Alpine OpenRC atomicity + cleanup

- `/etc/init.d/x-ui` install in both scripts, same fix pattern.
- Removed a `//` inline comment (CLAUDE.md hard-rule violation) in `panel_test.go`.
- Corrected PR-description numbering/typecheck-claim errors from the prior round.

## Commit 8 — geo-data file atomicity

`update_geofiles()` in `x-ui.sh` — `geoip`/`geosite` `.dat` downloads. Non-trivial because the original code pointed `-z` (if-modified-since) and `-o` at the *same* live path. Separated so `-z` checks the live file while `-o` writes to a temp file; verified via a local HTTP server that a `304` response never touches the temp file, so the existing "already up to date" branch needed no change.

## Commit 9 — verify the release tarball extraction itself

The most significant remaining gap: `tar zxvf`'s exit status was never checked in either script, and nothing verified the extracted `x-ui` binary exists/is non-empty — with the old install *already deleted* by that point, no rollback. Added: non-empty-archive check, `tar` exit-status check, post-extraction binary check — `_fail`/`exit 1` loudly instead of continuing silently. Also:
- `replace_xui_script()` was returning `chmod`'s exit code instead of success (missing explicit `return 0`).
- `update_geofiles()` had no `case` default branch.
- Added a test for `updatePanel`'s untested `ParseBool` failure branch.

## Known gaps (documented, not fixed here)

- Tarball extraction is verified but not fully atomic (would need temp-dir + tree-swap; judged out of proportion for this PR).
- `startUpdate()`'s systemd-run/detached-fallback branching is untested (unsafe to test without a dependency-injection refactor).
- Pre-existing, unrelated `TelegramFilled`/`@ant-design/icons` build break — tracked as a separate task.
- Pre-existing 401-mid-poll axios-interceptor edge case (unbounded promise on session expiry).

## Testing

`bash -n` all three scripts; empirical scratch tests per fix (trap semantics, atomic-write success/failure/empty paths, tarball corruption scenarios); `go build`/`vet`/`test -race` clean; `npm run gen` confirmed in sync; `npm run typecheck`/`build` fail only on the pre-existing `TelegramFilled` issue.


nima1024m:
Thanks for the review — all three findings are addressed:

**Blocking: update-slot staleness could let a second `update.sh` through while the first was still genuinely running.**
Fixed by checking real process liveness instead of relying on elapsed time alone. On the detached-fallback launch path (the one that can track a PID), the update's PID is now recorded and `acquireUpdateSlot` calls a `processAlive(pid)` check (`syscall.Kill(pid, 0)` on Linux) once the stale window is reached — the slot stays held if the process is confirmed alive. The `systemd-run` path can't track a real PID, so for it `updateStaleAfter` was raised from 5 to 20 minutes as a more conservative fallback. A `updateHardCeiling` of 2 hours backstops both paths so a genuinely wedged process can't lock out retries forever. Added `TestAcquireUpdateSlotWaitsForAliveProcessPastStaleWindow` and `TestAcquireUpdateSlotHardCeilingOverridesLiveness` to cover both edges; both `go-test` and `race` pass on Linux CI.

**Suggestion: `updatePanel` entry in `endpoints.ts` missing a response example.**
Added, matching the `runId` string wire format `getUpdateStatus` already documents.

**Suggestion: `install.sh` temp-file leak on the two post-extraction failure branches.**
Added `rm -f "${xui_script_temp}"` to both.

All CI checks are green, including `go-test` and `race`.

PR #5721, opened 2026-07-01
docs(settings): clarify Sub Port/Sub Domain double as subscription-link fallback

volov-de:
## Summary

Clarify the description text for "Listen Port"/"Listen Domain"/"Reverse
Proxy URI" (subPort/subDomain/subURI) in the Subscription settings tab,
across all 13 locales, so admins running the panel behind a reverse
proxy don't get a broken subscription link/QR in the Clients page.

## Why

subPort/subDomain are documented purely as the subscription service's
own listen address ("must be an unused port", "leave blank to listen
on all domains and IPs"). But when "Reverse Proxy URI" is left empty,
these same values are silently reused to build the subscription link
shown in the panel's Client Info/QR view (`SettingService.BuildSubURIBase`,
called from `GetDefaultSettings` with `c.Request.Host` as the fallback
domain — see `internal/web/controller/setting.go` and
`internal/web/service/setting.go`).

If the admin panel and the public subscription endpoint are reached
through different domains/ports — a common reverse-proxy setup where
the sub service listens on an internal port (e.g. 2096) and is exposed
externally through a proxy on 443 under the main domain — the panel
generates a broken link: it uses the admin's own request host and the
raw internal port instead of the externally reachable one.

The actual `/sub` endpoint mostly self-corrects (it derives its base
from the real inbound request via `internal/sub/service.go`'s
`BuildURLs`), but the link the panel *shows and generates the QR for*
does not, since it's built from the settings API request's own Host
header.

The fix for this is already in the codebase — "Reverse Proxy URI"
exists for exactly this case (`s.buildSingleURL` prefers it over the
derived base) — but nothing in the UI text connects the dots for
someone who only reads Sub Port/Sub Domain's descriptions and doesn't
know to look at Reverse Proxy URI. Hit this on our own deployment:
panel and subscription served from different subdomains behind Caddy,
clicking a client's subscription QR in the panel produced
`http://panel.example.com:2096/sub/<id>` (wrong scheme, wrong host,
wrong port) instead of the working `https://sub.example.com/sub/<id>`.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot

## How was this tested?

Text-only change to the 13 translation JSON files (`internal/web/translation/*.json`).
Validated every file still parses as JSON and matches `prettier` formatting
(`npx prettier --check internal/web/translation/*.json`). No behavior/logic
change, so no functional test needed. English and Russian wording reviewed
directly; the other 11 locales were translated consistently with the
existing terminology already used for these same settings keys in each
file — happy to have a native speaker double-check phrasing on review.

## Breaking changes

None.

## Checklist

- [ ] I added or updated tests for the new behavior (when applicable). — text-only, N/A.
- [ ] `go build ./...` and the test suite pass locally. — no Go code changed.
- [ ] For frontend changes: `npm run lint`, `npm run typecheck`, and `npm run build` pass. — no frontend code changed, only translation JSON (verified with `prettier --check`).
- [x] I updated the Wiki / README / API docs if user-facing behavior changed. — this PR is the doc/UI-text update itself.


volov-de:
Thanks for the review — pushed a follow-up commit addressing all three points: terminating punctuation before the appended sentence (en-US, id-ID, pt-BR, tr-TR, uk-UA, ar-EG), restored full-width CJK parentheses in zh-CN/zh-TW/ja-JP, and fixed the es-ES field-name mismatch in subURIDesc (Dominio de Escucha / Puerto de Suscripción).

PR #5768, opened 2026-07-03
fix(inbounds): apply runtime changes after the DB commit

n0ctal:
## Summary

- Move inbound runtime mutations out of the database transaction and run them only after a successful commit.
- Keep node-hosted Add/Update converged through the existing dirty-node reconcile path.
- Keep node-hosted Delete safe for `selected` sync mode by issuing a live post-commit delete, with dirty-node reconcile left as a backup.
- Update the node selected-tag allow-list inside the same database transaction as the inbound row change.

## Why

The old code could apply Xray/runtime changes before the inbound database write was durably committed. If the commit failed, runtime state and database state could diverge.

During review, the first draft also had a selected-mode delete regression: marking the node dirty alone is not enough when the deleted tag is no longer in the node's selected allow-list. Reconcile intentionally does not sweep unselected tags, so the worker could keep serving a deleted inbound. This revision fixes that path directly.

## Scope

- `AddInbound`: save DB state in a checked transaction, update selected-tag allow-list in that same transaction, then apply local runtime changes after commit.
- `UpdateInbound`: save DB state in a checked serialized transaction, update selected-tag allow-list in that same transaction, then apply local runtime changes after commit.
- `DelInbound`: delete DB state in a checked transaction, mark the node dirty in that transaction, then apply node/local delete after commit.
- PostgreSQL commit-failure tests cover Add and Update.
- A selected-mode regression test covers node-hosted delete.

## Validation

- `go test ./internal/web/service -run 'TestDelInbound_NodeSelectedModeDeletesRemoteImmediately|TestReconcileNode_|TestEnsureInboundTagAllowed|TestMarkNodeDirtyTxRollsBackWithTransaction|TestAddInbound_PostgresCommitFailure|TestUpdateInbound_PostgresCommitFailure' -count=1 -v`
- `go test -race ./internal/web/service -run 'TestDelInbound_NodeSelectedModeDeletesRemoteImmediately|TestReconcileNode_SelectedModeLeavesUnselectedRemoteInbounds|TestEnsureInboundTagAllowed' -count=1 -v`
- `go vet ./internal/web/service`
- `git diff --check`

The PostgreSQL commit-failure tests require `XUI_DB_TYPE=postgres` and `XUI_DB_DSN`; without those variables they skip locally and run in the dedicated CI job.

## Risk

Add/Update for node-hosted inbounds now converge asynchronously through dirty-node reconcile instead of pushing the remote row inline. In normal operation this means a short propagation delay of roughly one sync interval.

Delete does not rely solely on that async path: it sends a live post-commit delete to the node so `selected` sync mode cannot leave a ghost inbound running. If the live delete fails because the node is offline, the dirty flag remains as a retry signal for the normal reconcile path.

For selected-mode Add/Update, the selected-tag allow-list update is kept in the same database transaction as the inbound row. That preserves rollback semantics without reintroducing runtime side effects before commit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,6 +37,39 @@ jobs:
           go list ./... | grep -v '/frontend/node_modules/' > /tmp/go-packages.txt
           go test -shuffle=on -count=1 $(cat /tmp/go-packages.txt)
 
+  postgres-durable-first:
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:16
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: xui_durable
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd "pg_isready -U postgres -d xui_durable"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+    steps:
+      - uses: actions/checkout@v7
+      - uses: actions/setup-go@v6
+        with:
+          go-version-file: go.mod
+          cache: true
+      - name: Stub internal/web/dist for go:embed
+        run: mkdir -p internal/web/dist && touch internal/web/dist/.gitkeep
+      - name: PostgreSQL durable-first tests
+        run: |
+          set -o pipefail
+          XUI_DB_TYPE=postgres XUI_DB_DSN="host=127.0.0.1 port=5432 user=postgres password=postgres dbname=xui_durable sslmode=disable" \
+            go test ./internal/web/service -run 'PostgresCommitFailure' -count=1 -v | tee /tmp/postgres-durable-first.log
+          if grep -q -- '--- SKIP' /tmp/postgres-durable-first.log; then
+            exit 1
+          fi
+
   codegen:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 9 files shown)


n0ctal:
CI note: the failing `golangci` check is a pre-existing `main` issue, not introduced by this branch.

The workflow pins `golangci-lint: latest`, which currently resolves to `v2.12.2`, and runs with `only-new-issues: false`, so it lints the whole tree rather than just the diff. The only findings are SA5011 (staticcheck) — nil dereference after a nil check — in two test files this PR does not modify:

- `internal/web/service/setting_mtls_test.go:32,35,65,68`
- `internal/xray/api_wireguard_test.go:26,30`

Neither file is part of this PR's diff, which touches `internal/web/service/inbound.go`, `node.go`, `internal/logger/logger.go`, the CI workflow (adds only the `postgres-durable-first` job), and its own tests. Every other check is green: `go-test`, `race`, `postgres-durable-first`, all `build (*)` targets, CodeQL, `govulncheck`, `frontend`, `fuzz-smoke`, `codegen`.

I've kept the fix for those two unrelated tests out of this PR to keep it focused. Happy to open a separate small PR for them if useful; alternatively, pinning the `golangci-lint` version would stop `latest` from surfacing new findings on unrelated PRs.

PR #5774, opened 2026-07-03
fix(routing): allow dns.servers on private IPs past the geoip:private block rule

### LINKED ISSUE #5773: Default routing template silently blocks Xray's own DNS when dns.servers points to a private IP
## Summary

When a custom DNS server is configured on a **private/internal IP** (e.g. a self-hosted AdGuard Home / Pi-hole running in the same Docker network as Xray), the default xray template's routing rule `{"ip":["geoip:private"],"outboundTag":"blocked"}` silently blocks Xray-core's own DNS lookups to that server. This causes a consistent ~4 second stall on **every new domain-based connection** through the inbound (Xray falls back to a raw dial-by-hostname after its internal DNS attempt times out, so the connection eventually succeeds, but slowly — making this very hard to diagnose from the client side).

## Environment

- 3x-ui: 3.4.2 (Docker deployment)
- Xray-core: 26.6.27
- Reproduced on a VLESS+Reality inbound, but the root cause is generic to any inbound relying on the panel's default routing template.

## Reproduction

1. Run AdGuard Home (or any DNS resolver) as another container on the same Docker network as 3x-ui, e.g. reachable at `172.20.0.53:53`.
2. In the inbound's Xray template (`dns.servers`), point Xray's own DNS resolution at that internal IP: `"dns": {"servers": ["172.20.0.53"], "queryStrategy": "UseIPv4"}`.
3. Leave the routing rules at the panel's default (`panel/api/xray/getDefaultJsonConfig`), which include:
   ```json
   {"ip": ["geoip:private"], "outboundTag": "blocked", "type": "field"}
   ```
4. Connect through the inbound to any domain-based destination (not a raw IP).

**Result:** every fresh connection to a new domain takes ~4s before completing (still succeeds, just slow). Connections to raw IPs (no domain resolution needed) are fast (tens of ms).

## Root cause

Enabling Xray's debug log level (`log.loglevel: "debug"`) shows exactly what happens:

```
INFO - XRAY: app/dns: UDP:172.20.0.53:53 querying DNS for: example.com.
INFO - XRAY: transport/internet/udp: establishing new connection for udp:172.20.0.53:53
INFO - XRAY: app/dispatcher: taking detour [blocked] for [udp:172.20.0.53:53]
...
ERROR - XRAY: app/dns: failed to lookup ip for domain example.com at server UDP:172.20.0.53:53 in serial query mode > app/dns: record not found
ERROR - XRAY: proxy/freedom: failed to get IP address for domain example.com > app/dns: returning nil for domain example.com > app/dns: record not found
INFO - XRAY: transport/internet/tcp: dialing TCP to tcp:example.com:443
```

Xray's own DNS query (UDP to `172.20.0.53:53`) is dispatched through the routing table like any other traffic. Since `172.20.0.53` is a private IP, it matches the default `geoip:private -> blocked` rule and gets dropped. Xray waits out its internal DNS timeout (~4s), gives up, and falls back to dialing the destination by raw hostname (which succeeds via a different path) — so the bug is easy to miss since the connection *works*, just slowly, with no error surfaced to the client.

This is arguably correct/intentional behavior on Xray-core's side (routing rules apply uniformly to all traffic, including the engine's own DNS traffic, by design — many users rely on this). The gap is specifically in **3x-ui's default template**: it ships a private-IP block rule without reconciling it against whatever `dns.servers` the user configures, even though the panel has both pieces of information at config-generation time. Pointing Xray's DNS at a same-host/same-network internal resolver (AdGuard Home, Pi-hole, unbound, etc.) is a very common self-hosted pattern, so this combination is likely to bite other users too.

## Suggested fix

When generating/saving the Xray config, either:
- Automatically insert an allow-rule for each configured `dns.servers` entry (`{"type":"field","ip":["<dns-server-ip>/32"],"outboundTag":"direct"}`) immediately before the `geoip:private -> blocked` rule, or
- At minimum, surface a warning in the UI when a DNS server address falls inside a private range while `geoip:private` blocking is enabled in routing.

Manual workaround (confirmed working): add the following rule right after the `api` rule and before `geoip:private`:
```json
{"type": "field", "ip": ["<your-dns-server-ip>/32"], "outboundTag": "direct"}
```

Happy to open a PR for either the auto-rule generation or the UI warning if that's welcome — let me know which direction you'd prefer.

volov-de:
## Summary

Auto-generates a `direct` allow-rule for any private/internal IP found in `dns.servers`, inserted immediately before the panel's default `geoip:private -> blocked` routing rule, so Xray's own DNS lookups no longer get silently dropped by that rule.

## Why

Closes #5773.

Xray's own DNS client traffic is dispatched through the same routing table as proxied client traffic. When `dns.servers` points at a private IP (e.g. a self-hosted AdGuard Home / Pi-hole reachable on the same Docker network as Xray — a common self-hosted setup) and the panel's default `geoip:private` block rule is active, Xray's own DNS lookups get silently dropped by that rule. Xray then falls back to dialing destinations by raw hostname once its internal DNS attempt times out (~4s), so proxied connections still work — just with a multi-second stall added to every new domain-based connection, with no error surfaced anywhere. Root-caused with an Xray debug-log trace, details in the linked issue.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [x] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot

## How was this tested?

- New `EnsureDnsServerRouting` (`internal/web/service/xray_setting_dns_routing.go`, mirrors the existing `EnsureStatsRouting` hoist) is wired into `SaveXraySetting` right after `EnsureStatsRouting`.
- `internal/web/service/xray_setting_dns_routing_test.go` — 12 new cases: no-op paths (no `dns.servers` / public IP / no block rule present), correct insertion position, both server-entry shapes (bare string and `{address, port, ...}` object), scheme/port/path stripping (including bracketed IPv6), skipping special/domain addresses (`localhost`, `fakedns`, plain hostnames), idempotency across repeated saves, in-place update when the server set changes, rule removal once no longer needed, and leaving hand-authored rules with extra matchers untouched.
- `go test ./internal/web/service/...` — all packages pass, no regressions.
- `go vet ./internal/web/service/...` and `gofmt -l` on the changed files — clean.
- `go build ./internal/web/service/...` — clean (built via a `golang:1.26` container; couldn't run `go build ./...` for the full binary locally since `internal/web/web.go`'s `go:embed all:dist` needs a built frontend, which is unrelated to this change).
- Manually verified against the real-world repro from the issue (VLESS+Reality inbound + AdGuard Home on the same Docker network): before the fix, every new-domain connection stalled ~4s; after applying the equivalent rule by hand (and now via this change), connections complete in tens of ms.

## Screenshots / recordings

N/A — backend-only change, no UI.

## Breaking changes

None. The generated rule is only added when both a private `dns.servers` IP and an existing `geoip:private` block rule are present, and is kept in sync automatically; nothing changes for setups that don't have this combination.

## Checklist

- [x] `go build ./...` and the test suite pass locally. (see caveat above: built/tested the affected package directly rather than the full binary, since the frontend isn't part of this change)
- [ ] For frontend changes: `npm run lint`, `npm run typecheck`, and `npm run build` pass. — N/A, no frontend changes.
- [x] I updated the Wiki / README / API docs if user-facing behavior changed. — N/A, this is an internal, automatic fix; there was no prior doc describing the old (buggy) behavior to update.

### DIFF
--- a/internal/web/service/xray_setting.go
+++ b/internal/web/service/xray_setting.go
@@ -29,6 +29,9 @@ func (s *XraySettingService) SaveXraySetting(newXraySettings string) error {
 	if hoisted, err := EnsureStatsRouting(newXraySettings); err == nil {
 		newXraySettings = hoisted
 	}
+	if synced, err := EnsureDnsServerRouting(newXraySettings); err == nil {
+		newXraySettings = synced
+	}
 	return s.saveSetting("xrayTemplateConfig", newXraySettings)
 }
 
--- a/internal/web/service/xray_setting_dns_routing.go
+++ b/internal/web/service/xray_setting_dns_routing.go
@@ -3,18 +3,28 @@ package service
 import (
 	"encoding/json"
 	"net"
+	"reflect"
 	"sort"
+	"strconv"
 	"strings"
 
 	"github.com/mhsanaei/3x-ui/v3/internal/util/netsafe"
 )
 
 // dnsAllowRuleShape identifies routing rules this file manages: a plain
-// "type=field, ip=[...], outboundTag=direct" rule with no other matchers.
-// Rules shaped like this are treated as owned by EnsureDnsServerRouting and
-// are kept in sync with the current dns.servers config on every save;
-// anything else (including rules an admin wrote by hand that happen to also
-// allow-list an IP) is left untouched.
+// "type=field, ip=[...], port=..., outboundTag=direct" rule with no other
+// matchers. An "enabled" key is tolerated as long as it's true — the
+// Routing tab's rule editor (RuleFormModal.tsx submit()) and its enabled
+// switch (RoutingTab.tsx toggleRule()) always write that key back, even
+// when nothing else changed, so requiring its absence would disown the
+// rule the first time an admin so much as opens it in the UI. A rule
+// toggled off (enabled=false) is treated as no longer ours: the admin
+// explicitly turned it off, and re-enabling it on the next save would
+// silently override that choice.
+//
+// Rules shaped like this are kept in sync with the current dns.servers
+// config on every save; anything else (including rules an admin wrote by
+// hand that happen to also allow-list an IP) is left untouched.
 func dnsAllowRuleShape(rule map[string]any) bool {
 	if t, _ := rule["type"].(string); t != "field" {
 		return false
@@ -25,9 +35,17 @@ func dnsAllowRuleShape(rule map[string]any) bool {
 	if _, ok := rule["ip"]; !ok {
 		return false
 	}
+	if _, ok := rule["port"]; !ok {
+		return false
+	}
 	for key := range rule {
 		switch key {
-		case "type", "outboundTag", "ip":
+		case "type", "outboundTag", "ip", "port":
+			continue
+		case "enabled":
+			if enabled, ok := rule[key].(bool); !ok || !enabled {
+				return false
+			}
 			continue
 		default:
 			return false
@@ -45,8 +63,7 @@ func findPrivateBlockRule(rules []map[string]any) int {
 		if out, _ := rule["outboundTag"].(string); out != "blocked" {
 			continue
 		}
-		ips := readRuleIPs(rule["ip"])
-		for _, ip := range ips {
+		for _, ip := range readRuleIPs(rule["ip"]) {
 			if strings.EqualFold(ip, "geoip:private") {
 				return i
 			}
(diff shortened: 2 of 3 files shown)


volov-de:
Thanks for the thorough review — pushed a follow-up commit addressing all three points:

1. **Port scoping (blocking).** The allow-rule now carries an explicit `port` matcher, grouped by whichever port(s) the private `dns.servers` entries actually use, instead of opening every port on that IP. Two internal resolvers on different ports produce two separate `ip`+`port` rules rather than one rule that would cross-allow ip1:port2/ip2:port1.

2. **Position drift (blocking).** `EnsureDnsServerRouting` now strips every previously-managed rule on each save and rebuilds the current set fresh, reinserted immediately before the re-indexed `geoip:private` block rule — so a rule dragged below the block rule in the Routing tab (directly, or incidentally while reordering something else) gets moved back rather than silently reintroducing the stall with nothing to notice it.

3. **`enabled` key tolerance.** `dnsAllowRuleShape` now tolerates `enabled: true` (matching the existing `EnsureStatsRouting`/`delete(apiRule, "enabled")` precedent), so opening the rule in the editor or toggling it on doesn't disown it. An explicitly disabled rule is left alone and a fresh one is (re-)created instead of silently re-enabling it — respects the admin's choice.

Also had to change how no-op detection works: comparing the rebuilt rules against the original via `reflect.DeepEqual` on the parsed Go values false-positived on identical content stored as `[]any` (from `json.Unmarshal`) vs `[]string` (freshly built) — now both sides are decoded through `encoding/json` to a common representation before comparing.

5 new tests cover multi-port grouping, the reorder case, and both `enabled` states; existing tests updated for the `port` field. Full package + `go vet`/`gofmt` still clean.

volov-de:
Quick follow-up for maintainers: the current HEAD already addresses all three `github-actions` review points.

- The managed allow-rule is now port-scoped, rebuilt from the actual private `dns.servers` endpoints, and split per distinct port so it does not widen the exception beyond the DNS traffic that needs it.
- Managed DNS allow-rules are fully stripped and rebuilt on each save, then reinserted immediately before the `geoip:private` block rule, so UI reordering cannot silently reintroduce the stall.
- Managed-rule detection now tolerates the UI-added `enabled: true` field, avoiding disowning / duplication after opening or toggling the rule in the Routing tab.

The previously blocking threads point to older hunks; current CI / Release / CodeQL checks are all green. A re-review would be appreciated.

volov-de:
All previously raised review threads are now resolved, and current CI / Release / CodeQL checks are green. The follow-up commit addresses the earlier port-scoping, rule reordering, and `enabled`-handling concerns. Would appreciate a quick re-review when you have a moment.

PR #5807, opened 2026-07-05
fix(frontend): clarify client credential fields

### LINKED ISSUE #5803: [Feature]: Add a tooltip/hint to the "Password" field in the client form clarifying which protocols use it

The client edit/create form (`ClientFormModal.tsx`) is shared across all multi-client protocols (VLESS, VMess, Trojan, Shadowsocks, Hysteria, WireGuard), and shows every protocol-specific field at once — including a plain **"Password"** field with no label qualifier, no tooltip, and no indication of which protocol(s) it actually applies to.

Given how many different protocols the panel supports side by side, it's genuinely not obvious from the UI alone what this field affects. We only figured it out by reading the Go source (`internal/xray/api.go`, `buildUserAccount`):

```go
case "vmess":        Id                                           // Password not read
case "vless":        Id, Flow                                     // Password not read
case "trojan":       Password (required)                          // ← used here
case "shadowsocks":  Password (required) + cipher                 // ← used here
case "hysteria":     Auth (required)                               // separate field, not Password
case "wireguard":    PublicKey, PreSharedKey, AllowedIPs, KeepAlive // Password not read
```

So "Password" is only meaningful for Trojan and Shadowsocks clients. For VLESS/VMess it's silently ignored by the engine, and for Hysteria the relevant field is the separate "Hysteria Auth" input right below it — which, without reading the source, looks like it could be redundant with "Password" rather than being the actually-required field.

This caused real confusion on our end: on an older node the Password field was populated (apparently auto-generated by the browser form at some point), while on newer nodes provisioned via our own automation it was left empty — and it wasn't obvious from the panel alone whether this mattered at all for our protocol (VLESS), or was a misconfiguration.

### Describe the feature

Add a short tooltip (or inline hint text) to the Password `Form.Item` in `ClientFormModal.tsx` stating that it only applies to Trojan / Shadowsocks clients and is ignored for other protocols. Ideally something similar could be added near "Hysteria Auth" clarifying it's the credential Hysteria actually uses.

### Additional context

Not a functional bug — the field works correctly for the protocols that use it. This is purely a clarity/UX gap that could save other users from the same "why is this here / does it matter" confusion.

sanmaxdev:
## Summary

- Add inline hints to the client Password and Hysteria Auth fields.
- Add regression coverage for the credential help text.

## Why

Closes #5803

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot

## How was this tested?

- [x] `npm ci`
- [x] `npm run test -- src/test/client-form-modal.test.tsx`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm run test`
- [x] `npm run build`
- [x] `make gen-check`
- [x] `go build ./...`

## Screenshots / recordings

N/A

## Breaking changes

None

## Checklist


sanmaxdev:
Translated the credential hint text for all 12 non-English locale files. The English source in en-US.json is unchanged.

- `go build ./...`
- `go vet ./internal/web/...`
- `git diff --check`

MHSanaei:
https://github.com/MHSanaei/3x-ui/pull/5809

PR #5886, opened 2026-07-09
feat: Add outbound egress metadata (IP + country)

isultanov99:
Show egress IP and country information for outbound HTTP tests. The probe reuses the temporary SOCKS route from the existing HTTP test and fetches Cloudflare trace metadata after the reachability check succeeds.

The outbound list now adds separate Egress and Country columns, hides egress IPs until the user reveals them, and marks Cloudflare WARP results with an orange cloud pill. Mobile cards keep the same data compact by placing the country and IPv4/IPv6 values on separate lines.

Validation: npm run typecheck; npm run lint; npm run build; go test ./internal/web/service/outbound


isultanov99:
Без имени

isultanov99:
Good points, thanks. Fixed in the latest push:

- Restored the `Real Delay` selector and the original `TCP` default, so this PR no longer removes an existing test mode.
- Kept the existing HTTP probe unit tests hermetic by injecting the egress trace probe in tests instead of allowing real Cloudflare/DNS calls.
- Added a shorter dedicated timeout for egress trace metadata and fetch IPv4/IPv6 trace results concurrently.
- Scoped the mobile egress IP reveal state per outbound row instead of sharing one toggle across all cards.
- Added keyboard activation for the new eye toggles.
- Treat non-`off` WARP trace values, including `plus`, as WARP-like for the country pill.

Validation run locally:

- `go test ./internal/web/service/outbound`
- `npm run typecheck`
- `npm run lint`
- `npm run build`

PR #5888, opened 2026-07-10
fix(clients): include Telegram ID in client list search

### LINKED ISSUE #5880: [Bug]: Client search by Telegram user ID does not work in Web UI in v3.4.2
### Describe the bug

In previous versions of 3x-ui, when I entered a client's Telegram user ID in the Web UI search box, the panel correctly showed all clients related to that Telegram user ID.

However, in 3x-ui `v3.4.2`, searching by Telegram user ID no longer works in the Web UI.

For example, if a client has a Telegram user ID such as `759865`, searching for `759865` in the client list does not return the related client anymore.

Searching by other fields such as email still works, but searching by Telegram user ID seems to be ignored or not included in the client search/filter logic.

This looks like a regression, because the same behavior worked correctly in previous versions.

1. Install or upgrade to 3x-ui `v3.4.2`.
2. Create or use an existing client that has a Telegram user ID set.
3. Open the 3x-ui Web UI.
4. Go to the clients list.
5. Enter the Telegram user ID in the search box.

Example search value:

```text
759865

### Expected behavior

_No response_

### Actual behavior

### Actual behavior

In 3x-ui `v3.4.2`, searching by Telegram user ID in the Web UI does not return the related clients.

The client exists and has a Telegram user ID, but when I enter that Telegram user ID in the search box, the client list does not show the matching account.

It seems that the Web UI search/filter logic no longer checks the Telegram user ID field.

This issue only affects searching/filtering. The client itself still exists and can be found by other fields such as email.

### Relevant logs

```shell
### Relevant logs

No backend error is shown.

The issue happens in the Web UI client list search.

Example:

Client has Telegram user ID:


759865
```

### Screenshots

_No response_

### 3x-ui version

3.4.2

### Xray-core version

26.6.27

install.sh script

### Operating system

ubuntu 26

_No response_

chrome Version 149

None

### Before submitting

mrnickson-hue:
## Summary

Fixes #5880 — searching the client list by Telegram user ID no longer matched, because `clientMatchesSearch` (`internal/web/service/client_paging.go`) only checked `Email, SubID, Comment, UUID, Password, Auth`.

This is a genuine regression, not a field that simply was never included. Before the paged/search backend endpoint was added in #4500, the frontend searched with `ObjectUtil.deepSearch()` over the *entire* client object, which recursed into every field — including `tgId` — since it's just a number that gets `toString()`'d and substring-matched. Replacing that with a fixed backend candidate list silently dropped Telegram ID from being searchable. Since `tgId` is normally populated through the panel's own Telegram bot linking flow rather than typed in manually, being able to look a client up by it in the panel is the actual point of the field.

## Change

- `clientMatchesSearch` now also matches on `TgID`. It's `int64` (0 = unset), so it can't sit in the existing `[]string` candidates array — matched separately via `strconv.FormatInt`, and skipped when `0` so a needle of `"0"` doesn't spuriously match every client without a Telegram ID.
- Added `client_search_tgid_test.go`: full match, substring match, no false-positive on an unrelated numeric needle, the `"0"` guard, and confirmation the existing fields still match (no regression).

## Test plan

- [x] `go test ./internal/web/service/...` — all packages pass
- [x] `go vet ./internal/web/service/...` — clean
- [x] New test `TestClientMatchesSearch_TgID` covers the fix directly


mrnickson-hue:
Thanks for the review — both points addressed:

1. Dropped the `//` comment block above the `TgID` check (repo convention: code should read on its own).
2. Updated `searchPlaceholder` in all 13 locale files to mention Telegram ID, since the search box now actually matches on it.

Both in a188a473.

MHSanaei:
We don't need test for this
Remove the test

mrnickson-hue:
Removed in fd92e949.

PR #5941, opened 2026-07-13
fix(email): build an RFC 5322 message with a proper From address

yukh975:
### Problem

Sending a notification or test email from the panel can be rejected by strict receivers. With Gmail the server returns:

> 550-5.7.1 ... Messages missing a valid address in From: header, or having no From: header, are not accepted.

### Root cause

`buildMessage` set the `From:` header to the raw SMTP **username** and emitted only `From`/`To`/`Subject`/`MIME-Version`/`Content-Type`:

1. When the SMTP login is not a bare email address (common with relays / submission services), `From:` carries no valid address and the message is rejected.
2. There is **no `Date:`** (mandatory per RFC 5322 §3.6) and **no `Message-ID:`**, both of which also raise the spam score (e.g. SpamAssassin `MISSING_DATE` / `MISSING_MID`).

### Fix

- Add two settings: **`smtpFrom`** (sender address) and **`smtpFromName`** (optional display name). `From` falls back to the username when `smtpFrom` is empty, so existing setups are unchanged.
- Reassemble the message with `net/mail`: a proper name-addr `From` (`"Name" <addr>`), a `Date`, a `Message-ID`, and an RFC 2047 encoded `Subject`, in a deterministic header order.

Resulting message:

```
Date: Mon, 13 Jul 2026 15:47:21 +0300
From: "3x-ui" <panel@example.com>
To: admin@example.com
Message-ID: <c99cb6a1...@example.com>
Subject: =?utf-8?q?...?=
MIME-Version: 1.0
Content-Type: text/html; charset=utf-8
```

### Changes

- **Backend:** `smtpFrom` / `smtpFromName` settings (model, defaults, getters), SMTP send + test paths, `buildMessage` rewrite.
- **Frontend:** two fields in the Email settings tab; regenerated Zod/OpenAPI artifacts (`make gen`).
- **i18n:** new keys in all 13 locale files.
- **Tests:** `buildMessage` unit tests (message parsed back with `net/mail`).

### Compatibility

Backward compatible — with `smtpFrom` empty the `From` address is still the username, as before (now correctly formatted as a `name-addr`).


yukh975:
Thanks for the review. I've addressed the header-injection finding in the latest push:

- **Validation at the source** — `AllSetting.CheckValid()` now rejects `smtpFrom` unless `mail.ParseAddress` accepts it. This blocks the CRLF vector (the parser refuses embedded newlines) and surfaces a malformed address at configuration time rather than at send time.
- **Defense in depth** — `buildMessage` strips CR/LF from the address and name unconditionally, so it is safe even if called in isolation or reached through a future code path.
- **Tests** — `TestCheckValidSmtpFrom` (valid addresses pass; a `\r\n`-laden or malformed value is rejected) and `TestBuildMessageStripsHeaderInjection` (a crafted `Bcc:` / `X-Evil:` payload does not leak into the headers).

`gofmt`, `go build`, `go vet`, and the `email` + `entity` tests pass locally.

PR #5952, opened 2026-07-14
fix(script): remove release download time limit

### LINKED ISSUE #5933: [Bug]: Remove hard-coded download timeout from the installer script
## Description

The installer currently fails on slow or unstable network connections because it uses a fixed download timeout for large files.

In environments with limited bandwidth, high latency, or proxy chains (for example, Privoxy → SOCKS5 → SSH tunnel), large downloads may legitimately take longer than the configured timeout. As a result, the installation aborts even though the download is still making progress.

Typical error:

```text
curl: (28) Operation timed out after 300 seconds
curl: (28) Send failure: Broken pipe
```

## Suggested Improvement

Please remove the hard-coded download timeout from the installer script.

Instead of:

```bash
curl --max-time 300 ...
```

consider either:

```bash
curl ...
```

or make the timeout configurable through an environment variable, for example:

```bash
CURL_MAX_TIME=${CURL_MAX_TIME:-0}
```

or

```bash
if [ "${CURL_MAX_TIME:-0}" -gt 0 ]; then
    curl --max-time "$CURL_MAX_TIME" ...
else
    curl ...
fi
```

## Why

* Supports slow Internet connections.
* Works better behind HTTP/SOCKS proxies and SSH tunnels.
* Prevents installation failures while data is still being transferred.
* Gives users the option to define their own timeout if needed.
* Improves reliability without affecting users on fast networks.

## Additional Suggestion

It may also be helpful to expose other networking parameters (such as retry count and connection timeout) as optional environment variables, allowing users in restricted network environments to customize the installer without modifying the script.

in install script

### Expected behavior

in install script

### Actual behavior

in install script

### Relevant logs

```shell
in install script
```

### Screenshots

_No response_

### 3x-ui version

all version

### Xray-core version

all version

install.sh script

### Operating system

ubuntu 24.0

_No response_

chrome

None

### Before submitting

sanmaxdev:
## Summary

Remove the 300-second transfer cap from both release archive downloads while retaining retries and the connection timeout.

## Why

Slow but active downloads should not fail after five minutes.

Closes #5933

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

## How was this tested?

- `bash -n install.sh`
- `shellcheck -S error install.sh`
- `golangci-lint run --timeout 10m`
- `go test -shuffle=on -count=1 $(go list ./... | grep -v '/frontend/node_modules/')`
- `go build ./...`
- `npm run typecheck`
- `npm run lint`
- `npm test`
- `npm run build`
- `npm run build-storybook`
- `git diff --check`

## Breaking changes

None.

## Checklist

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1424,7 +1424,7 @@ install_x-ui() {
             exit 1
         fi
         echo -e "Got x-ui latest version: ${tag_version}, beginning the installation..."
-        curl -fLR --retry 5 --retry-delay 3 --connect-timeout 15 --max-time 300 -o ${xui_folder}-linux-$(arch).tar.gz https://github.com/MHSanaei/3x-ui/releases/download/${tag_version}/x-ui-linux-$(arch).tar.gz
+        curl -fLR --retry 5 --retry-delay 3 --connect-timeout 15 -o ${xui_folder}-linux-$(arch).tar.gz https://github.com/MHSanaei/3x-ui/releases/download/${tag_version}/x-ui-linux-$(arch).tar.gz
         if [[ $? -ne 0 ]]; then
             echo -e "${red}Downloading x-ui failed, please be sure that your server can access GitHub ${plain}"
             exit 1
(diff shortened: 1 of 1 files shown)


sanmaxdev:
Added a low-speed watchdog to both archive downloads, so active transfers have no fixed deadline while stalled transfers still fail after five minutes.

Checks: `bash -n install.sh`, `shellcheck -S error install.sh`, transfer-option assertions, and `git diff --check`.

PR #5988, opened 2026-07-16
fix(clients): allow multiple clients to share a subId

oxsean:
## Problem

Editing a client fails with `Duplicate subId` when its subscription ID matches another client's. In a multi-node setup, clients reconciled from different nodes are separate records with distinct emails and cannot be merged into one client, so a shared subId is the only way to aggregate them under a single subscription link.

## Why the uniqueness check is wrong

Commit 88a36773 introduced per-client subId uniqueness on the grounds that the first-class client model aggregates via one client attached to many inbounds. That holds for top-down provisioning, but not for clients synced bottom-up from nodes. Meanwhile the subscription pipeline explicitly supports many records per subId: `getInboundsBySubId` joins across all clients with the subId, and `indexStatsBySubId` plucks every matching email (plural). The checks blocked a flow the sub server already serves.

## Changes

- Remove the subId collision checks in `ClientService.Create`, `Update`, `BulkCreate` and portable import (`internal/web/service/client_crud.go`, `client_bulk.go`, `client_portable.go`).
- Email uniqueness is untouched, including the re-add identity checks that compare an existing email's subId.
- Replace `TestClientUpdateDuplicateSubIDDoesNotRenameEmail` with `TestClientUpdateAllowsSharedSubID`, asserting an update to a shared subId succeeds and leaves the other client intact.

## Testing

- `go test -shuffle=on -count=1 ./internal/web/service/ ./internal/sub/` — all pass (547 + sub package).
- `gofmt` / `go vet` clean on changed files.

### DIFF
--- a/internal/web/service/client_bulk.go
+++ b/internal/web/service/client_bulk.go
@@ -1148,9 +1148,7 @@ func (s *ClientService) BulkCreate(inboundSvc *InboundService, payloads []Client
 	}
 	prep := make([]prepared, 0, len(payloads))
 	emails := make([]string, 0, len(payloads))
-	subIDs := make([]string, 0, len(payloads))
 	seenEmail := make(map[string]struct{}, len(payloads))
-	seenSubID := make(map[string]string, len(payloads))
 
 	for i := range payloads {
 		client := payloads[i].Client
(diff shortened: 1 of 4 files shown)


oxsean:
Filed #6065 describing the underlying problem and reproduction steps in detail.

PR #6003, opened 2026-07-16
fix: refresh stale client_traffics row when an inbound-deleted client's email is reused

### LINKED ISSUE #5958: [Bug]: stale client_traffics row survives when a client's email is reused after its inbound is deleted
### Describe the bug

Deleting an inbound does not delete the `clients` / `client_traffics` rows of the clients that were attached to it — they become orphaned (no `client_inbounds` link) but stay in the database with their last known values.

If a client is later re-added under the *same email* (e.g. after recreating the inbound, or attaching a "new" client that happens to reuse an old identifier) via `POST /panel/api/clients/add`, the `clients` table row is correctly refreshed with the new values — but the `client_traffics` row is not. `AddClientStat` (internal/web/service/inbound_traffic.go) inserts with `OnConflict{Columns: [email], DoNothing: true}`, so if a `client_traffics` row for that email already exists (the orphaned one), the insert is silently skipped and the stale row is kept as-is: old `enable`, old `expiry_time`, old `reset`, old `total`, old `up`/`down`, and even the `inbound_id` of the now-deleted inbound.

This is not just leftover garbage — `client_traffics` is the *only* place traffic usage (`up`/`down`) is tracked (the `clients` table has no traffic columns), and it's also the table `autoRenewClients` actually queries against (`reset > 0 AND expiry_time > 0 AND expiry_time <= now`), not `clients`. So a client that looks correctly configured in the UI (`clients.enable=true`, `clients.reset=1`, fresh `expiryTime`) can silently:
- report traffic usage left over from a *different, long-deleted* client that used to own that email, and
- never actually get auto-renewed, because the enforcement query is filtering on the stale `client_traffics.reset`/`expiry_time`, not the fresh ones visible in the UI.

I ran into this while investigating #5954 (a different, still-unexplained "auto-renew doesn't re-enable the client" report) — my reproduction attempts for that issue didn't reproduce its exact symptom, but surfaced this instead.

1. Create an inbound with an embedded client (any protocol), e.g. `email=test`, `enable=false`, `expiryTime` in the past, `reset=0`.
2. Delete that inbound via `POST /panel/api/inbounds/del/:id`.
3. Confirm the client is now orphaned: `GET /panel/api/clients/get/test` still returns the client (`inboundIds: []`), and it still shows up in `GET /panel/api/clients/list/paged`.
4. Add a *new* client with the same email to a different (or newly recreated) inbound via `POST /panel/api/clients/add`, this time with different values, e.g. `enable=true`, `expiryTime` a future date, `reset=1`, `totalGB=1073741824`.
5. `GET /panel/api/clients/list/paged?search=test` again.

### Expected behavior

The `client_traffics` row for that email should reflect the values of the client that currently exists — either by being refreshed on insert, or by being deleted along with the client/inbound so a genuinely fresh row gets created.

### Actual behavior

The list response shows the divergence directly — top-level fields (from `clients`) are fresh and correct, but the nested `.traffic` object (from `client_traffics`) is untouched from before the inbound was deleted, still referencing the deleted inbound's id:

```json
{
  "email": "test-orphan-repro",
  "enable": true,
  "expiryTime": 1784212100951,
  "reset": 1,
  "inboundIds": [1],
  "traffic": {
    "id": 37,
    "inboundId": 5,
    "enable": false,
    "expiryTime": 1784039027944,
    "total": 0,
    "reset": 0,
    "up": 0,
    "down": 0
  }
}
```

(`inboundId: 5` was the inbound I had already deleted two steps earlier.)

I did *not* find a UI-visible symptom from this specific divergence — the enable toggle and the expiry/duration column both read the top-level fields (`clients` table), which are correct in this repro, so the client looks fine at a glance. The impact is in traffic accounting and auto-renew enforcement silently using the stale row.

### Relevant logs

N/A — no errors are logged; the insert is silently skipped by design (`DoNothing` on conflict).

### Screenshots

N/A (see reproduction JSON above)

### 3x-ui version

Tested against current `main` (post-3.5.0, includes the #5619/#789e92c auto-renew fix)

### Xray-core version

26.6.27

N/A for this report — tested directly against the API on an existing install

### Operating system

Ubuntu 26.04, Postgres-backed panel

Client traffic tracking / auto-renew (`client_traffics` table), inbound deletion

### Suggested direction

A couple of options, not sure which fits the codebase's intent best:
- Change `AddClientStat`'s `OnConflict` from `DoNothing` to an actual update of the conflicting row (mirroring what `ClientService.SyncInbound` already does for the `clients` table on an existing row).
- Or, when an inbound is deleted, clean up `client_traffics` (and `NodeClientTraffic`/`ClientGlobalTraffic`) rows for any client that ends up with zero remaining `client_inbounds` links as a result — i.e. treat "fully detached" the same as "deleted" for stats purposes.

Happy to help test a fix.

mrnickson-hue:
## Problem

Closes #5958.

Deleting an inbound (`InboundService.DelInbound`) only removes the `client_inbounds` link for its clients — it does **not** delete their `clients` / `client_traffics` rows. That's intentional: it mirrors `ClientService.Detach`, which does the exact same thing on purpose so a client can be re-`Attach`ed later with its accumulated traffic intact.

The problem is what happens if that same email is instead reused for a **freshly (re)created** client on a different inbound, via `ClientService.Create`. `AddClientStat`'s `OnConflict` clause was `DoNothing` on the `email` column:

```go
return tx.Clauses(clause.OnConflict{Columns: []clause.Column{{Name: "email"}}, DoNothing: true}).
    Create(&clientTraffic).Error
```

So the orphaned `client_traffics` row — with the old, possibly disabled/expired/depleted `enable` / `expiry_time` / `reset` / `total` / `inbound_id` — survives completely untouched. The new client's config is silently ignored. In practice this means:

- The client's real usage (`up`/`down`, only stored in `client_traffics`) shows traffic from the previous, unrelated client that happened to share the email.
- `autoRenewClients` filters on `client_traffics.reset` / `expiry_time`, not the fresh values in `clients` — so auto-renew can silently never trigger for the "new" client, because it's still reading the old row.

Full repro / investigation notes are in the issue.

## Fix

Change the conflict path from `DoNothing` to a scoped `DoUpdates`:

```go
return tx.Clauses(clause.OnConflict{
    Columns:   []clause.Column{{Name: "email"}},
    DoUpdates: clause.AssignmentColumns([]string{"inbound_id", "total", "expiry_time", "enable", "reset"}),
}).Create(&clientTraffic).Error
```

This is safe for the *legitimate* case `AddClientStat`'s conflict handling also has to serve — a client attached to more than one inbound shares a single `client_traffics` row. Every inbound attachment for the same identity calls `AddClientStat` with the same `enable`/`expiry_time`/`reset`/`total` (they all come from the same `clients` row), so refreshing those columns on each call is a no-op in that case — confirmed by `TestAddClientStat_MultiInboundReattachStaysIdempotent` below.

`up`/`down` are **deliberately excluded** from the refresh — they're the accumulated traffic counters, and zeroing them here would erase real usage every time an existing, actively-used client is attached to one more inbound.

**Known limitation** (not fully solved by this change, flagged in the code comment): a *genuinely new* client that happens to reuse an orphaned email still inherits that row's leftover `up`/`down`, since nothing at this call site can distinguish "this is the same logical identity re-attaching" from "this is an unrelated new client that picked the same email." Fixing that would need to detect the reuse case explicitly (e.g. at `Create` time) and is out of scope for this narrow, low-risk fix.

## Testing

Added `internal/web/service/client_stat_reuse_test.go`:

- `TestAddClientStat_RefreshesStaleRowOnInboundDeleteThenReuse` — full repro: create a client, disable it, delete its only inbound (orphaning it), then re-create the same email/subId on a new inbound with different settings. Asserts the `client_traffics` row picks up the fresh `enable`/`reset`/`total`/`expiry_time`/`inbound_id`, while `up`/`down` stay untouched. **Verified this test fails without the fix** (reproduces the exact stale values reported in #5958) and passes with it.
- `TestAddClientStat_MultiInboundReattachStaysIdempotent` — guards the legitimate multi-inbound-share case: seeds real `up`/`down` usage, then re-`Create`s the same client onto a second inbound, and asserts the usage counters and config columns are unchanged.

Full suite, to check for regressions elsewhere:

```
go test ./internal/web/service/... -v
```

All packages pass (`service`, `email`, `integration`, `outbound`, `panel`, `tgbot`), no failures. `go vet ./internal/web/service/...` is clean.

### DIFF
--- a/internal/web/service/client_stat_reuse_test.go
+++ b/internal/web/service/client_stat_reuse_test.go
@@ -1,0 +1,177 @@
+package service
+
+import (
+	"testing"
+
+	"github.com/mhsanaei/3x-ui/v3/internal/database"
+	"github.com/mhsanaei/3x-ui/v3/internal/database/model"
+	"github.com/mhsanaei/3x-ui/v3/internal/xray"
+)
+
+// TestAddClientStat_RefreshesStaleRowOnInboundDeleteThenReuse covers #5958:
+// deleting a client's only inbound leaves its clients/client_traffics rows in
+// place (matching ClientService.Detach, which does the same on purpose so a
+// later Attach can resume a client with its accumulated traffic intact). If
+// that same email is instead reused for a freshly (re)created client via
+// ClientService.Create, the new enable/expiry/reset/total must win over
+// whatever the orphaned row still holds instead of being silently ignored by
+// AddClientStat's OnConflict.
+func TestAddClientStat_RefreshesStaleRowOnInboundDeleteThenReuse(t *testing.T) {
+	setupBulkDB(t)
+	svc := &ClientService{}
+	inboundSvc := &InboundService{}
+
+	const email = "reused@example.com"
+	const subID = "sub-reused"
+
+	ibA := mkInbound(t, 22001, model.VLESS, `{"clients":[]}`)
+
+	// Create starts every client enabled (it forces Enable=true), so the
+	// depleted/disabled shape a client has by the time it's naturally deleted
+	// is reached the same way production reaches it: a follow-up Update, not
+	// the initial Create.
+	if _, err := svc.Create(inboundSvc, &ClientCreatePayload{
+		Client: model.Client{
+			Email: email, SubID: subID, Enable: true,
+			TotalGB: 0, ExpiryTime: 1000, Reset: 0,
+		},
+		InboundIds: []int{ibA.Id},
+	}); err != nil {
+		t.Fatalf("initial Create: %v", err)
+	}
+	rec0 := lookupClientRecord(t, email)
+	if _, err := svc.Update(inboundSvc, rec0.Id, model.Client{
+		Email: email, SubID: subID, Enable: false,
+		TotalGB: 0, ExpiryTime: 1000, Reset: 0,
+	}); err != nil {
+		t.Fatalf("Update to disabled: %v", err)
+	}
+
+	db := database.GetDB()
+	var before xray.ClientTraffic
+	if err := db.Model(xray.ClientTraffic{}).Where("email = ?", email).First(&before).Error; err != nil {
+		t.Fatalf("lookup client_traffics before delete: %v", err)
+	}
+	if before.Enable || before.Reset != 0 || before.Total != 0 {
+		t.Fatalf("unexpected initial client_traffics row: %+v", before)
+	}
+
+	// Delete the client's only inbound. This must NOT delete the client or its
+	// traffic row (that's the documented, intentional Detach-parity behavior) —
+	// it only orphans them.
+	if _, err := inboundSvc.DelInbound(ibA.Id); err != nil {
+		t.Fatalf("DelInbound: %v", err)
+	}
+
+	rec := lookupClientRecord(t, email)
+	ids, err := svc.GetInboundIdsForRecord(rec.Id)
+	if err != nil {
+		t.Fatalf("GetInboundIdsForRecord: %v", err)
+	}
+	if len(ids) != 0 {
+		t.Fatalf("client should be fully detached after its only inbound was deleted, still attached to: %v", ids)
+	}
+	var stillThere xray.ClientTraffic
+	if err := db.Model(xray.ClientTraffic{}).Where("email = ?", email).First(&stillThere).Error; err != nil {
+		t.Fatalf("client_traffics row should survive inbound deletion (Detach parity), lookup failed: %v", err)
+	}
+
+	// Reuse the same email + subId (the only way ClientService.Create allows
+	// re-adding under an already-used email) on a freshly created inbound, with
+	// deliberately different, "fresh" settings.
+	ibB := mkInbound(t, 22002, model.VLESS, `{"clients":[]}`)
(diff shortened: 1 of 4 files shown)


mrnickson-hue:
Thanks for the review — the High severity finding was correct and reproducible: importing an inbound with a client disabled at export time silently un-forced the controller's `Enable = true` on import, because the `AddClientStat` call for the Settings-derived client ran *after* the `ClientStats` loop and (with the new `DoUpdates`) won the conflict.

Fixed in 49dd0fd: the `ClientStats` loop's emails are now tracked and skipped by the subsequent `AddClientStat` loop, so the import path's forced values stay authoritative and the plain (non-import) create path is unaffected (its `ClientStats` is always empty, so nothing is skipped there). Added `TestAddInbound_ImportForcedEnableSurvivesDisabledSettingsClient`, which reproduces the exact scenario — confirmed it fails on the pre-fix code and passes now.

Also updated the stale `addClientTraffic` comment (Low finding) that still described `AddClientStat` as `DoNothing`.

Full `internal/web/service/...` suite still green, `go vet` clean.

PR #6020, opened 2026-07-17
feat: add SSH access mode for node management (Phase 1-3)

yangbin1322:
Extend node management so the panel can adopt a plain Linux server over SSH, alongside the existing API mode that manages an installed 3x-ui panel. SSH and API are two access modes of one node, not separate systems: the server list and management flow stay unified, and an SSH node is designed to become an API node in place once a panel is installed on it (a later phase).

Why on the Node model rather than a new model: the requirement keeps one unified server list and foresees converting an SSH row into an API-managed node in place, which a split model would need a migration between. model.Node gains a Mode (api|ssh, defaulting to api so every existing row is untouched) plus SSH fields; GORM AutoMigrate adds the columns, no hand-written migration needed.

Credentials must be replayed to the remote host, so unlike a login password they cannot be hashed. SSH password and private key are encrypted at rest with AES-256-GCM under XUI_SECRET_KEY (internal/util/crypto/secret.go), fail-closed if the key is unset, and are write-only over the API (json:"-"). Host-key verification mirrors the existing TlsVerifyMode pattern via SshHostKeyMode (trust|pin|skip) + SshHostKeySha256 rather than blindly accepting any key, so a credential is never handed to a machine-in-the-middle.

Backward compatibility: normalize() branches on mode and keeps API-node validation identical; the heartbeat probes SSH nodes over SSH (reachable/ unreachable) instead of HTTP so they never park at a false "offline" and never feed the panel-only traffic-sync or CPU/memory history.

Adds POST /panel/api/nodes/testSSH, the SSH connect/test service, the React add/edit dialog with an access-mode toggle, i18n keys across all 13 locales, and unit + real-SSH-server integration tests. Verified end to end against a live Ubuntu 22.04 host (password and private-key auth, host-key capture, OS detection, encrypted-at-rest confirmed in the DB).

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot

## Checklist

- [ ] I tested the change locally and confirmed the described behavior.
- [ ] I added or updated tests for the new behavior (when applicable).
- [ ] My commits follow the project's existing message style.
- [ ] I have no unrelated changes mixed into this PR.


yangbin1322:
Closing this upstream PR — moving it to my own fork instead. The SSH access mode work will be merged within yangbin1322/3x-ui-Manager.

Issue #4760, opened 2026-06-01
[Bug]: Hysteria2 Inbound wrong import to outbound and wrong connect


gomanuke:
### Describe the bug

I do appreciate for Your hard work. 3 2 5 became really good.
I just create hysteria 2 inbound on one x-ui and try to add that via copy client's config to another x-ui

### How to reproduce the problem

Just add on one x-ui Hysteria 2 inbound with UDP mask  (that's one of missed config option while importing)
Then add client and connect to that inbound
then make a copy URL from client information panel.
O another x-ui add Ounbound switch JSON tab paste and import.

Check Alpn (always h3 ignores another), check  UDP Mask (lost).  Manually correcting it doesn't help to make connection.
In fact i didn't find a way to solve that trivial task ((( I mean connect from x-ui 3 2 5 to x-ui 3 2 5 by hysteria 2

### Relevant logs

```shell

```

### 3x-ui version

3.2.5

### Xray-core version

26.6.1

### How did you install 3x-ui?

install.sh script

### Operating system

Debian 13

### Is the panel behind a reverse proxy or CDN?

No — direct access

### Before submitting

- [x] I searched [existing issues](../issues?q=is%3Aissue) and this bug has not been reported.
- [x] I am running the latest released version of 3x-ui (or have verified the bug still exists on it).
- [x] This bug report is written in English.
- [x] I have redacted any sensitive data (tokens, real domains, client UUIDs).

gomanuke:
**client share URL** 

hysteria2://78e7795a209c4c099f896a816fc8448f@news.domain.org:8443?alpn=h2%2Chttp%2F1.1&fm=%7B%22udp%22%3A%5B%7B%22settings%22%3A%7B%22password%22%3A%22ftwfgb9655hh2mgo%22%7D%2C%22type%22%3A%22salamander%22%7D%5D%7D&fp=chrome&obfs=salamander&obfs-password=655hh2mgo&security=tls&sni=news.domain.org#hy2-ej596ty350qs

**raw JSON that appears in the outbound**

{
  "protocol": "hysteria",
  "settings": {
    "address": "news.domain.org",
    "port": 8443,
    "version": 2
  },
  "tag": "hy2-ej596ty350qs",
  "streamSettings": {
    "network": "hysteria",
    "security": "tls",
    "hysteriaSettings": {
      "version": 2,
      "auth": "78e7795a209c4c099f896a816fc8448f",
      "udpIdleTimeout": 60
    },
[...]

PR #6048, opened 2026-07-20
fix(sub): omit non-standard fm param from Hysteria2 URI

### LINKED ISSUE #5982: [Bug]: Hysteria2 subscription link includes non-standard "fm" (finalmask) param that breaks parsing in mihomo/sing-box/subconverter
### Describe the bug

The Hysteria2 line in the plain/universal subscription (base64 list of URIs) includes a non-standard `fm` query parameter that raw-dumps the internal `finalmask.udp[]` JSON config, in addition to the already-present standard `obfs`/`obfs-password` fields:

hysteria2://<uuid>@host:443?alpn=h3&fm=%7B%22udp%22%3A%5B%7B%22settings%22%3A%7B%22password%22%3A%22...%22%7D%2C%22type%22%3A%22salamander%22%7D%5D%7D&fp=firefox&obfs=salamander&obfs-password=...&security=tls&sni=host

`fm` is not part of the Hysteria2 URI scheme (https://v2.hysteria.network/docs/developers/URI-Scheme/) and is not understood by mihomo, sing-box, the official hysteria2 client, or subconverter. Since `obfs`/`obfs-password` already fully describe the salamander obfuscation, `fm` is redundant.

This did not happen before upgrading from 3x-ui < 3.5.0 (Xray-core 26.6.27) to 3.5.0. It appears to be tied to the addition of FinalMask support (commit c59f54b, "outbound: finalmask"). After the upgrade, the generated subscription link stopped working with mihomo (fails the whole proxy-provider update).

### Expected behavior

Do not serialize the raw `finalmask` object into the public subscription URI — `obfs`/`obfs-password` (and `security=tls`, `sni`, `alpn`) are already sufficient for standard clients.

### Actual behavior

_No response_

### Relevant logs

```shell

```

### Screenshots

_No response_

### 3x-ui version

3.5.0

### Xray-core version

26.7.11

install.sh script

### Operating system

Ubuntu 24.04

_No response_

Yes — Nginx

### Before submitting

H-TTTTT:
## Summary
- Stop serializing the internal `finalmask` object as the non-standard `fm` query param in Hysteria2 subscription URIs (`genHysteriaLink`).
- Keep standard `obfs` / `obfs-password` for salamander obfuscation (same approach as the Clash generator).

## Why
`fm` is a v2rayN-specific extension, not part of the Hysteria2 URI scheme. With salamander enabled, mihomo (and other strict clients) reject the link during proxy-provider updates.

## Test plan
- [x] `go test ./internal/sub/ -run TestGenHysteriaLinkOmitsFinalMaskQueryParam`
- [ ] Manual: Hysteria2 inbound with salamander → raw/base64 sub link has `obfs`/`obfs-password`, no `fm=`
- [ ] Mihomo proxy-provider refresh succeeds

Fixes #5982

### DIFF
--- a/internal/sub/service.go
+++ b/internal/sub/service.go
@@ -1025,11 +1025,10 @@ func (s *SubService) genHysteriaLink(inbound *model.Inbound, email string) strin
 		}
 	}
 
-	// salamander obfs (Hysteria2). The panel-side link generator already
-	// emits these; keep the subscription output in sync so a client has
-	// the obfs password to match the server.
+	// salamander obfs (Hysteria2). Emit only the standard URI fields;
+	// the non-standard fm=<json> finalmask dump breaks mihomo and other
+	// Hysteria2 clients that reject unknown query params.
 	if finalmask, ok := stream["finalmask"].(map[string]any); ok {
-		applyFinalMaskParams(finalmask, params)
 		if udpMasks, ok := finalmask["udp"].([]any); ok {
 			for _, m := range udpMasks {
 				mask, _ := m.(map[string]any)
--- a/internal/sub/service_test.go
+++ b/internal/sub/service_test.go
@@ -1110,3 +1110,32 @@ func TestHysteriaHopPorts(t *testing.T) {
 		})
 	}
 }
+
+func TestGenHysteriaLinkOmitsFinalMaskQueryParam(t *testing.T) {
+	stream := `{
+		"security":"tls",
+		"tlsSettings":{"serverName":"hy.sni","alpn":["h3"],"settings":{"fingerprint":"chrome"}},
+		"finalmask":{"udp":[{"type":"salamander","settings":{"password":"obfs-secret"}}]}
+	}`
+	in := &model.Inbound{
+		Listen:         "203.0.113.1",
+		Port:           443,
+		Protocol:       model.Hysteria,
+		Remark:         "hy2",
+		Settings:       `{"version":2,"clients":[{"auth":"hyauth","email":"user"}]}`,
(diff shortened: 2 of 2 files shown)


pterodaktil02:
After refreshing the outbound subscription, the generated Hysteria2 outbound keeps the address, port, SNI and ALPN, but completely loses streamSettings.finalmask.

3x-ui v3.6.0 generates valid standard Hysteria2 subscription links, but its own outbound subscription importer cannot restore Salamander FinalMask from the standard obfs and obfs-password parameters after fm was removed.

The generated outbound contains:

"streamSettings": {
  "hysteriaSettings": {
    "udpIdleTimeout": 60,
    "version": 2
  },
  "network": "hysteria",
  "security": "tls",
  "tlsSettings": {
    "alpn": ["h3"],
    "fingerprint": "chrome",
    "serverName": "REDACTED"
  }
}

The expected Salamander configuration is missing:

"finalmask": {
  "udp": [
    {
      "type": "salamander",
      "settings": {
        "password": "REDACTED"
      }
    }
  ]
}

Packet capture confirms that the consumer sends UDP packets to the correct server and port, but the server does not reply because the packets are not Salamander-obfuscated. Other clients using the same node work correctly.

Downgrading only the source panels from v3.6.0 to v3.5.0 and refreshing the outbound subscription immediately restores the connection.

It appears that #6048 correctly removed the non-standard fm parameter and retained the standard obfs / obfs-password fields, but the 3x-ui outbound subscription importer does not convert those standard fields back into Xray finalmask.udp.

Related: #4760.

PR #6110, opened 2026-07-26
fix: stop deleting client_traffics for detached-but-alive clients

mrnickson-hue:
## Summary

Fixes two related bugs behind the reported "Ended/Disabled clients hidden" symptom: a migration that deletes `client_traffics` rows for clients live in only one of the two places that can now hold "this client still exists" (the `clients` table or an inbound's JSON), and a frontend live-recompute that silently drops any client missing a traffic row out of every summary bucket.

## Why

Closes #6102.

The reporter's screenshot shows `Clients 67` but `Ended 0 + Depleting 0 + Disabled 0 + Active 58 = 58` — the buckets don't sum to the total, and the hover popovers that normally list matching emails are empty, leaving the Filter drawer as the only way to reach those clients.

Root cause has two independent parts:

1. **`MigrationRemoveOrphanedTraffics` (`internal/web/service/inbound_migration.go`)** originally keyed "orphaned" off a client's absence from every inbound's `settings.clients[]` JSON — a definition that predates #4469's standalone `clients` table. `ClientService.Detach` intentionally keeps a client's `client_traffics` row when it drops its last inbound attachment (so the client can be re-attached later without losing stats/expiry), but such a client has no entry in any inbound's JSON anymore. Every `x-ui migrate` run or backup restore (`MigrateDB`) therefore deleted its traffic row anyway, even though the client itself was untouched and still listed — reproducing the exact "row exists, `—` under Attached inbounds, invisible to the summary" state from the screenshot. My first pass at this PR simply switched the predicate to key off the `clients` table instead — an automated review on this PR (see the thread below) caught that this trades one gap for a worse one: the one-shot `ClientsTable` seeder (`internal/database/db.go`) skips a client it fails to unmarshal and never retries, so a client that's still live in an inbound's JSON can have no `clients` row at all, and an empty `clients` table (edge case, but reachable) would delete every `client_traffics` row outright. Fixed by unioning both keep-sets — a row survives if it's referenced by either the `clients` table or any inbound's JSON, and is only removed when it's in neither.
2. **`frontend/src/hooks/useClients.ts`** recomputed the clients summary from the `client_stats` WebSocket snapshot as soon as it arrived, without checking whether that snapshot actually covered every client. `allClientStats` is `client_traffics` rows (email-keyed), so any client missing a row for the reason above (or any other reason) silently fell out of every bucket in the recompute while the headline `total` — kept from the server — still counted it. The added guard falls back to the server summary whenever the snapshot is shorter than the server's total; note this is a count comparison, not an exact coverage check (a genuinely orphaned `client_traffics` row — the case #1 prunes, only on migrate/restore — could in theory offset a missing one and let a short snapshot slip past the guard), and it's a deliberately visible trade: when it does trip, the summary cards stop updating live and only refresh on the next list poll, rather than risk showing an undercount.

I went through the whole master/node sync path (`setRemoteTrafficLocked`) separately, since the reporter also described a master+node deployment where clients "completely disappear" after depleting/being disabled there. I could not reproduce that through that path — `disableInvalidClients` only flips `enable=false` both in the DB row and inside the settings JSON, the client stays in the array, and every deletion branch in the node sync is already guarded against exactly this kind of false-positive sweep. Left a comment on the issue asking the reporter to confirm whether `x-ui migrate` or a backup restore happened around when they started seeing it, since that would point at the same bug from a different, more surprising trigger.

## Type of change

- [x] Bug fix

## Areas affected

- [x] Database / migrations
- [x] Multi-node (sub-nodes) — investigated, not changed (see above)

## How was this tested?

- `TestMigrationRemoveOrphanedTraffics` (`internal/web/service/inbound_migration_test.go`) covers all four combinations via real code paths, not hand-built rows where a real path exists: a client attached normally (`ClientService.SyncInbound`), a client detached from every inbound but kept alive (driven through the actual `ClientService.Detach(keepTraffic=true)` call), a client present in an inbound's JSON with no `clients` row (the seeder-gap case, built by hand since no production path currently produces it on purpose), and a truly orphaned row in neither. Confirmed each case fails against the wrong predicate (the original bug, and separately the regression from unioning-instead-of-replacing) and passes with the fix.
- Three new cases in `frontend/src/test/clients-summary.test.ts` for the extracted `pickClientsSummary`: falls back to the server summary when the snapshot is shorter than the server's total, uses the live recompute when the snapshot covers every client, and falls back before the first WS snapshot arrives.
- `go build ./...`, `go vet ./...`, `gofmt -l` / `goimports -l` on the changed files — clean. (`golangci-lint` itself couldn't run in my sandbox — the resolvable release is built against Go 1.25 and refuses to lint a go.mod targeting 1.26.5 — but gofmt/goimports cover the formatting side of it.)
- `go test ./...` and `go test -race ./internal/web/service/...` — all green.
- Frontend: `npm run typecheck`, `npm run lint`, full `npm run test` (890 tests), `npm run build`, `npm run build-storybook` — all green.

## Screenshots / recordings

N/A — this is a data-correctness fix, not a visual change. The summary cards' appearance is unchanged; they just stop undercounting (and, in the rare short-snapshot case described above, stop live-updating until the next poll instead).

## Breaking changes

None. `MigrationRemoveOrphanedTraffics` still runs automatically on `x-ui migrate` / backup restore; it just no longer deletes rows it shouldn't.

## Checklist

- [x] I added or updated tests for the new behavior.
- [ ] I updated the Wiki / README / API docs if user-facing behavior changed. — N/A, no user-facing behavior/docs to update.

### DIFF
--- a/frontend/src/hooks/useClients.ts
+++ b/frontend/src/hooks/useClients.ts
@@ -117,6 +117,24 @@ export function computeClientsSummary(
   return { total: stats.length, active, online, depleted, expiring, deactive };
 }
 
+// A client_stats snapshot shorter than the server's own total means it does
+// not cover every client — computeClientsSummary would then drop the missing
+// ones out of every bucket while total still counts them (#6102). The server
+// summary is always internally consistent (its buckets sum to total), so
+// fall back to it rather than publish an undercount.
+export function pickClientsSummary(
+  serverSummary: ClientsSummary,
+  allClientStats: ClientStatRow[],
+  onlineSet: Set<string>,
+  expireDiffMs: number,
+  trafficDiffBytes: number,
+): ClientsSummary {
+  if (allClientStats.length === 0) return serverSummary;
+  if (serverSummary.total > allClientStats.length) return serverSummary;
+  const live = computeClientsSummary(allClientStats, onlineSet, expireDiffMs, trafficDiffBytes);
+  return { ...live, total: serverSummary.total || live.total };
+}
+
 function buildQS(p: ClientQueryParams): string {
   const sp = new URLSearchParams();
   sp.set('page', String(p.page || 1));
@@ -266,17 +284,14 @@ export function useClients() {
   const pageSize = (defaults.pageSize as number) ?? 0;
 
   // Live summary: the client_stats WS event refreshes allClientStats every few
-  // seconds, so the top counters track reality without a page refresh. Falls
-  // back to the server-computed summary until the first event lands, and keeps
-  // the server's authoritative total for the headline count.
+  // seconds, so the top counters track reality without a page refresh — see
+  // pickClientsSummary for why it isn't always trusted over the server value.
   const [allClientStats, setAllClientStats] = useState<ClientStatRow[]>([]);
   const [clientSpeed, setClientSpeed] = useState<Record<string, ClientSpeedEntry>>({});
-  const summary = useMemo<ClientsSummary>(() => {
-    const serverSummary = listQuery.data?.summary ?? DEFAULT_SUMMARY;
-    if (allClientStats.length === 0) return serverSummary;
-    const live = computeClientsSummary(allClientStats, new Set(onlines), expireDiff, trafficDiff);
-    return { ...live, total: serverSummary.total || live.total };
-  }, [allClientStats, onlines, expireDiff, trafficDiff, listQuery.data?.summary]);
+  const summary = useMemo<ClientsSummary>(
+    () => pickClientsSummary(listQuery.data?.summary ?? DEFAULT_SUMMARY, allClientStats, new Set(onlines), expireDiff, trafficDiff),
+    [allClientStats, onlines, expireDiff, trafficDiff, listQuery.data?.summary],
+  );
 
   const invalidateAll = useCallback(
     () => {
--- a/frontend/src/hooks/useClients.ts
+++ b/frontend/src/hooks/useClients.ts
@@ -117,11 +117,6 @@ export function computeClientsSummary(
   return { total: stats.length, active, online, depleted, expiring, deactive };
 }
 
-// A client_stats snapshot shorter than the server's own total means it does
-// not cover every client — computeClientsSummary would then drop the missing
-// ones out of every bucket while total still counts them (#6102). The server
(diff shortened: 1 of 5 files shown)


mrnickson-hue:
Thanks — the union-vs-replace catch is correct and I appreciate the specific `db.go` line references, that's exactly the gap. Went through every point:

**High/Medium — union instead of replace.** Fixed in `07d13fc2`: the predicate now keeps a row if it's referenced by either the `clients` table or any inbound's `settings.clients[]` JSON, matching your suggested SQL. Verified the regression you found is real by temporarily reverting to my original (replace) version and confirming the new test case fails against it, then passes against the union fix.

**Medium — test coverage gap.** Rewrote the test as `TestMigrationRemoveOrphanedTraffics`, a table of all four combinations: attached (both sources), detached-but-alive (clients table only, driven through the real `ClientService.Detach(keepTraffic=true)` call rather than a hand-built row, per your suggestion), the seeder-skipped case you found (JSON only, no `clients` row — built by hand since no production path currently produces it on purpose), and truly orphaned (neither). All four pass against the fix; the JSON-only case reproduces your finding against the pre-union code.

**Medium — the frontend guard is a count, not a coverage check.** Agreed, and I don't think it's closeable without the server shipping buckets alongside `client_stats` as you note, which is out of scope here. Documented the limitation and the resulting trade-off (a short snapshot makes the summary stop updating live rather than risk an undercount) in the PR description instead of an inline comment, for the reason below.

**Low — silently discarding the delete's result.** Fixed: now logs a warning on error and, on success, logs `RowsAffected` at info level so a large delete is visible in the panel log instead of silent.

**Low — behavior-change not mentioned in the PR body.** Added a paragraph to the "Why" section covering exactly this.

**Low — `// ` comments.** You're right that `CLAUDE.md` is explicit about this and I shouldn't have added them — removed every comment this PR introduced across all four files (the doc comments on `MigrationRemoveOrphanedTraffics` and `pickClientsSummary`, the inline explanations in both test files). The context that was in them now lives only in the commit messages and PR description.

**Suggestion — merge the type-only imports.** Done.

All Go tests (`go test ./...`, `go test -race ./internal/web/service/...`) and the full frontend suite (`typecheck`, `lint`, `test`, `build`, `build-storybook`) are green after these changes.

PR #6118, opened 2026-07-27
fix: preserve VLESS+WS inbound clients when clients table is empty

### LINKED ISSUE #6117: [Bug]: VLESS+WebSocket inbound generates config.json with settings.clients: null despite clients stored in the DB
## Describe the bug

When creating a VLESS + WebSocket inbound through the panel, the generated `config.json` shows `"clients": null` instead of the actual client list. This happens even though the database stores clients correctly. The gRPC runtime injection does not trigger for VLESS+WS protocol.

## To Reproduce

1. Add a new inbound: VLESS, port 48888, network=ws, security=none (or tls)
2. Add a client with UUID
3. Restart x-ui
4. Check `/usr/local/x-ui/bin/config.json` → `"clients": null`

## Expected behavior

Clients should be properly serialized as:
```json
"settings": {
  "clients": [{"id": "..."}],
  "decryption": "none"
}
```

## Actual behavior

```json
"settings": {
  "clients": null,
  "decryption": "none"
}
```

## Environment

- 3x-ui version: latest (installed via install.sh)
- Xray version: 26.7.11
- OS: Ubuntu 22.04
- Protocol: VLESS + WebSocket (both tls and none security affected)

## Workaround

Running a separate Xray instance with clients hardcoded in config:
```json
{"port":48888,"protocol":"vless","settings":{"clients":[{"id":"..."}]},"streamSettings":{"network":"ws","security":"none","wsSettings":{"path":"/ws"}}}
```

## Additional context

`GenXrayInboundConfig()` in `internal/database/model/model.go` calls `StripVlessInboundEncryption()` which only removes the `encryption` key. The Settings JSON with clients is correctly passed to `InboundConfig.Settings`. However the gRPC `AddUser`/`AlterInbound` injection that normally adds clients at runtime does not execute for VLESS+WS inbounds.

Ports using xhttp+tls (e.g., port 8443) also show `clients:null` in config.json, but their clients ARE injected via gRPC and connections work. The difference is that WS transport inbounds never receive the gRPC client injection.

talwayh1:
## Problem

When clients are defined in the inbound `settings` JSON (e.g., VLESS+WebSocket inbounds created via database/API) but not in the separate `clients` table, `GetXrayConfig()` overwrites the existing clients with `nil`, resulting in `config.json` showing `"clients": null`.

This causes VLESS+WS inbounds to reject all connections because Xray has no valid client IDs.

## Root Cause

In `internal/web/service/xray.go:224-228`:

```go
_, hadClients := settings["clients"]
mutated = hadClients || len(finalClients) > 0
if mutated {
    settings["clients"] = finalClients  // ← overwrites with nil when finalClients is empty
}
```

`finalClients` is declared as `var finalClients []any` (nil). When the `clients` table has no entries but settings already contains clients, `finalClients` (nil) overwrites the valid clients array.

## Fix

Preserve existing clients from settings when `finalClients` is empty:

```go
existingClients, hadClients := settings["clients"]
hasFinalClients := len(finalClients) > 0
mutated = hadClients || hasFinalClients
if mutated {
    if hasFinalClients {
        settings["clients"] = finalClients
    } else if existingClients != nil {
        settings["clients"] = existingClients
    }
}
```

Closes #6117

### DIFF
--- a/internal/web/service/xray.go
+++ b/internal/web/service/xray.go
@@ -221,10 +221,15 @@ func (s *XrayService) GetXrayConfig() (*xray.Config, error) {
 			settings["peers"] = wgPeers
 			mutated = true
 		} else {
-			_, hadClients := settings["clients"]
-			mutated = hadClients || len(finalClients) > 0
+			existingClients, hadClients := settings["clients"]
+			hasFinalClients := len(finalClients) > 0
+			mutated = hadClients || hasFinalClients
 			if mutated {
-				settings["clients"] = finalClients
+				if hasFinalClients {
+					settings["clients"] = finalClients
+				} else if existingClients != nil {
+					settings["clients"] = existingClients
+				}
 			}
 		}
 


MHSanaei:
Thanks for tracking this down to the exact lines — the `clients: null` symptom you identified is real, and 6f4cc1e5 now fixes that half: `finalClients` is built as an empty slice, so an inbound whose client list comes out empty serializes as `[]` rather than `null`.

I have not taken the fallback in this PR, and I want to be clear about why rather than leave it sitting. The automated review above is correct: `finalClients` being empty is not evidence that the `clients` table is out of sync. It is also exactly what happens when every client on an inbound is disabled, expired or over quota — that is the intended way the list becomes empty. Falling back to the stored `settings.clients` array in that state re-admits those clients into the generated config with their credentials intact, because `markClientsDisabledInSettings` only flips an `enable` field that xray-core has no concept of. On the next restart, clients the panel had cut off for exceeding their quota would be served again.

If the desynced-table state you hit is reproducible, that is the thing worth fixing, and I would very much like to see it. `ClientService.SyncInbound` rebuilds the `clients` / `client_inbounds` rows from the settings JSON and runs from `AddInbound`, `UpdateInbound`, and at startup from `MigrationRequirements` for every existing VMess/VLESS/Trojan/Shadowsocks/Hysteria inbound — so a path that survives all three is a genuine bug in one of them. #6117 is where that half is being tracked; if you can say how the inbound and client were created (panel UI, REST API, imported backup, or direct DB edit) and post the `clients` / `client_inbounds` rows for that inbound, it would narrow it down quickly.

Closing this PR is your call — happy to leave it open if you would like to rework it toward detecting the desync explicitly (e.g. only when `ListForInbound` returned zero rows *and* the settings array is non-empty, running those clients through the same enable/protocol filtering rather than copying the array verbatim).

PR #6121, opened 2026-07-28
fix(settings): keep the stored port when a port field is cleared

PathGao:
## Summary

Clearing a port field in panel settings (**panel port**, **subscription listen port**, **LDAP port**) no longer silently saves port `1`; the field snaps back to the last valid value instead.

## Why

Each of these `InputNumber`s fires `onChange(null)` when cleared. The handlers coerced that to `0`, then Ant Design clamped the empty field to `min=1` on blur, and the next save persisted port `1` — with no error at any point (the backend validators `gte=1,lte=65535` accept it). Consequences, reproduced end-to-end on `main` (values verified in the SQLite `settings` table):

- `subPort=1`: subscription links are generated with `:1` and the sub server cannot bind the port.
- `webPort=1`: the panel itself moves to port 1 on restart, locking the admin out until the port is repaired via the `x-ui` CLI.
- `ldapPort=1`: LDAP sync silently points at a dead port.

The other numeric settings are not affected: `smtpPort` already falls back to 587 on clear, and fields whose min value is a legitimate setting (e.g. session max age) merely clamp.

## Type of change

- [x] Bug fix

## Areas affected

## How was this tested?

Regression tests added to the existing `frontend/src/test/subscription-general-tab.test.tsx`:

- clearing the port field (change + blur) never reaches `updateSetting`;
- a typed port still passes through unchanged as `{ subPort: 8443 }`.

Manual verification on a local dev panel (Chromium):

1. Settings → Subscription → Listen Port shows `2096`; select-all + delete + blur: **before** the field showed `1` and Save stored `subPort=1`; **after** it restores `2096`.
2. Settings → General → Panel Port shows `2053`; same clear + blur: restores `2053` instead of `1`.
3. Normal edits (typing a port, stepping) still update and save as before.
4. `npm run lint`, `npm run typecheck`, `npm run build` pass; full frontend vitest suite passes.

## Screenshots / recordings

N/A — no visual change; behavior fix only (repro steps above).

## Breaking changes

None.

## Checklist

### DIFF
--- a/frontend/src/pages/settings/GeneralTab.tsx
+++ b/frontend/src/pages/settings/GeneralTab.tsx
@@ -170,7 +170,7 @@ export default function GeneralTab({ allSetting, updateSetting }: GeneralTabProp
 
             <SettingListItem paddings="small" title={t('pages.settings.panelPort')} description={t('pages.settings.panelPortDesc')}>
               <InputNumber value={allSetting.webPort} min={1} max={65535} style={{ width: '100%' }}
-                onChange={(v) => updateSetting({ webPort: Number(v) || 0 })} />
+                onChange={(v) => { if (v != null) updateSetting({ webPort: v }); }} />
             </SettingListItem>
 
             <SettingListItem paddings="small" title={t('pages.settings.panelUrlPath')} description={t('pages.settings.panelUrlPathDesc')}>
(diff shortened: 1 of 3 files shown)


PathGao:
Thanks for the review — all three findings addressed in the current head:

1. **webPort / ldapPort (Medium):** already fixed in this PR — the fix commit covers `GeneralTab.tsx` (`webPort`, `ldapPort`) alongside `subPort`, and the PR title/body were updated to "port fields" accordingly. The review snapshot predates that push.
2. **Regression test (Low):** added to the existing `subscription-general-tab.test.tsx` — clearing the port field (change + blur) asserts `updateSetting` is never called, and a companion subtest pins that a typed port still passes through as `{ subPort: 8443 }`.
3. **Redundant `Number(v)` (Suggestion):** dropped at all three call sites; after the null guard the handlers now forward `v` directly. Agreed the ~20-site `Number(v) || 0` pattern deserves a shared helper — left for a separate refactor as suggested (tracked on our side).

PR #6122, opened 2026-07-28
fix(ui): commit date-picker selections immediately instead of on confirm

PathGao:
## Summary

Date picked in the expiry-date picker now reaches the form state the moment it is clicked. Previously a date selected without pressing **OK** was silently discarded, so an inbound/client saved that way ended up with `expiryTime=0` (never expires).

## Why

With `showTime`, Ant Design's DatePicker stages a clicked date until the OK button confirms it. Closing the dropdown any other way — clicking elsewhere in the form, or hitting **Create**/**Save** directly — dropped the staged date without any hint. Because the **Now** shortcut commits in one click while calendar dates need the extra OK, users read this as "only the current time can be set; future dates don't stick".

Two failure paths were reproduced end-to-end on `main` (DB value checked after each):

1. Pick a future date → click **Create** directly → inbound saved with `expiryTime=0`.
2. Even with `needConfirm={false}` alone, the commit-on-close happens after the dropdown closes, so path 1 still lost the race against the submit. Propagating every selection through `onCalendarChange` closes that hole: the value is committed synchronously on each click.

## Intentional behavior change

Because every calendar click commits immediately, **Esc / clicking away no longer discards the selection** (verified manually: pick a date, press Esc — the value stays). Backing out of a mis-click now means clearing the field (✕) or picking another date. This trade-off is deliberate: a silent discard that writes "never expires" is strictly worse than a sticky mis-click that remains visible in the field. The array branch of `onCalendarChange` (a range-picker-only shape, required by the prop's type) normalizes to the first element rather than dropping the event.

## Type of change

- [x] Bug fix

## Areas affected

## How was this tested?

Regression tests added in `frontend/src/test/date-time-picker.test.tsx` (jsdom renders the AntD branch since `persian-calendar-suite` is stubbed in `setup.components.ts`):

- clicking a day cell commits the exact timestamp through `onChange` with no OK press anywhere in the test;
- the picker footer renders no OK confirm button.

Manual verification on a local dev panel (Chromium), `expiry_time` read from SQLite after each save:

1. Add Inbound → pick `2026-08-27` → click **Create** immediately (dropdown still open): stored `2026-08-27` (was `0` before the fix).
2. Pick a date → click elsewhere → **Create**: stored correctly.
3. **Now** shortcut: unchanged, commits instantly.
4. Esc after picking a date: value retained (see "Intentional behavior change").
5. Time-of-day columns: picking hour `05`, minute `30`, second `45` in turn accumulates to `05:30:45`; each click registers, selected cells persist, no column jump or reset.
6. Edit-mode changes and the clear (✕) action still work; the Jalali (`persian-calendar-suite`) path is untouched — it already commits on selection.
7. `npm run lint`, `npm run typecheck`, `npm run build` pass; full frontend vitest suite passes.

## Screenshots / recordings

N/A — the only visible change is that the redundant OK button is gone from the picker footer; selection behavior is otherwise identical.

## Breaking changes

None.

## Checklist


PathGao:
Thanks — all four findings addressed in the current head:

1. **Regression test (Medium):** added `frontend/src/test/date-time-picker.test.tsx` in the suggested style. The AntD branch renders in jsdom (Persian picker is stubbed in `setup.components.ts`); the test clicks a specific day cell and asserts `onChange` received that exact timestamp with no OK press anywhere, plus a second case asserting the footer renders no `.ant-picker-ok` at all. This pins both behaviors this fix rests on (`needConfirm={false}` and single-picker `onCalendarChange`) against a dependency bump.
2. **Esc semantics (Low):** manually verified — after clicking a date, Esc closes the dropdown and the value stays committed. Documented in the PR description under "Intentional behavior change" with the rationale (a sticky, visible mis-click beats a silent "never expires").
3. **Time-column round-trip (Low):** manually verified — picking hour `05`, then minute `30`, then second `45` accumulates to `05:30:45`; every click registers, the selected cells persist, and the columns do not jump or reset when the re-rendered `value` feeds back into the open panel.
4. **`Array.isArray` guard (Suggestion):** the prop's type is `Dayjs | Dayjs[]`, so the plain `onChange(next || null)` form does not typecheck. Replaced the drop-the-event guard with first-element normalization — `onChange((Array.isArray(next) ? next[0] : next) || null)` — so even the (unreachable for a single picker) array shape can no longer reintroduce the discard.

PR #6125, opened 2026-07-28
fix(ui): explain the REALITY client version gate and drop the impossible placeholder

PathGao:
## Summary

Surface the hidden REALITY client-version gate in the panel: add tooltips to **Min Client Ver** / **Max Client Ver** explaining that an empty minimum is not "no limit", remove the impossible `25.9.11` placeholder from Max Client Ver, and add the pitfall to the REALITY docs in all four languages.

## Why

When `minClientVer` is left empty, 3x-ui passes it through untouched — and Xray-core then falls back to a hardcoded built-in minimum (`infra/conf/transport_security.go`, currently `26.3.27`):

```go
} else {
    config.MinClientVer = []byte{26, 3, 27}
}
```

Third-party cores such as Mihomo and sing-box report versions far below that, so they fail REALITY verification even with a perfectly correct config. The user-visible symptom is timeouts in those apps while Xray-core based clients connect fine — and nothing in the panel or docs points at this field. (Real case: Hiddify/sing-box and Clash/mihomo both timing out against a fresh REALITY inbound; setting Min Client Ver to `1.0.0` immediately fixed both.) Upstream considers the gate intentional TLS-fingerprint-freshness protection (XTLS/Xray-core#6477 was closed as not planned), so this PR deliberately does **not** change any default behavior — it only makes the gate visible and explains the trade-off of lowering it.

Separately, the **Max Client Ver** placeholder was `25.9.11`, which is *below* the built-in minimum `26.3.27` shown next to it. A user who fills in both placeholders as written gets a version range that rejects every client. It looks like a stale example from before the 26.x gate was introduced. Since an empty max genuinely means "no upper limit", the placeholder is removed rather than replaced.

## Type of change

- [x] Bug fix (misleading placeholder producing an impossible version range)
- [x] Documentation

## Areas affected

## How was this tested?

- Local panel (`go run .` + built frontend): opened Add Inbound → Security → Reality, hovered both tooltip icons — the hint text renders correctly, the Min Client Ver placeholder still shows `26.3.27`, and the Max Client Ver placeholder is gone (verified via DOM: `input.placeholder === ""`).
- The new `minClientVerHint` / `maxClientVerHint` keys were added to **all 13 locale files**; a script asserted key presence and JSON validity across every file.
- `make verify` passes locally (gen diff clean, golangci-lint, Go tests, frontend lint/typecheck/tests/build, storybook build).
- No behavior change anywhere: no schema, wire-format, or backend changes; `frontend/src/generated` and `openapi.json` are untouched.

## Screenshots / recordings

The only visible changes are a question-circle tooltip icon next to both version fields (hover shows the explanation) and the removed Max Client Ver placeholder. Happy to attach a screenshot on request.

## Breaking changes

None.

Behavior notes for reviewers:

- The tooltip and docs quote the concrete built-in version (`26.3.27`) because that is what users need to recognize the symptom; both are worded with "currently / at the time of writing" so they age gracefully if the core bumps the gate. The value matches the existing Min Client Ver placeholder, which already tracked it.
- Removing the Max Client Ver placeholder is a deliberate visible change: empty is the correct "no upper limit" state, and any concrete example number next to a hardcoded lower gate risks going stale again.

## Checklist

- [ ] I added or updated tests for the new behavior (not applicable — tooltip copy, a placeholder removal, and docs; no behavior to pin)
- [x] I updated the Wiki / README / API docs if user-facing behavior changed (REALITY docs page, all 4 languages).


PathGao:
Thanks for the review — all findings addressed in 01cc6d6, point by point:

**1. `maxClientVerHint` re-establishing the "empty means unset" model (Medium)** — Agreed, this was the sharpest catch: the hint contradicted the PR's own premise. Reworded in all 13 locales to name the *effective* minimum: "Empty means no upper limit. If set, it must not be lower than the effective minimum — Min Client Ver, or Xray-core's built-in minimum when that field is empty — otherwise every client is rejected."

**2. Hardcoded `26.3.27` (Low)** — Rephrased everywhere (13 locales + 4 docs pages) to tie the number to the running core: "the built-in minimum of the core build you run (26.3.27 in current releases)". On verification: the constant is in `infra/conf/transport_security.go` of the pinned module — with the module cache populated, `MinClientVer = []byte{26, 3, 27}` sits in the `else` branch of the empty-string check (line ~117 in the pinned revision). Rendering the value from the core version the panel reports is the right long-term fix; leaving that as a follow-up since it turns a copy PR into a feature.

**3. zh punctuation + missing verb (Low)** — Fixed: full-width quotes around 不限制 matching the rest of the zh docs, and 也会 REALITY 验证失败 → 也会导致 REALITY 验证失败.

**4. Arabic idiom (Low)** — Fixed: both strings now open with تركه فارغًا, matching the repo's existing construction (ar-EG.json:522 et al.).

**5. Save-time validation (Suggestion)** — Agreed it's the real fix for the impossible range, but it changes behavior (new validation path + error keys across 13 locales + tests), so keeping this PR copy-only and deferring to a separate PR.

**6. `x.y.z` placeholder on Max Client Ver (Suggestion)** — Applied. Format-only, so it conveys the expected shape without a number that can go stale.

Re-verified after the changes: all 13 locale files parse and carry both keys, `npm run typecheck` / `npm run build` pass, and the new copy renders in a local panel (tooltip text and the `x.y.z` placeholder confirmed in the browser).

PR #6126, opened 2026-07-28
feat(ui): validate the REALITY client version range at save time

PathGao:
## Summary

Follow-up to #6125, as promised in its review thread: validate the REALITY **Min Client Ver** / **Max Client Ver** fields at save time — format matching xray-core's parser, plus a cross-field check that a non-empty max is not below a non-empty min — so an inverted range that rejects every client can no longer be saved.

## Why

#6125 documented the client-version gate, but a tooltip only helps a user who hovers it. The automated review on that PR pointed out the remaining hole: nothing stopped a user from saving `maxClientVer = 25.9.11` against the effective minimum `26.3.27` — a range that rejects every client, with no feedback at save time and a hard-to-diagnose "clients time out" symptom later.

The validators mirror xray-core's config loader exactly (`infra/conf/transport_security.go`): up to three dot-separated numeric parts, each 0-255, empty meaning "not set". Short versions are padded before comparison the way the core pads them (`26` == `26.0.0`). A malformed min is reported by the min field itself and skipped by the max comparison, so each field shows one precise error.

## Type of change

- [x] Bug fix (an impossible version range could be saved silently)

## Areas affected

## How was this tested?

- 9 new unit tests in `frontend/src/test/stream-wire-normalize.test.ts` covering: core-style accepts (`26`, `26.3`, `26.3.27`, `0.0.255`), core-style rejects (four parts, part > 255, non-numeric, empty parts), the exact stale-placeholder trap from #6125 (`25.9.11` vs `26.3.27`), short-version padding, and precedence of format errors over range errors.
- Local panel end-to-end: filling `26.3.27` / `25.9.11` and pressing Create blocks the save with a field-level "Max Client Ver must not be lower than Min Client Ver"; a malformed value shows the format error; correcting the values and resubmitting saves cleanly.
- `npm run lint`, `npm run typecheck`, `npm run test`, `npm run build` pass; both new error keys are present in all 13 locale files (script-asserted).

## Screenshots / recordings

The only visible change is a standard field-level validation error under the offending field when saving an invalid version or an inverted range — same presentation as the existing REALITY target validation.

## Breaking changes

None. Previously-savable invalid values are now rejected at the form; values already stored in the DB are unaffected.

Behavior notes for reviewers:

- Validation follows the existing REALITY-target precedent in the same file: pure validators in `src/lib/xray/stream-wire-normalize.ts` returning i18n keys, wired via `rules.validate`. Errors surface when the user presses Create (the modal validates via `trigger()`), matching how every other field in this form behaves.
- The comparison is only between the two fields when both are non-empty. When min is empty the effective floor is the core's built-in minimum, which depends on the Xray-core build the panel runs — the frontend cannot know it reliably, so that case stays a tooltip/docs concern (#6125) rather than a hard block.

## Checklist

- [x] I updated the Wiki / README / API docs if user-facing behavior changed (no doc change needed; #6125 already documents the gate).


PathGao:
Thanks — addressed in e6b0fcb:

**1. Trim asymmetry (Medium)** — Confirmed and fixed with your suggested shape: a value that differs from its trimmed form is rejected (`'26.3.27 '`, `' 26.3.27'`, and whitespace-only `' '` all now return the format error; tests added). This keeps the save path untouched and makes the form's green state match what the core will actually load.

**2. Stale cross-field error (Low)** — Applied the optional fix via `onAfterChange` on the min field: when max already shows an error, a min edit re-triggers max's validation, so correcting the min clears the message immediately. Gated on an existing error so pre-submit typing surfaces nothing early (the form stays onSubmit-mode otherwise).

**3. Server-side gap (Suggestion)** — Agreed on the framing: the guarantee is "the panel form cannot save an impossible range". A server-side check in the inbound service is a separate change; tracked as a follow-up candidate on our side.

**4. Soft warning for max below the built-in floor with empty min (Suggestion)** — Deliberately not doing this one: it would put the core's internal constant into executable comparison logic, where staleness produces a *wrong warning* rather than stale prose. The #6125 review's own concern about hardcoding cuts harder against logic than copy. The tooltip on both fields already names the effective-minimum trap; if the panel ever learns the running core's actual floor (the dynamic-threshold follow-up), a warning becomes safe to build on top of that.

PR #6127, opened 2026-07-28
refactor(ui): share one onNumber handler for numeric setting inputs

PathGao:
## Summary

Extract the null-safe InputNumber change handling introduced with the cleared-port fix (#6121) into a shared `onNumber(apply)` helper, and convert all eighteen null-unsafe numeric onChange sites in the settings and xray pages to it (sixteen `Number(v) || 0` sites, the hand-rolled guard on the observatory sampling field, and the smtp port's undocumented fallback — see review thread). Follow-up to the automated review on #6121, which flagged the idiom as the drift-prone root pattern.

## Why

Ant Design's InputNumber reports a cleared field as `null`. The `Number(v) || 0` idiom turns that into a stored `0` — or, combined with `min={N}`, into a silently clamped value, which is exactly how the cleared-port bug happened. #6121 fixed the three port fields with an inline `if (v != null)` guard, but the same idiom remained at sixteen other call sites, so every new numeric setting copies whichever neighbor it was pasted from.

`onNumber(apply)` centralizes the semantic: null / empty / NaN change events are ignored (the field snaps back to its stored value on blur), numeric events — including `0` typed explicitly and stringMode strings — pass through unchanged.

One site intentionally keeps a different semantic and is left alone: the Telegram notify interval clamps through `Math.max(1, …)`, a visible floor. The smtp port's `|| 587` fallback was initially kept as "deliberate", but the review correctly noted nothing documents that intent and clearing silently overwrites a configured non-standard port — it is now converted like the other port fields (a one-line revert if the maintainer wants the reset affordance back).

## Type of change

- [x] Bug fix (see behavior note below for the one deliberate change)

## Areas affected

## How was this tested?

- New unit tests for the helper (`frontend/src/test/on-number.test.ts`): numeric pass-through including zero and negatives; null/undefined/NaN ignored. The helper is number-only by design (no `stringMode` input exists in the repo).
- New `general-tab.test.tsx`: clearing `pageSize` calls nothing and the input snaps back to the stored value on blur; typing `0` forwards `{ pageSize: 0 }`.
- The #6121 regression tests (`subscription-general-tab.test.tsx`) pass through the helper, now also asserting the blur snap-back display value.
- Full frontend suite: 100 test files / 906 tests pass; `npm run lint`, `npm run typecheck`, `npm run build` pass.

## Screenshots / recordings

N/A — no visual change; only what a cleared field writes.

## Breaking changes

None.

Behavior note for reviewers: for the thirteen non-port fields (e.g. `sessionMaxAge`, `pageSize`, DNS `poolSize`, outbound observatory interval), clearing the input previously stored `0`; it now keeps the stored value, consistent with the port fields. Zero remains reachable by typing `0`. If a maintainer prefers the old store-zero behavior for any specific field, that site can trivially opt out by not using the helper.

## Checklist

- [x] I updated the Wiki / README / API docs if user-facing behavior changed (not applicable — no documented behavior changes).

### DIFF
--- a/frontend/src/pages/settings/EmailTab.tsx
+++ b/frontend/src/pages/settings/EmailTab.tsx
@@ -3,6 +3,7 @@ import { useTranslation } from 'react-i18next';
 import { Alert, Button, Input, InputNumber, Select, Space, Switch, Tabs } from 'antd';
 import { MailOutlined, SendOutlined, SettingOutlined } from '@ant-design/icons';
 import { HttpUtil } from '@/utils';
+import { onNumber } from '@/utils/onNumber';
 import type { AllSetting } from '@/models/setting';
 import { SettingListItem } from '@/components/ui';
 import { EmailNotifications } from '@/components/ui/notifications/EmailNotifications';
@@ -64,7 +65,7 @@ export default function EmailTab({ allSetting, updateSetting }: EmailTabProps) {
 
             <SettingListItem paddings="small" title={t('pages.settings.smtpPort')} description={t('pages.settings.smtpPortDesc')}>
               <InputNumber value={allSetting.smtpPort} min={1} max={65535} style={{ width: '100%' }}
-                onChange={(v) => updateSetting({ smtpPort: Number(v) || 587 })} />
+                onChange={onNumber((v) => updateSetting({ smtpPort: v }))} />
             </SettingListItem>
 
             <SettingListItem paddings="small" title={t('pages.settings.smtpUsername')} description={t('pages.settings.smtpUsernameDesc')}>
--- a/frontend/src/pages/settings/GeneralTab.tsx
+++ b/frontend/src/pages/settings/GeneralTab.tsx
@@ -17,6 +17,7 @@ import {
 } from '@ant-design/icons';
 import type { AllSetting } from '@/models/setting';
 import { HttpUtil, LanguageManager } from '@/utils';
+import { onNumber } from '@/utils/onNumber';
 import { SettingListItem } from '@/components/ui';
 import { useMediaQuery } from '@/hooks/useMediaQuery';
 import { catTabLabel } from './catTabLabel';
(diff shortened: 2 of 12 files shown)


PathGao:
Thanks — all six addressed in 26ca7db, with one PR-description correction:

**1. ObservatorySettingsTab sampling (Low)** — Converted. It was indeed a third hand-rolled variant inside the stated scope.

**2. smtpPort fallback (Low)** — You're right that "deliberate" was my inference, not documented intent. Converted to `onNumber` for consistency with the other port fields: clearing now keeps the configured port instead of silently shipping 587. The PR description's exclusion list is updated — the Telegram `Math.max` clamp is now the only intentional exception (that one *is* visibly a floor, not a fallback constant). If the maintainer wants the 587 reset affordance back, it's a one-line revert on that site.

**3. GeneralTab test coverage (Low)** — Added `general-tab.test.tsx`: clearing `pageSize` calls nothing and the input snaps back to the stored value; typing `0` forwards `{ pageSize: 0 }` (pinning that zero stays reachable by typing).

**4. Docblock overpromise (Suggestion)** — Softened to describe what the helper does. The remaining `Number(v) || 0` sites outside settings/xray (clients modals, FallbacksCard — mostly RHF `transform` paths with Zod validation on submit, a different semantic) plus a possible `no-restricted-syntax` lint rule are tracked as a follow-up on our side rather than stretched into this PR.

**5. String branch (Suggestion)** — Took your first option: the helper is now typed `number | null | undefined`, the string parsing and its test are gone, and the docblock states it is number-only and not `stringMode`-safe. The whitespace concern disappears with the branch.

**6. Blur snap-back unverified (Suggestion, low confidence)** — Verified with the one-line assertion you suggested, in both `subscription-general-tab.test.tsx` and the new GeneralTab test: after clear + blur the input's displayed value is the stored one. The desync you reasoned about does not occur — rc-input-number resyncs to the controlled value on blur.

PR #6128, opened 2026-07-28
feat(ui): tag settings that sit at their shipped default value

PathGao:
## Summary

Add a small grey **Default** tag next to numeric settings whose current value equals the shipped default, backed by a new `POST /panel/api/setting/factoryDefaults` endpoint that exposes only browser-safe default values. Applied to eleven fields on the General and Subscription tabs (panel/sub/LDAP ports, session age, page size, expiry/traffic thresholds, LDAP client defaults).

## Why

A field showing `2096` reads identically whether the install never configured it or the operator explicitly saved `2096`. Newcomers cannot tell which knobs they have touched — and since #6121 a cleared port snaps back to its stored value, so there is no visual way back to "unset". The tag answers the question users actually have: *"is this knob at its factory position?"*

Two design decisions worth stating:

- **Value semantics, not provenance.** The tag compares the current value against the shipped default; it does not ask whether a DB row exists. A stored 2096 and a fallback 2096 behave identically at runtime, so they read identically in the UI. This also makes the tag purely client-side and live — type `8443` and it disappears as you type, type `2096` and it returns — and it stays truthful across upgrades (if a future default changes, a pinned old value simply stops matching).
- **The endpoint is field-filtered, not the raw map.** `defaultValueMap` contains per-install material (`secret`, `panelGuid`, node mTLS keys) that deliberately never reaches the settings UI. `GetFactoryDefaults` iterates the `entity.AllSetting` json field set instead, and additionally drops the redacted credential fields (`tgBotToken`, `twoFactorToken`, `ldapPassword`, `smtpPassword`), so nothing sensitive can leave the server. A Go test pins both the inclusion (`webPort` → `2053`, `subPort` → `2096`) and every exclusion.

## Type of change

- [x] New feature

## Areas affected

## How was this tested?

- Go: `internal/web/service/setting_factory_defaults_test.go` — browser-safe keys present with expected values; `secret`, `panelGuid`, mTLS material, `xrayTemplateConfig`, and credential fields all absent. Runs without a DB (pure reflection over the field set).
- Frontend: `default-setting-tag.test.tsx` — type-aware value comparison (number/bool/string), tag renders when equal, nothing when different, nothing while defaults are unloaded. `renderWithProviders` gained a `QueryClientProvider` so component tests can seed the defaults cache.
- Live panel: on the Subscription tab, Listen Port at `2096` shows the tag; typing `8443` removes it as you type; typing `2096` brings it back. Verified the network call and that the response contains no sensitive keys.
- `make gen` diff committed (new endpoint registered in `endpoints.ts`); full frontend suite (101 files / 918 tests), `lint`, `typecheck`, `build`, and Go tests pass locally.

## Screenshots / recordings

The visible change is a small grey `Default` tag after the setting title when the value equals the shipped default. Happy to attach a screenshot on request.

## Breaking changes

None. The endpoint is additive; existing clients are unaffected. The tag is display-only and changes no editing or saving behavior.

Notes for reviewers:

- The existing `/setting/defaultSettings` endpoint returns *current effective* values (stored-or-default), which cannot distinguish the two states this feature is about — hence the separate, deliberately named `factoryDefaults`.
- v1 scope is the eleven numeric fields where the confusion has actually bitten (ports and thresholds). Extending to other fields is a one-prop addition per site (`badge={<DefaultSettingTag …/>}`). A "restore default" affordance is intentionally left to a follow-up so this PR stays display-only.

## Checklist

- [x] I updated the Wiki / README / API docs if user-facing behavior changed (endpoint registered in `endpoints.ts` → OpenAPI docs).


PathGao:
Thanks — all eight addressed in a4341d7:

**1. Accessible name regression (Medium)** — Fixed with your span approach: `titleId` now lives on a span around the title text only, the badge is its sibling. Field names read clean again and `getByLabelText` stays stable regardless of whether defaults are loaded.

**2. Three copies of the defaults (Medium)** — Took the "server is authoritative, guard the drift" option, and the guard earned its keep immediately: a new contract test (`frontend/src/test/factory-defaults-contract.test.ts`) parses the string literals out of `defaultValueMap` in `setting.go` and asserts every key shared with the frontend `AllSetting` class matches through `matchesFactoryDefault` itself — all three homes of a default pinned by one test. On its first run it caught two *pre-existing* drifts you predicted nothing would catch: `tgEnabledEvents` and `smtpEnabledEvents` defaulted to `''` in the class but `login.attempt,cpu.high` on the server. Both aligned to the server value.

**3. `Number('')` coercion (Low)** — Applied your guard verbatim, plus the boolean mirror: only literal `'true'`/`'false'` count, anything else is "unknown → no tag". Tests added for blank, whitespace and junk defaults against `0` and `false`.

**4. Test structure (Low)** — Both Go tests are now table-driven `t.Run` subtests with `t.Errorf` (all mismatches surface under `-shuffle=on`), and the structural invariant you asked for exists: every key `GetFactoryDefaults` returns must be an `entity.AllSetting` json tag and must not be in the credential deny-list.

**5. Doc comment overstatement (Low)** — Reworded to describe the projection mechanism (AllSetting json tags minus the deny-list) instead of claiming a blanket "no per-install material" property; you're right that `webBasePath`'s default is env-derived.

**6. i18n indent/placement (Low)** — Re-indented to 6 spaces and moved to the head of `pages.settings`, in all 13 files.

**7. Validation fallback (Suggestion)** — Tightened: the fetch now returns `{}` when the schema doesn't validate, instead of passing the raw object through under a false type.

**8. Partial coverage (Suggestion)** — Extended to `smtpPort`, so the tag now covers every plain numeric `SettingListItem` field that has a shipped default. The remaining keys you listed (`smtpCpu`/`smtpMemory`/`tgCpu`/`tgMemory`, `warpUpdateInterval`) live inside composite widgets (the per-event notification threshold editor, the WARP modal's RHF form), where a title badge has no natural anchor — bringing the signal there needs a different UI treatment, so it's deferred deliberately rather than half-done. Within the settings lists, absence of the tag now does reliably mean "not at the shipped default".

PR #6129, opened 2026-07-28
chore(lint): forbid the Number-or-clamp idiom in direct-write settings pages

PathGao:
## Summary

Add a scoped ESLint `no-restricted-syntax` rule that rejects the `Number(v) || N` idiom inside `onChange` attributes in the settings and xray pages — the drift-prevention follow-up promised in #6127's review thread — and convert the one remaining match (the Telegram notify interval) onto the shared `onNumber` helper.

## Why

#6127 centralized the null-safe handling, but as its review noted, nothing *enforced* the helper: the next numeric field pasted from an unconverted neighbor would reintroduce the cleared-port bug (#6121) silently. The rule makes that a lint error at the exact blast site:

```
A cleared InputNumber must not write a synthetic value; wrap the handler
with onNumber() from @/utils/onNumber (see #6127).
```

**Scope is deliberate.** The rule applies only to `src/pages/settings/**` and `src/pages/xray/**`, where onChange handlers write straight into persisted state. The clients/inbounds form modals stage values behind Zod validation, and several of their fields carry a real clear-means-zero semantic (`totalGB` 0 = unlimited, `limitIp` 0 = no limit) — converting those would break intended UX, so they are outside the rule's directories, with the reasoning recorded in a comment on the rule itself.

With the Telegram interval converted (its `Math.max(1, …)` floor is kept for typed values; clearing now keeps the stored count instead of writing 1), the rule lands with **zero suppressions** — no `eslint-disable` anywhere.

## Type of change

- [x] Build / CI / tooling
- [x] Refactoring (one site converted; see behavior note)

## Areas affected

## How was this tested?

- Verified the rule fires: reverting the Telegram site to the old idiom produces the lint error at the expected position (`TelegramTab.tsx:127`); the converted code lints clean.
- `npm run lint` passes with no suppressions; full frontend suite (100 files / 913 tests), `typecheck`, `build` pass.

## Screenshots / recordings

N/A — no visual change.

## Breaking changes

None.

Behavior note: clearing the Telegram notify-interval count previously wrote `1`; it now keeps the stored value (consistent with every other converted numeric field), and `Math.max` still floors typed values at 1.

## Checklist

- [x] I added or updated tests for the new behavior (not applicable — the rule is its own guard; the helper's behavior is already pinned by the #6127 tests)

### DIFF
--- a/frontend/eslint.config.js
+++ b/frontend/eslint.config.js
@@ -54,18 +54,36 @@ export default [
     },
   },
   {
-    // The settings and xray pages write numeric input changes straight into
-    // state, so `Number(v) || N` turns a cleared field into a stored N (the
-    // cleared-port bug, #6121). Handlers here go through onNumber()
-    // (src/utils/onNumber.ts) instead. Form modals that stage values behind
-    // Zod validation keep their own clear-means-zero semantics and are
-    // deliberately outside this rule's scope.
+    // The settings and xray pages write numeric InputNumber changes straight
+    // into state, so a null-collapsing handler (`Number(v) || N`, or the
+    // ternary `typeof v === 'number' ? v : N`) turns a cleared field into a
+    // stored N — the cleared-port bug, #6121. Handlers here go through
+    // onNumber() (src/utils/onNumber.ts) instead. Known limit: a handler
+    // extracted into a variable and passed as onChange={handler} is not
+    // matched; the inline shapes below are the ones that drift in practice.
     files: ['src/pages/settings/**/*.tsx', 'src/pages/xray/**/*.tsx'],
     rules: {
       'no-restricted-syntax': ['error', {
-        selector: 'JSXAttribute[name.name="onChange"] LogicalExpression[operator="||"] > CallExpression[callee.name="Number"]',
+        selector: 'JSXElement[openingElement.name.name="InputNumber"] JSXAttribute[name.name="onChange"] LogicalExpression[operator="||"] > CallExpression[callee.name="Number"]',
+        message: 'A cleared InputNumber must not write a synthetic value; wrap the handler with onNumber() from @/utils/onNumber (see #6127).',
+      }, {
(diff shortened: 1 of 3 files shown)


PathGao:
Thanks — all four addressed in 1365a63:

**1. Selector narrower than the claim (Medium)** — Both remedies applied. The rule now also matches the ternary `typeof v === 'number' ? v : N` form and the `?? N` form, and `BasicsTab`'s Happy Eyeballs try-delay — the one genuine direct-write ternary you found — is converted to `onNumber` (clearing keeps the stored delay; `0` stays reachable by typing it). Verified the widened selector fires by reinjecting the ternary: it errors at `BasicsTab.tsx:222` with the expected message. `BalancerFormModal`'s ternary is handled by the scope fix below rather than conversion. The extracted-handler shape (`onChange={handler}`) remains out of reach — matching bare `Number()` calls file-wide would flag parsing helpers, the false-positive cost you noted — so the rule's comment now names that limit instead of implying completeness.

**2. Comment vs. xray modals (Low)** — You're right that the stated rationale didn't match the glob. Took your explicit-exclusion option: a following config object turns the rule off for `src/pages/xray/**/*Modal.tsx` with the Zod-staging rationale attached, so a future clear-means-zero field in those modals needs no suppression and the comment tells the truth.

**3. Message/selector mismatch (Low)** — Took your tightening option: all three selectors are now anchored under `JSXElement[openingElement.name.name="InputNumber"]`, so the message can never point a `ChangeEvent` handler at a number-typed helper.

**4. Decimal interval round-trip (Suggestion)** — Applied `precision={0}` on the interval `InputNumber`. A typed `2.5` can no longer compose an `@every 2.5m` that `EVERY_RE` refuses to re-parse into `every` mode on reload.

On the failing `frontend` check: the failure is `ConfigBlock.stories.tsx > Collapsed` (913/914 passing), a storybook browser-mode test untouched by this diff — the same load-flakiness family we've hit locally on other stories. The push above re-runs the suite.

PR #6130, opened 2026-07-28
ci: attach provenance and SBOM attestations to the published images

kobihikri:
Hi, and thanks for 3x-ui.

`.github/workflows/docker.yml` publishes to both `hsanaeii/3x-ui` on Docker Hub and `ghcr.io/mhsanaei/3x-ui` on every tag, but the pushed manifests carry no provenance or SBOM attestation. Someone pulling the image cannot check that it was built by this workflow, from this repository, at that tag.

That felt worth raising for 3x-ui in particular. It is a panel people deploy on a VPS, usually straight from the image with a one-line install, and it holds the credentials for everything it fronts. Where the image came from is exactly the thing a deployer would want to be able to confirm.

The change is two lines on the single build step:

```yaml
        with:
          ...
          push: true
          provenance: mode=max
          sbom: true
```

I used BuildKit's own attestation support rather than a separate signing action for a specific reason: your step pushes one build to two registries at once, and BuildKit attaches the attestation to the image manifest itself, so Docker Hub and GHCR both get it with no extra plumbing and no second credential. It also means **no permissions change** — `contents: read` and `packages: write` stay exactly as they are, and nothing needs `id-token`.

After a release, anyone can check with:

```
docker buildx imagetools inspect ghcr.io/mhsanaei/3x-ui:<tag> --format '{{ json .Provenance }}'
```

Two things worth knowing before merging:

- `mode=max` records the full build including build arguments. That is normally right for a public image; `provenance: true` produces a smaller record if you would rather.
- Attestations add an extra manifest to the index. Docker Hub and GHCR both handle it, but it is worth knowing if a mirror ever sits in front.

I have not claimed any SLSA level — the attestation is what BuildKit produces, and how the whole build should be characterised is your call.

Disclosure: I used AI assistance to help spot this and prepare the change, and I read the workflow and both push targets myself.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -55,6 +55,8 @@ jobs:
         with:
           context: .
           push: true
+          provenance: mode=max
+          sbom: true
           platforms: linux/amd64,linux/arm64/v8,linux/arm/v7,linux/arm/v6,linux/386
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}


kobihikri:
**Correction — I got a fact wrong in this PR, and I would rather flag it myself than let it sit.**

I wrote that the pushed manifest "carries no provenance or SBOM attestation". That is half wrong, and the wrong half matters.

**Provenance is already there.** For *public* repositories, `docker/build-push-action` adds provenance attestations with `mode=max` by default — [Docker's documentation](https://docs.docker.com/build/ci/github-actions/attestations/) states it plainly: *"Public repos: provenance attestations with `mode=max` are automatically added"*. I checked published images and they do already carry attestation manifests. So the `provenance: mode=max` line in my diff makes existing behaviour **explicit**; it does not add anything new.

**The SBOM is genuinely new.** That part stands — the same page says *"SBOM attestations aren't automatically added to the image"*, and `sbom: true` is what enables them.

I also wrote in the caveats that `provenance: true` gives "a smaller record". That is wrong as well: `true` resolves to max on a public repo, and the smaller setting is `provenance: mode=min`.

**So the honest description of this PR is:** it adds an SBOM attestation, and pins the provenance mode explicitly instead of relying on the default. Both are still defensible — an explicit line means the behaviour will not change quietly if the default ever does — but it is a smaller change than my description implied, and you should judge it on that basis rather than on what I originally wrote.

Happy to retitle and rewrite the description accordingly, or to close this if the SBOM alone is not worth the diff to you. Either is fine — just say which and I will act on it.

Apologies for the inaccuracy. It was caught by a maintainer reviewing the same change on another project, and they were right to.

PR #6131, opened 2026-07-29
chore(build): stop shipping production sourcemaps inside the binary

PathGao:
## Summary

Disable `build.sourcemap` for the production Vite build. Everything in `internal/web/dist` is embedded into the Go binary via `embed.FS`, so the 112 `.map` files (18MB, 72% of dist) currently ship inside every release download.

## Why

Nothing consumes the maps in production: the served pages never reference them and no error-reporting pipeline reads them. `npm run dev` serves its own sourcemaps regardless of this build flag, so developer experience is unchanged. Anyone needing to debug a production bundle can flip the flag locally — the config comment says so.

Measured: `internal/web/dist` 25MB → 6.7MB, which comes straight off the embedded payload of every release binary and every panel download/update.

## Type of change

- [x] Build / CI / tooling

## Areas affected

## How was this tested?

- `npm run build` → zero `.map` files in dist, panel serves and loads normally from the rebuilt embed.
- No test/lint/gen surface touched.

## Screenshots / recordings

N/A.

## Breaking changes

None.

## Checklist

- [x] I added or updated tests for the new behavior (not applicable — build flag)

### DIFF
--- a/frontend/vite.config.js
+++ b/frontend/vite.config.js
@@ -174,9 +174,10 @@ export default defineConfig({
     // Everything in outDir is embedded into the Go binary via embed.FS, so
     // production sourcemaps (~18MB across 112 files, 72% of dist) ship inside
     // every release build. Nothing consumes them there; `npm run dev` serves
-    // its own maps regardless of this setting. Flip locally to debug a
-    // production bundle.
-    sourcemap: false,
+    // its own maps regardless of this setting. To debug a minified bundle
+    // (including the XUI_DEBUG serve-from-disk path), build once with
+    // XUI_SOURCEMAP=true — no tracked-file edit to accidentally commit.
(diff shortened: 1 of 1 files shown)


PathGao:
Thanks for the approve — took the first suggestion anyway since it's strictly better:

**1. Env gate** — Applied verbatim: `sourcemap: process.env.XUI_SOURCEMAP === 'true'`. Verified both directions locally (`XUI_SOURCEMAP=true` → 112 maps, default → 0), and the comment now points at the env var instead of a tracked-file edit. Good catch on the `XUI_DEBUG` serve-from-disk path being the real consumer of a zero-diff escape hatch.

**2. `.map` in `staticAssetExts`** — Left as-is per your own recommendation; it's what keeps a map request from being answered with `index.html` whenever someone does build with maps.

PR #6132, opened 2026-07-29
chore(i18n): delete 230 dead translation keys and guard against new ones

PathGao:
## Summary

Remove 230 translation keys (11% of the set) that nothing in the frontend or Go sources references, from all 13 locale files — net **-3,107 lines** — and add a test that keeps the set clean: it fails when an en-US key has no reference anywhere, and when any locale's key set drifts from en-US.

## Why

Nothing ever detected a dead key: a missing key falls back to en-US at runtime and an unused key fails nothing, so leftovers of renamed features accumulate silently. The families removed tell the story — the email notifier reuses `tgbot.messages.*` for its subjects, so the whole `email.subject*`/`email.title*` set was orphaned; likewise the old `menu.*` theme entries, the clients bulk-copy strings, the `secAlert*` family, and 13 orphaned `tgbot.*` keys.

The guard test (`frontend/src/test/i18n-dead-keys.test.ts`) covers dynamic keys by harvesting concatenation and template-literal prefixes from the sources (`'pages.settings.' + msg` keeps its subtree alive), and its second assertion pins 13-locale key parity, so a forgotten translation surfaces at test time instead of as a silent en-US fallback.

Every locale shrinks by exactly the same 230 keys, so parity is preserved by construction.

## Type of change

## Areas affected

## How was this tested?

- The new guard test passes (0 dead keys, 13/13 parity) and demonstrably works: it caught 10 keys a first-pass scan had missed (the scan had treated `storybook-static` build output as source, which embeds the full translation set and masks dead keys).
- Full frontend suite: 101 files / 916 tests pass; `typecheck` and `build` pass; all `internal/web` + `internal/sub` Go test packages pass (15 ok).
- Verified each removed family by hand before deleting (e.g. the email subjects are genuinely built from `tgbot.messages.*` — `internal/web/service/email/subscriber.go`).

## Screenshots / recordings

N/A — no visible change; every removed key was unreachable.

## Breaking changes

None. Keys removed are unreferenced by any code path; runtime fallback behavior for existing keys is unchanged.

## Checklist


PathGao:
Thanks — this was the review that mattered. All four addressed in a2cc826:

**1. The 29 deleted live keys (High)** — Confirmed and restored, byte-for-byte from the parent commit, in all 13 locales. Your root-cause was exact: the template-literal harvester demanded a trailing dot, `remarkVars.desc${token}` ends mid-leaf, and deletion + guard shared the blind spot so CI stayed green on a real regression. The harvester now uses your suggested shape (≥1 dot required, trailing dot optional), and the restored subtree is prefix-alive under it. Thanks also for hand-re-enumerating the blast radius — with the fixed detector I can confirm the desc* family was indeed the only live casualty.

**2. Self-whitelisting comment (Medium)** — Fixed both ways: the test now excludes its own file from the scan (path equality against `import.meta.url`), and the comment describes the pattern without quoting a real prefix. `pages.settings.*` is now exempted only by the genuine `EmailTab`/`NotificationGroup` call sites.

**3. Substring masking (Low)** — Took your tokenize-once recommendation: the blob is tokenized into a `Set` of dotted runs and keys must match exactly. That unmasked 9 more dead keys (`pages.index.xraySwitchClick` — your example — plus `pages.clients.add/edit/delete/reset`, `pages.index.backup`, `pages.xray.restart`, `pages.xray.outbound.test`, `tgbot.day`). Each was verified by hand (quoted-string grep + dynamic-prefix check) before deletion, given how finding 1 went.

**4. Performance (Low)** — The token `Set` makes the per-key check O(1) and the prefix list is built once; the 2122 × 10MB substring scans are gone.

Net effect: −210 keys per locale (was −230), and the guard that enforces it no longer shares a blind spot with the cleanup it guards.

PR #6133, opened 2026-07-29
fix(api): authenticate GET /panel/api/openapi.json + pin the route registry to the router

PathGao:
## Summary

Add a two-way contract test between the hand-maintained API registry (`frontend/src/pages/api-docs/endpoints.ts`) and the routes the Gin server actually registers: every `/panel/api` route must have a registry entry, and every registry entry must correspond to a real route. On arrival the check found (and this PR fixes) one gap: `GET /panel/api/openapi.json` — the endpoint that serves the docs — was itself undocumented.

## Why

The repo's own agent guide calls the risk out explicitly: *"It is a hand-maintained registry — nothing checks it against the Go routes, so an omitted route silently vanishes from the docs."* The reverse failure exists too — an entry left behind after a route is removed documents an endpoint that 404s. Both directions are now test failures with actionable messages naming the exact route.

The test constructs the real router (`initRouter`) against a throwaway DB (`database.InitDB` in `t.TempDir()`, the `internal/sub` test pattern), so it exercises the same registration code production runs — no parallel route list to maintain. The registry side is parsed from `endpoints.ts` with a shape-checked regex that fails loudly if the file format changes.

## Type of change

- [x] Tests only
- [x] Documentation (one missing registry entry added)

## Areas affected

## How was this tested?

- Both new tests pass; before the registry fix, the forward test failed naming exactly `GET /panel/api/openapi.json`, and the reverse test found zero stale entries.
- `make gen` diff for the new entry is committed (`openapi.json` gains the path); `golangci-lint`, frontend `lint`/`typecheck` pass.

## Screenshots / recordings

N/A.

## Breaking changes

None.

## Checklist

- [x] I updated the Wiki / README / API docs if user-facing behavior changed (the missing endpoint is now in the API docs).


PathGao:
Thanks — this turned a docs test into a real fix. All four addressed in 8df4dde (PR title updated to lead with the auth fix):

**1. Unauthenticated openapi.json (Medium, security)** — Fixed. Moved the registration from the base-path group into the `/panel/api` group after `checkAPIAuth`, per your recommendation. Verified against a running panel: unauthenticated it now returns 404 exactly like `server/status` (was 200 serving the full admin surface + build version), and a logged-in browser session still fetches it 200, so `ApiDocsPage` is unaffected. Updated the registry summary to say it requires auth, so the Authentication section is no longer contradicted.

**2. Duplicate coverage / the old test's switch trap (Medium)** — Folded and deleted. You're right that "nothing else checks it" was wrong — `api_docs_test.go` checked the forward direction from source text, which is exactly why it couldn't see a route registered in `web.go`, and its fall-through `basePath = ""` default is the silent-misp­ath trap you describe. The router-based test is a superset, so I extended its scope to the surface the old one uniquely guarded (`/login`, `/logout`, `/csrf-token`, `/getTwoFactorEnable`, `/ws`) and removed `api_docs_test.go`. Net −211 lines in the diff.

**3. Double construction + leaked goroutines + global clobber (Medium)** — The server is now built once and shared across two `t.Run` subtests; cleanup calls `s.cancel()` (so the hub goroutine exits) and restores the previous `global.GetWebServer()` value.

**4. Brace-crossing parser (Low)** — Rewritten to locate each `method:` and pair it with the next `path:` sequentially, so a brace-bearing field between them can't drop an entry. It now also asserts the parsed count equals the declared `method: '` count and fails loudly on any mismatch — which caught the four `method: 'WS'` message-type rows (documentation pseudo-entries with `path: '→ type: …'`), now filtered by the leading-slash check.

PR #5167, opened 2026-06-10
refactor: focused service files, leaf subpackages, and an internal/ layout

MHSanaei:
## What & why

A structural reorganization of the Go backend so it's easier to navigate and follows an idiomatic, professional Go application layout. **No runtime behavior change** — the file splits are verbatim moves and the package moves are import-path-only. Best reviewed commit-by-commit (17 self-contained commits, each builds and tests green on its own).

## Highlights

**1. Split the three oversized service files** (verbatim; identical declaration set verified before/after):

| File | Before | After |
|---|---|---|
| `client.go` | 4455 lines | 10 focused `client_*.go` |
| `inbound.go` | 4100 lines | 10 focused `inbound_*.go` |
| `tgbot.go` | 3738 lines (incl. a 1246-line method) | 6 focused `tgbot_*.go` |

**2. Extract leaf service subpackages** — services the core doesn't depend on (so no import cycles): `web/service/{panel,integration,outbound,tgbot}/`. The mutually-coupled core (client ↔ inbound ↔ node ↔ setting) necessarily stays flat.

**3. Adopt an `internal/` layout** — moved `config, database, logger, mtproto, sub, util, web, xray` under `internal/` (149 import rewrites). `main.go` and `tools/` stay at the root; the module path is unchanged. Includes the non-Go couplings the compiler can't see: Vite `outDir`, the frontend's `web/translation` imports, Dockerfile COPYs, `.gitignore`/CI embed stubs, the openapigen package paths, and a repo-root-relative test path.

**4. Smaller cleanups & fixes**
- dedupe 3 near-identical "set one field by email" setters into a single helper
- `util/wireguard.go` → `util/wireguard/` (it was the lone file of the root `util` package)
- drop the redundant `sub` prefix from `sub/` filenames; `controller/xui.go` → `spa.go`
- move the orphan `sub_templates/README.md` → `docs/custom-subscription-templates.md`
- remove the long-obsolete `/xui → /panel` redirect middleware
- **fix:** tests no longer scatter `log/` directories through the source tree (`GetLogFolder()` now honors `testing.Testing()`)
- **fix:** dev-mode disk-serving paths (`os.DirFS("web/dist/...")`) the `internal/` move missed — production was unaffected (embedded FS); only `go run` with a live frontend build silently fell back to embed
- `.dockerignore`: corrected path and excluded the local `x-ui/` runtime dir from the build context

## Verification
- `go build ./...`, `go vet ./...`, `go test ./...` — all green
[...]

PR #6134, opened 2026-07-29
fix(sub): honor trustedProxyCIDRs before letting forwarded headers steer URLs

PathGao:
## Summary

Subscription URLs now apply `trustedProxyCIDRs` with the same trust boundary as panel-generated links. Before this change, a request reaching the subscription server could steer generated hosts and schemes through forwarded headers even when the setting was unset, empty, or still at its default; now those states trust only loopback, while a configured value trusts exactly its listed addresses and CIDRs.

This is an intentional default-behavior change. Deployments using a remote reverse proxy for the subscription endpoint must add that proxy to `trustedProxyCIDRs` or configure `subURI`; this matches the setting's documented meaning and the panel's existing behavior. Rejected forwarded headers are logged at debug level, with one process-level warning when they first affect URL generation.

## Tests

- Table-driven request coverage for unset, empty, factory-default, loopback, custom-CIDR, malformed, and IPv4-mapped proxy origins.
- Coverage that untrusted `X-Real-IP` cannot bypass the forwarded-header gate.
- The local checkout has no Go toolchain, so the Go suite could not be run here.

### DIFF
--- a/internal/sub/forwarded_trust.go
+++ b/internal/sub/forwarded_trust.go
@@ -12,48 +12,28 @@ import (
 	"github.com/mhsanaei/3x-ui/v3/internal/web/service"
 )
 
-// shippedTrustedProxyCIDRs is the factory default of the trustedProxyCIDRs
-// setting, read from the setting service rather than restated here so the gate
-// cannot drift from the value an unconfigured install actually carries.
 var shippedTrustedProxyCIDRs = sync.OnceValue(func() string {
 	return (&service.SettingService{}).GetFactoryDefaults()["trustedProxyCIDRs"]
 })
 
-var forwardedHeaderNames = []string{"X-Forwarded-Host", "X-Forwarded-Proto", "X-Real-IP"}
-
 var warnSuppressedForwardedOnce sync.Once
 
-// forwardedHeadersTrusted reports whether X-Forwarded-* headers on this request
-// may steer the generated subscription URLs. The gate engages only once the
-// operator has moved trustedProxyCIDRs off its shipped default; an install that
-// never touched the setting keeps the legacy behavior. The recover mirrors
-// internal/web/controller/util.go, where a setting read before InitDB panics
-// rather than returning an error.
 func (s *SubService) forwardedHeadersTrusted(c *gin.Context) (trusted bool) {
-	trusted = true
+	trusted = remoteAddrInCIDRs(c.Request.RemoteAddr, shippedTrustedProxyCIDRs())
 	defer func() {
 		_ = recover()
 	}()
 
 	configured, err := s.settingService.GetTrustedProxyCIDRs()
-	if err != nil {
-		return true
+	if err == nil && strings.TrimSpace(configured) != "" {
+		trusted = remoteAddrInCIDRs(c.Request.RemoteAddr, configured)
 	}
-	configured = strings.TrimSpace(configured)
-	if configured == "" || configured == shippedTrustedProxyCIDRs() {
-		return true
-	}
-	return remoteAddrInCIDRs(c.Request.RemoteAddr, configured)
+	return trusted
 }
 
-// warnSuppressedForwardedHeaders makes the gate diagnosable: an operator whose
-// subscription links start carrying the raw request host otherwise has nothing
-// connecting that to the setting they changed. The warning fires once per
-// process so a polling client cannot flood the log; every occurrence is logged
-// at debug level.
 func warnSuppressedForwardedHeaders(c *gin.Context) {
-	present := make([]string, 0, len(forwardedHeaderNames))
-	for _, name := range forwardedHeaderNames {
+	present := make([]string, 0, 3)
+	for _, name := range []string{"X-Forwarded-Host", "X-Forwarded-Proto", "X-Real-IP"} {
 		if c.GetHeader(name) != "" {
 			present = append(present, name)
 		}
@@ -68,9 +48,6 @@ func warnSuppressedForwardedHeaders(c *gin.Context) {
 	})
 }
 
-// remoteAddrInCIDRs reports whether remoteAddr falls inside any CIDR or bare
-// address in the comma-separated list, mirroring the panel-side check in
-// internal/web/controller/util.go.
 func remoteAddrInCIDRs(remoteAddr, cidrs string) bool {
 	host := remoteAddr
 	if h, _, err := net.SplitHostPort(remoteAddr); err == nil {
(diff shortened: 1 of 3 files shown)


PathGao:
Thanks — all eight addressed in 161da9d, plus a correction to the PR description:

**1. Per-request DB amplification (Medium, Performance)** — Fixed exactly as you wrote it: the gate is evaluated once and the closure captures the boolean. I measured it rather than assuming, with a GORM `Query().After("gorm:query")` counter around a single `ResolveRequest`: **6 → 1** query for a request from outside the boundary, **4 → 1** from inside it. So the path that previously did zero setting lookups now does one, not six.

**2. Compatibility guarantee narrower than claimed (Medium, Correctness)** — You're right, and the overclaim was mine. "Unchanged, byte for byte" holds only at the default value; an install that already customized `trustedProxyCIDRs` for `getRemoteIp` / fail2ban does change behavior, and a split-proxy setup will see its links flip. I took your recommendation: `warnSuppressedForwardedHeaders` now names the setting and the rejected origin. One deviation — a bare `logger.Warning` per request would let a polling client flood the log, so the warning fires **once per process** and every occurrence is logged at debug. If you would rather have it unconditional, or throttled per origin instead, that is a two-line change. The PR description now states the narrower guarantee instead of the old one.

**3. Two edges in the exact-string comparison (Low, Correctness)** — Case 2 (empty stored value) is now an explicit, pinned branch rather than an accident of `configured == ""`. I kept it on the permissive side deliberately: under this PR's semantics an empty value is *more* clearly "never expressed an opinion" than the default is, so treating it as unconfigured is the consistent reading. That does leave the divergence from the panel's `trustedProxyCIDRs()` you describe — it is a genuine consequence of (a), and it disappears entirely under (c). Worth noting the states are three, not two: an install that never touched the setting has **no row at all**, so it takes the `err != nil` fail-open branch, not the empty-string one. All three are separate rows in the table now, so the fail-open branch you flagged as uncovered is covered.

Case 1 (an operator deliberately typing the default string) is inherent to (a), as you say. I have not tried to paper over it — a whitespace variant being treated as declared while the exact string is not is the implicit-semantics cost, and it is an argument for (b) or (c) rather than something to patch inside (a).

**4. Missing panic guard (Low, Reliability)** — Added, mirroring `internal/web/controller/util.go` with a named return so the recover fails open. Your read of it is the one I agree with: the sub server starts after `InitDB` in the shipped binary and I could not construct a live path either, so it is defensive. The commit body now says so explicitly rather than leaving it unexplained.

**5. Fourth copy of the shipped default (Low, Maintainability)** — Taken. It now reads `SettingService.GetFactoryDefaults()["trustedProxyCIDRs"]`, wrapped in `sync.OnceValue` so the reflection cost is paid once per process rather than per request, per your note. The constant is gone. The failure mode you describe — a default change silently flipping every unconfigured install from trusting to enforcing — is now structurally impossible rather than merely caught by a test.

**6. Testing gaps (Low, Testing)** — All three:
- The three near-identical `ResolveRequest` bodies fold into one table over six states (no row, empty row, stored default, declared boundary rejecting an outside origin / accepting an inside one / rejecting an unparsable one), each asserting all four return values instead of a subset.
- Added `TestResolveRequest_GatesRealIPFallback`. `X-Real-IP` is the fallback for both the base host and the displayed header host, so it is a second lever on the same fields and deserved its own case rather than riding along on the `X-Forwarded-Host` one.
- `setTrustedProxyCIDRs` still writes through `database.GetDB()` — going through `UpdateAllSetting` would drag in full-settings validation for a one-key write — but it now reads the value back through `SettingService.GetTrustedProxyCIDRs()` and fails with an explicit "the key this helper writes has drifted" message. That closes the drift risk you named without the heavier write path.

I also re-ran the falsification check against the new table: reverting only the wiring in `ResolveRequest` turns exactly the three enforcement subtests red and leaves the three compatibility ones green.

**7. (a) vs (c) (Suggestion, Design)** — This is the finding that changed my view, and I verified both halves before acting on it. `resolveHost` does gate on `isTrustedForwardedRequest` with no escape hatch, and `git log -S "func resolveHost"` puts that in the tree since at least #5167 — so a default install behind a remote proxy is already getting the raw host from the panel's link endpoints today. And `trustedProxyCidrsDesc` does promise unconditional behavior in the shipped copy. Together those mean the trust inconsistency I set out to close is precisely the one (a) preserves, which is a real hole in my own framing.

Where I land differently is on what follows from it. Working through it again, what this PR treats as one defect is really two, and (a) and (c) do not fix the same one:

- **(i) the setting does not do what it says** — `trustedProxyCidrsDesc` promises the listed IPs are the ones allowed to set forwarded host, proto and client IP headers; an operator who fills that field gets it honored by the panel and ignored by the sub server. No attacker needed for this to be a defect.
- **(ii) a default install has no boundary at all** — anyone reaching the sub port can steer the headers.

|  | (i) setting not honored | (ii) default installs unprotected |
|---|---|---|
| (a) | fixed | untouched |
| (c) | fixed | fixed |

(ii) is the only thing (c) buys over (a), so I went and traced what it is worth rather than arguing it abstractly. The three tainted values land in: the node addresses in the subscription body (`host`), the `Profile-Web-Page-Url` response header (`scheme` + `hostWithPort`, which several client apps render as a clickable link), and the sub info page (`hostHeader`). No response-splitting variant — Go's `net/http` collapses newlines in header values; I checked rather than assumed. All three come back in the requester's own response, so third-party harm needs something further: a caching layer in front that does not key on the header, or a service fetching subscriptions on a user's behalf. Both real, neither default.

So (c) trades a certain cost — non-loopback-proxy installs at the default lose their proxied host on upgrade — for a conditional benefit. (a) takes the certain benefit and leaves the conditional one. That is the reasoning behind shipping (a); it is a deliberate trade rather than reluctance to touch the compatibility question.

None of which makes it my call. A maintainer may weigh (ii) higher than I do, and "this project should be strict by default" is a coherent answer. The description now carries both of your points and this breakdown, so @MHSanaei has the argument rather than my summary of it. If the answer is (c), the diff on top of this one is small — drop the default-vs-declared branch, fall back to the loopback default on an empty value so it matches `trustedProxyCIDRs()`, invert three table rows — and I will push it as soon as you say.

**8. Comment volume (Suggestion, Style)** — Taken. The "why not always enforce" argument and the recover justification are in the commit body now; the file keeps one contract comment per function.

PathGao:
Closing this revision to re-scope the compatibility boundary and resubmit a focused follow-up.

PR #6135, opened 2026-07-29
fix(sub): honor trustedProxyCIDRs before forwarded URLs

PathGao:
## Summary

Subscription URLs now honor an operator's customized `trustedProxyCIDRs` before forwarded headers determine their host or scheme. This aligns the subscription server with the setting's intended boundary without changing existing deployments that have no stored value, an empty value, or the shipped default.

The compatibility rule is deliberate: those untouched states retain the legacy forwarded-header behavior. Deployments that have declared a proxy boundary now receive the same protection for subscription links as they do for panel-generated links; rejected headers are diagnosable without permitting a polling client to flood warning logs.

## Tests

- Table-driven coverage for unset, empty, factory-default, custom-CIDR, malformed, and IPv4-mapped proxy origins.
- Coverage that untrusted `X-Real-IP` cannot bypass the forwarded-header gate.
- The local checkout has no Go toolchain, so the Go suite could not be run here.


PathGao:
Follow-up in 419ab8e1:

- Skip the `trustedProxyCIDRs` lookup when none of the three forwarded headers is present, with focused coverage for the short-circuit predicate.
- Replace the reflection-derived sentinel and duplicated panel default with `service.DefaultTrustedProxyCIDRs`.
- Document the subscription-link effect and `subURI` override in en/fa/ru/zh.

The default/empty compatibility carve-out remains deliberate: this PR applies a declared custom trust boundary without changing legacy deployments that have not opted into one. The raw default comparison is part of that compatibility policy; there is no persisted intent marker that could distinguish an untouched default from an intentional value.

I did not extract the proxy matcher here. The two current call sites have different input contracts (`RemoteAddr` versus an arbitrary forwarded-header candidate), and the subscription path now covers the standard `RemoteAddr` forms plus IPv4-mapped addresses. A shared helper would expand this focused fix without removing a verified production divergence.

A settings read error likewise preserves the selected compatibility fallback, and the existing one-per-process warning bounds logs on a polling endpoint.

Local Go tooling is unavailable in this workspace; the pushed commit will be validated by the PR CI.

PathGao:
CI follow-up in 2672b22b: the frontend failure was an existing Storybook axe violation in `ConfigBlock` (11px text resolved to 3.33:1 contrast in the light theme). This commit sets the light-theme text color to `#595959`, which exceeds the required 4.5:1 ratio against the rendered `#f8f8f8` background. Local lint and typecheck pass; browser-story execution remains CI-only because this workspace has no Playwright Chromium binary.

PR #6137, opened 2026-07-29
fix(database): create SQLite backup snapshots online

PathGao:
## Problem

SQLite downloads and SQLite migration exports previously ran a WAL checkpoint and then read the live database file. A commit between those operations could produce a copy that does not represent a single SQLite state.

## Change

- Create the copy with SQLite's online backup API before reading or dumping it.
- Use a private temporary directory beside the database and remove it through an explicit cleanup function.
- Sweep interrupted snapshot directories when SQLite starts.
- Use a single backup step in WAL mode; use bounded incremental steps for the supported DELETE journal mode.
- Bound connection acquisition and retry handling.
- Reuse one scheduled Telegram snapshot across all administrators.
- Remove the checkpoint helper, whose only production callers were these two paths.

## Regression coverage

- Snapshot integrity and exact committed rows while another goroutine writes.
- Timeout while the source pool has no available connection.
- Existing destinations are rejected without overwriting them.
- Journal-mode step selection and startup cleanup of interrupted snapshot directories.

Validated locally with `go test -race -shuffle=on -count=1 ./internal/database ./internal/web/service/tgbot`, `go vet` for those packages, and `golangci-lint run`.


PathGao:
> `SendBackupToAdmins` calls `GetDb()` once per admin... and cleanup relies entirely on `defer os.RemoveAll(...)`.

Fixed in a348c9dd. Scheduled Telegram backups now build one snapshot for the run and reuse its bytes for every administrator. Snapshot cleanup is represented by the helper's returned cleanup function, and SQLite startup removes interrupted `.x-ui-backup-*` directories.

> `DELETE` is a supported journal mode... blocking every writer for its full duration.

Fixed in a348c9dd. WAL keeps the single `Step(-1)` snapshot; DELETE uses bounded 128-page steps, retaining the existing two-minute retry limit. The mode selection has regression coverage.

> The new tests ... inherit the ambient environment.

Fixed in a348c9dd. Every backup test explicitly sets `XUI_DB_JOURNAL_MODE`; backup coverage now lives in `backup_test.go`.

> Add a short doc comment ... covering those three points.

Not adding a Go line comment: this repository's `CLAUDE.md` explicitly prohibits newly committed `//` comments. The explicit cleanup return and tests make the ownership boundary visible without violating that convention.

> a `Finish()` that reports an error can leave the finalizer armed.

Not addressing: `go-sqlite3@v1.14.48` clears the finalizer and nulls the handle before returning the prior-operation error from `Close`, so the described double-finish path does not exist in the pinned version.

PR #6138, opened 2026-07-29
fix(xray): synchronize lifecycle state

PathGao:
## Summary

Status, result, traffic, Telegram, and outbound-bridge requests could read a different Xray process or configuration than a concurrent restart/hot apply, yielding stale errors, ports, or online-client data. They now retain one lifecycle/configuration snapshot; cached failure output is accepted only from the current process.

Version probing now times out after five seconds, so a replacement binary that hangs on `-version` cannot keep a restart locked indefinitely. The Windows stop result is no longer cached as an error.

## Validation

- `make verify`
- `go test -race -shuffle=on -count=1 ./internal/xray ./internal/web/service ./internal/web/service/tgbot -run 'Test(ProcessConfigSnapshotsAreRaceSafe|RefreshVersionTimesOut|XrayLifecycle)'`

`make race` also exposed pre-existing races in `internal/mtproto` tests; those are outside this PR and will be handled separately.

### DIFF
--- a/internal/web/service/inbound_disable.go
+++ b/internal/web/service/inbound_disable.go
@@ -19,7 +19,7 @@ func (s *InboundService) disableInvalidInbounds(tx *gorm.DB) (bool, int64, error
 	now := time.Now().Unix() * 1000
 	needRestart := false
 
-	if p != nil {
+	if process := currentXrayProcess(); process != nil {
 		var tags []string
 		err := tx.Table("inbounds").
 			Select("inbounds.tag").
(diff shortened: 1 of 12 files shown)


PathGao:
Follow-up pushed in cb3518c9. It protects Process.config with the existing process mutex, snapshots Telegram reads once with a nil check, strengthens the timeout fixture so it must run for the configured deadline, covers result store/clear behavior, and makes the concurrent traffic assertion explicit. I also corrected the stale lifecycle wording. I am not retrying `-version` synchronously: RestartXray holds the lifecycle lock through Start, so a retry turns one bounded five-second stall into a longer restart outage. An Unknown version was already possible on execution failure; an asynchronous retry would need a separate lifecycle design and is better handled independently.

PR #6139, opened 2026-07-29
fix(sub): coalesce external subscription refreshes

PathGao:
## Summary

When a cached external subscription expired, simultaneous client requests each fetched the provider independently. Refreshes are now coalesced per URL; callers share the outcome, including the existing stale-on-error fallback.

The local cache is capped at 256 entries and removes the least recently fetched non-current entry when it fills.

## Validation

- `make verify`
- `go test -race -shuffle=on -count=1 ./internal/sub`

### DIFF
--- a/internal/sub/external_subscription.go
+++ b/internal/sub/external_subscription.go
@@ -16,8 +16,9 @@ import (
 // with a short timeout so a slow/dead provider can't stall a client's sub.
 
 const (
-	subscriptionCacheTTL = 5 * time.Minute
-	subscriptionMaxBytes = 2 << 20 // 2 MiB
+	subscriptionCacheTTL      = 5 * time.Minute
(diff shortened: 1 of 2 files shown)


PathGao:
Follow-up pushed in 021e9cd6. In-flight cleanup is deferred immediately after publication, so panic unwinding releases waiters and removes the entry. Waiters now receive the leader outcome from subscriptionFetch rather than re-reading the cache, which preserves stale-on-error even when another refresh evicts the stale entry. The capacity test derives from subscriptionCacheCapacity and asserts its exact retained size; the cache reset restores both maps; and the trim helper is named trimSubscriptionCacheLocked. A new concurrent stale-refresh test fills the cache, causes eviction during a failed refresh, and verifies every caller still receives the stale links. I left the fixed short delay as a non-failing scheduling aid; the request-count assertion remains a regression guard because the handler keeps the leader request open.

PR #6140, opened 2026-07-29
fix(mtproto): synchronize child process state

PathGao:
## Summary

The MTProto child-process watcher called `Cmd.Wait` while status and shutdown paths read mutable `exec.Cmd` fields. Lifecycle state is now snapshotted under a mutex and completion is determined by the process-owned done channel, preventing `Wait`/`ProcessState` races during reload and shutdown.

## Validation

- `make verify`
- `go test -race -shuffle=on -count=1 ./internal/mtproto`


PathGao:
Closing to consolidate the regression coverage and the lifecycle rationale into a clean replacement PR.

PR #6141, opened 2026-07-29
fix(mtproto): synchronize child-process lifecycle

PathGao:
## Summary

MTProto status and shutdown requests can run while the child watcher calls `Cmd.Wait`. `Wait` mutates `exec.Cmd` process state, so reading that state concurrently could race during a reload or shutdown.

The wrapper now snapshots its lifecycle fields under a mutex and uses its process-owned completion channel as the sole exit signal. Exit errors are protected independently of logging, so a slow log sink does not extend the lifecycle lock. The focused regression test reads status and result concurrently with a real child-process stop under `-race`.

This replaces the closed #6140 after consolidating the test coverage and lifecycle rationale. The previous arm64/armv6 release failures occurred during external packaging downloads after the Go binary had compiled; this PR does not change release packaging.

## Validation

- `make verify`
- `go test -race -shuffle=on -count=1 ./internal/mtproto`

### DIFF
--- a/internal/mtproto/manager_reload_test.go
+++ b/internal/mtproto/manager_reload_test.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 )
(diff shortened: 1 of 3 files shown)


PathGao:
Follow-up pushed in 75757ed4. The lifecycle regression coverage now lives in `process_race_test.go`. It registers process cleanup immediately after start and always stops reader goroutines through a deferred close/wait sequence, including failure paths. A controlled fake-child non-zero exit drives `wait` while concurrent readers call both `IsRunning` and `GetResult`, and the companion lifecycle-field test churns `cmd`, `done`, and `exitErr` under the production locking contract. `go test -race -shuffle=on -count=1 ./internal/mtproto` and `make verify` pass locally. I kept the current completion ordering: it preserves publication of the exit result before waiters observe completion; moving the signal ahead of log flushing is a separate behavioral tradeoff not required for the race fix.

PR #6142, opened 2026-07-29
fix(frontend): preserve request cancellation and schema failures

PathGao:
## Summary

Requests that combine a timeout with a caller-provided cancellation signal now honor either cancellation source. Query parameters append correctly when callers already supply a query string or URL fragment.

Successful envelopes with malformed client, settings, or Xray payloads now fail at their schema boundary instead of entering TanStack Query caches as asserted values. The failure includes missing payloads; Query consumers receive their normal error state rather than corrupted data.

## Validation

- `make verify`
- `npm test -- src/test/http-init.test.tsx src/test/zodValidate.test.ts`

### DIFF
--- a/frontend/src/api/http-init.ts
+++ b/frontend/src/api/http-init.ts
@@ -88,6 +88,21 @@ function encodeForm(data: unknown): string {
   return parts.join('&');
 }
 
+function appendQuery(url: string, query: string): string {
+  if (query === '') return url;
+  const hashIndex = url.indexOf('#');
+  const path = hashIndex === -1 ? url : url.slice(0, hashIndex);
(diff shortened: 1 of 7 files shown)


PathGao:
Follow-up pushed in 6194ed52. Schema failures are now strict only at the settings and paged-clients query boundaries, where TanStack Query presents an error state. The shared parser preserves successful null envelopes and non-strict callers retain their existing degraded behavior. Xray config schema mismatches again keep the raw JSON available for repair while emitting the diagnostic warning. I also added the missing timeout-side signal regression test and simplified the query separator expression. Targeted frontend tests, typecheck, and lint pass locally.

PathGao:
Closing this PR so the finalized scope and validation can be resubmitted cleanly.

PR #6143, opened 2026-07-29
fix(frontend): preserve cancellation and reject invalid query data

PathGao:
## Summary

A request that specifies both a timeout and an `AbortSignal` now cancels when either fires, including browsers without `AbortSignal.any`. Existing query strings and fragments no longer cause appended parameters to form malformed URLs.

Malformed paged-client payloads are rejected before entering the TanStack Query cache. Settings remain tolerant because the backend accepts values outside the frontend schema; operators can still open and repair those settings. Malformed Xray config is logged but remains available in the raw JSON editor for the same recovery path.

## Validation

- `make verify`

### DIFF
--- a/frontend/src/api/http-init.ts
+++ b/frontend/src/api/http-init.ts
@@ -88,6 +88,21 @@ function encodeForm(data: unknown): string {
   return parts.join('&');
 }
 
+function appendQuery(url: string, query: string): string {
+  if (query === '') return url;
+  const hashIndex = url.indexOf('#');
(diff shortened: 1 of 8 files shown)


PathGao:
Implemented the review fixes in c0d639d2. Settings payloads now stay tolerant so backend-accepted values remain repairable; paged clients retain strict query validation. The timeout test waits for the abort event, the AbortSignal.any fallback preserves both cancellation paths, and URL tests cover fragments and separators. `make verify` passes locally.

PR #6144, opened 2026-07-29
fix(frontend): preserve edited server drafts

PathGao:
## Problem

A successful settings/Xray save invalidates the query. If the user continues editing before that refetch returns, the old hydration path can overwrite the newer unsaved input. Xray also polls every second only to derive whether Save is enabled.

## Change

- Add a small server-draft hook shared by Settings hydration.
- Hydrate only on first load, when the draft is clean, after explicit discard, or when the refetch confirms the just-saved draft.
- Preserve a later edit while an earlier save refetches.
- Replace Xray's dirty-state polling with render-time comparison.

## Regression coverage

- Keep edited data through a refetch.
- Clear dirty state when the refetch confirms the saved draft.
- Hydrate clean drafts and explicit discards.

## Validation

- `make verify`

### DIFF
--- a/frontend/src/api/queries/useAllSettings.ts
+++ b/frontend/src/api/queries/useAllSettings.ts
@@ -28,7 +28,7 @@ export function useAllSettings() {
   });
 
   const server = useMemo(() => new AllSetting(query.data), [query.data]);
-  const { draft, setDraft, isDirty, discard } = useServerDraft(
+  const { draft, setDraft, isDirty } = useServerDraft(
     query.data === undefined ? undefined : server,
     (setting) => new AllSetting(setting),
     (left, right) => left.equals(right),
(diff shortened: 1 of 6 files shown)


PathGao:
Addressed the review in b888a1b3.

- Server-derived tag and subscription lists now refresh even when the editable Xray draft is dirty.
- Save-state comparison normalizes an empty outbound test URL consistently with the request payload.
- `useServerDraft` now tracks the latest server baseline in state and no longer mutates refs inside a state updater.
- Added regression coverage for the preserved-draft baseline, StrictMode hydration, settings query invalidation, dirty-editor subscription refresh, and empty URL save.
- Removed the stale hydration comment and unused Settings `discard` return value.

PathGao:
Superseded by #6147. The v2 branch includes the now-green #6146 CI baseline correction, so it receives a fresh CI and Bot review.

PR #6145, opened 2026-07-29
fix(frontend): keep DNS hosts synchronized

PathGao:
## Problem

DnsTab mirrors `dns.hosts` into local state and refreshes that mirror only when DNS is enabled or disabled. An external config update while DNS remains enabled leaves the editor on old hosts, so a later edit can overwrite the new server value.

## Change

- Derive host rows directly from `templateSettings.dns.hosts`.
- Keep edits on the existing immutable template update path.
- Remove the effect and local mirror.

## Regression coverage

- An external hosts refresh updates the rendered hosts row.

## Validation

- `make verify`

### DIFF
--- a/frontend/src/pages/xray/dns/DnsTab.tsx
+++ b/frontend/src/pages/xray/dns/DnsTab.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useMemo, useState } from 'react';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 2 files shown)


PathGao:
Addressed the review in 267ed6bd.

- Restored local host rows only for in-progress edits, while re-seeding them whenever an externally supplied hosts value differs from the value this component last serialized.
- Incomplete added or edited rows remain visible; external refreshes still replace stale rows.
- Added coverage for adding an empty row, clearing a domain mid-edit, and an external hosts refresh.

PathGao:
Superseded by #6148. The v2 branch includes the now-green #6146 CI baseline correction, so it receives a fresh CI and Bot review.

PR #6150, opened 2026-07-29
fix(frontend): preserve edited server drafts

PathGao:
## Summary

- Preserve Settings and Xray edits across background refetches.
- Establish the saved Settings baseline before invalidating its query, so redacted-secret rehydration cannot depend on refetch timing.
- Keep server-derived Xray projections current while an editable draft is dirty, remove dirty-state polling, and normalize an emptied test URL to its effective default.

## Validation

- `npm run test -- useServerDraft.test.tsx use-all-settings.test.tsx use-xray-setting.test.tsx`
- `npm run typecheck`
- `npm run lint`

### DIFF
--- a/frontend/src/api/queries/useAllSettings.ts
+++ b/frontend/src/api/queries/useAllSettings.ts
@@ -28,7 +28,7 @@ export function useAllSettings() {
   });
 
   const server = useMemo(() => new AllSetting(query.data), [query.data]);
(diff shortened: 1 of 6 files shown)


PathGao:
Closing this revised attempt because Claude Bot only reviews the PR creation event; its visible review is for an earlier head and its findings have since been addressed. Re-opening as a fresh, clean-base PR for a review of the current diff.

PR #6152, opened 2026-07-29
fix(install): preserve custom bin/ files (e.g. hand-added geoip) across updates

kuzzrus:
## Problem

Every reinstall/update (`install_x-ui`) stops the panel and then wipes
`/usr/local/x-ui/` wholesale before re-extracting the release tarball:

```bash
rm ${xui_folder}/ -rf
```

The release tarball only ships the assets 3x-ui itself bundles (the
`xray`/`mtg` binaries, the default `geoip.dat`/`geosite.dat` set). Anything
else an admin has placed in `bin/` — most commonly a hand-added custom
geoip/geosite file referenced from a routing rule via `ext:<file>:<code>` —
has no owner other than the admin, and is silently deleted by this wipe.

Real-world impact: after an update, Xray refuses to start at all, with
every inbound down, because a routing rule now points at a file that no
longer exists (`failed to open <file>: no such file or directory`). The
only way back is restoring the missing file from a separate backup — the
installer itself gives no warning that this will happen or has happened.

## Fix

`install_x-ui` now backs up the existing `bin/` directory before the wipe,
and after extracting the fresh release, restores only the files the new
release doesn't already provide. Bundled assets (`geoip.dat`, `xray-linux-*`,
etc.) always end up as the fresh release's copy — nothing about the update
path for standard files changes. Anything extra that was in `bin/` and
isn't part of the new release (custom geoip/geosite files, or anything else
hand-placed there) survives the update untouched, and the installer prints
which file(s) it restored so the admin can see it happened.

Also cleans up the temporary backup on the existing early-exit error paths
(failed extraction, missing `x-ui` binary after extraction) so a failed
update doesn't leave a stray temp directory behind.

## Testing

- `bash -n install.sh` — syntax check passes.
- `shellcheck install.sh` — no new warnings introduced by this change
  (one pre-existing-style `SC2086` info note, consistent with how the
  surrounding function already handles `${xui_folder}` elsewhere).
- Verified in isolation: standard files (`geoip.dat`, the `xray` binary)
  end up as the fresh release's copy after an update; a custom file that
  isn't part of the release survives an update untouched.

### DIFF
--- a/deploy/test/smoke-noninteractive.sh
+++ b/deploy/test/smoke-noninteractive.sh
@@ -87,6 +87,24 @@ docker run --rm \
             *) echo "FAIL: panel did not serve (status ${code:-none})"; tail -n 30 /tmp/xui.log; exit 1 ;;
         esac
 
+        echo "--- verifying a second install preserves custom bin/ files ---"
+        echo "custom-sentinel" > /usr/local/x-ui/bin/geoip_custom.dat
+        geoip_sum_before=$(sha256sum /usr/local/x-ui/bin/geoip.dat | cut -d" " -f1)
+
+        if [ -n "${XUI_SMOKE_VERSION:-}" ]; then
+            cat /root/install.sh | bash -s -- "$XUI_SMOKE_VERSION"
+        else
+            cat /root/install.sh | bash
+        fi
+
+        test -f /usr/local/x-ui/bin/geoip_custom.dat \
+            || { echo "FAIL: custom bin/ file did not survive a second install"; exit 1; }
+        [ "$(cat /usr/local/x-ui/bin/geoip_custom.dat)" = "custom-sentinel" ] \
+            || { echo "FAIL: custom bin/ file content changed across a second install"; exit 1; }
+        geoip_sum_after=$(sha256sum /usr/local/x-ui/bin/geoip.dat | cut -d" " -f1)
+        [ "$geoip_sum_after" = "$geoip_sum_before" ] \
+            || { echo "FAIL: bundled geoip.dat changed across a same-version reinstall"; exit 1; }
+
         echo "SMOKE_PASS: user=$XUI_USERNAME port=$XUI_PANEL_PORT path=$XUI_WEB_BASE_PATH"
     '
 
--- a/install.sh
+++ b/install.sh
@@ -1498,11 +1498,23 @@ install_x-ui() {
         # from a routing rule via "ext:<file>:<code>") and would otherwise be
         # silently deleted on every update, breaking Xray at next start with
         # "failed to open <file>: no such file or directory" for any routing
-        # rule that references it.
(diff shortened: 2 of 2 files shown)


kuzzrus:
Thanks — went through this point by point in `dc377d06`:

**install.sh:1503 (Medium, unchecked `mktemp`)** and **install.sh:1504 (Medium, silently-ignored `cp`)** — Fixed together. Replaced the `mktemp -d` + `cp -a` snapshot with a same-filesystem `mv` of `bin/` to a sibling path under `${xui_folder%/x-ui}`, checked with `||`. A rename is atomic (no truncated file if disk space runs out mid-copy) and needs no extra space in a separate, possibly small `$TMPDIR` — removes both failure modes at once rather than just adding checks on top of them. If the `mv` itself fails, it backs off cleanly (`custom_bin_backup=""`) and prints why instead of proceeding as if a backup exists.

**install.sh:1535 (Medium, `config.json`/`mtproto/*.toml` misclassified as custom)** — Confirmed and fixed. Verified both paths directly: `internal/xray/process.go:41` writes `bin/config.json`, `internal/mtproto/process.go` writes `bin/mtproto/mtg-<id>.toml` via `configDir()`, and `internal/mtproto/manager.go:606` creates that directory at `0750`. Added a `case` skip for `config.json` and `mtproto/*` in the restore loop so neither gets treated as admin-placed.

**install.sh:1535 (Low, restore runs before the arm rename)** — Fixed. Moved the restore block to after the arch-rename/chmod section, so `xray-linux-arm32`/`mtg-linux-arm` already exist under their final names by the time the restore loop runs and aren't needlessly copied back and misreported as custom.

**install.sh:1544 (Low, symlinks dropped)** — Fixed. `find` now matches `\( -type f -o -type l \)` instead of `-type f` only — `cp -a` was already preserving a symlink in the snapshot, the restore loop was just never picking it back up.

**install.sh:1502 (Low, no trap for an interrupted update)** — Fixed. Added `trap '...rm -rf "${custom_bin_backup}"...' EXIT INT TERM` right after the snapshot; this also let me drop the two manual `[[ -n ... ]] && rm -rf` cleanups at the extraction/binary-missing exit paths, since the trap now covers those exits too.

**install.sh:1502 (Low, unquoted `${xui_folder}`)** — Fixed, both new expansions quoted. Left the pre-existing snapshot-vs-restore path asymmetry alone per your own note that it's out of scope here.

**Suggestion (smoke test coverage)** — Added. `deploy/test/smoke-noninteractive.sh` now drops a sentinel file into `bin/` after the first install, reinstalls over the existing install, and asserts the sentinel survived and `geoip.dat`'s checksum is still the release's own copy.

Verified: `bash -n` and `shellcheck` clean on every changed line (both `install.sh` and the smoke test's embedded script, extracted and checked separately). Same fix set ported to this fork's own `main`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #6153, opened 2026-07-29
fix(frontend): preserve theme body classes

PathGao:
## Summary

Panel and Storybook theme application now preserve unrelated `<body>` classes instead of replacing the complete class attribute. Theme classes are installed before paint in both paths, and Storybook now mirrors the panel DOM for normal themes by removing `data-theme` rather than inventing `light` or `dark` values.

Regression coverage verifies that the panel and Storybook retain an unrelated body class while applying the selected theme, and that Storybook has no normal-theme `data-theme` attribute.

## Validation

- `npm run test -- storybook-theme.test.tsx`
- `npx vitest run --project storybook src/components/clients/ConfigBlock.stories.tsx`
- `npm run typecheck`
- `npm run lint`

### DIFF
--- a/frontend/.storybook/preview.tsx
+++ b/frontend/.storybook/preview.tsx
@@ -17,7 +17,7 @@ if (!i18next.isInitialized) {
   });
 }
 
-const withTheme: Decorator = (Story, context) => {
+export const withTheme: Decorator = (Story, context) => {
   const dark = context.globals.theme === 'dark';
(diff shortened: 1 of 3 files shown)


PathGao:
Closing this revised attempt because Claude Bot only reviews the PR creation event; the visible review is for an earlier head and its follow-up hardening has since been applied. Re-opening as a fresh PR for review of the current diff.

PR #6154, opened 2026-07-29
feat(xray): autocomplete geosite/geoip categories in the routing rule editor

kuzzrus:
## Problem

Writing a routing rule today means remembering exact `geosite:`/`geoip:`/
`ext:<file>:<code>` syntax by hand, with no way to discover what categories
actually exist in the `.dat` files sitting in the `bin` folder — including
custom ones added via the Geodata auto-update feature (`geosite_roscom.dat`,
`geoip_IR.dat`, etc.). Getting this wrong means a routing rule silently
doesn't match anything.

## Fix

The Domain/IP fields in the routing rule editor now suggest categories as
you type, built live from whatever `.dat` files are actually on disk, while
still accepting any free-typed value exactly as before:

Domain field suggesting geosite categories, including custom ext: files, while typing "youtu"

IP field suggesting a geoip category built from a custom ext: file while typing "gith"

**Backend:** `GET /panel/api/xray/getGeodataCategories` scans the `bin`
folder, parses matched `geosite*`/`geoip*.dat` files via xray-core's own
exported protobuf types, and formats each category as the exact rule syntax
xray-core's parser accepts — `geosite:`/`geoip:` for the default files,
`ext:<file>:<code>` for anything else (there's no shorthand for custom
files). Results are cached in memory keyed by each file's `(name, size,
modTime)`, so a request-time scan is cheap until a file actually changes.

**Frontend:** the Domain/IP inputs become `Select` "tags" fields fed by a
new `useGeodataCategories()` query hook, with an explicit substring filter
so `you` matches `geosite:youtube` (not just a prefix match). The
array↔CSV-string adapter lives entirely at the `FormField` transform
boundary, so the underlying form schema and saved rule shape are unchanged
— existing rules and the API contract for saving a rule are unaffected.

## Testing

- `go build`/`go vet` — clean for the touched backend packages.
- `go test ./internal/web/service/... ./internal/web/controller/...` — new
  geodata tests pass (`TestScanGeodataFiles`, `TestScanGeodataFilesMissingDir`,
  `TestParseGeodataFile`, `TestFormatGeodataSuggestion`,
  `TestGeodataFingerprintOf`, `TestGetGeodataCategories_SkipsMalformedFile`,
  `TestGetGeodataCategories_EndToEnd`, `TestGetGeodataCategoriesEndpoint`).
- `npm run typecheck`, `npm run lint`, `npm run test` (823 tests) and
  `npm run build` all pass.
- Manually exercised in a real panel (screenshots above) with both the
  default bundled `geosite.dat`/`geoip.dat` and several custom `ext:`-style
  files already on disk from the Geodata auto-update feature.


kuzzrus:
Thanks for the review — went through it point by point in `b0e65f69`:

**Finding 1 (High, full protobuf unmarshal)** — Fixed. `parseGeodataFile` now walks the raw wire format via `google.golang.org/protobuf/encoding/protowire` and extracts only each entry's `Code`, skipping every `Domain`/`CIDR` payload via `ConsumeFieldValue` without ever allocating it. Also added a 256 MiB read cap (`maxGeodataFileSize`) with `io.LimitReader`.

**Finding 2 (Medium, concurrency)** — Fixed. `GetGeodataCategories` now holds `geodataMu` across the full scan-and-maybe-parse instead of releasing it around the parse, so concurrent cache misses can't all redundantly re-parse. Also added `cloneGeodataCategories` so the cache never hands out its own backing slices — a caller mutating its result can't corrupt what other callers read next.

**Finding 3 (Medium, fires on every Routing-tab visit)** — Fixed. `useGeodataCategories` now takes an `enabled` param, wired to `RuleFormModal`'s own `open` state — the scan/parse only runs once the editor is actually opened.

**Finding 4 (Medium, test assertions)** — Fixed. Switched `Contains`-only assertions to exact `slices.Equal`, wrapped the table-driven tests in `t.Run`, and added `TestGetGeodataCategories_CacheHitSkipsReparse` (via a test-only atomic parse counter) plus `TestGetGeodataCategories_ReturnsIndependentSlices` for the cache-cloning fix above.

**Finding 5 (Medium, no frontend test for the tags adapter)** — Fixed. Added `frontend/src/test/rule-form-geodata-tags.test.tsx`: round-trip of an existing comma-separated value, adding one via Enter, the blur-commit case below, and the omitted-when-empty case.

**Finding 6 (Medium, unconfirmed blur-commit gap)** — Confirmed real, not just hypothetical: an RTL test typing a value and clicking Save without Enter/comma dropped it silently, with no `domain` key at all in the saved rule. Fixed with a small `TagsAutocomplete` wrapper around `Select mode="tags"` that also commits the pending search text on blur, covered by the new test.

**Finding 7 (Low, sourceIP inconsistency)** — Fixed. `sourceIP` now uses the same `TagsAutocomplete` + `ipOptions` as `ip`.

**Finding 8 (Low, case-sensitivity mismatch)** — Fixed. `formatGeodataSuggestion` now compares filenames with `strings.EqualFold`, matching `scanGeodataFiles`' own case-insensitive match — a file that IS the default one on a case-insensitive filesystem (e.g. Windows) gets the short form again. Added a test case for it.

**Finding 9 (Low, unguarded malformed response)** — Fixed. `fetchGeodataCategories` now guards `domain`/`ip` independently with `Array.isArray` instead of a single top-level `?? EMPTY_CATEGORIES`, since `parseMsg` returns the original unvalidated `obj` (not null) on a schema mismatch.

**Finding 10 (Low, codegen conventions)** — Partially addressed, and one correction. Added `GeodataCategories` to `StructAllow`, `example:` tags on both fields, and switched `endpoints.ts` to `responseSchema: 'GeodataCategories'` — that's what CLAUDE.md actually mandates ("response examples come from Go `example:` tags... never hand-write them"). The other half of the finding doesn't hold up, though: CLAUDE.md is explicit that "Zod schemas in `src/schemas/` are the source of truth" and are meant to be hand-written, with only the TS type inferred via `z.infer` — no existing hook in this codebase validates against the generated `zod.ts` schemas (e.g. `useNodesQuery` uses its own hand-written `NodeListSchema`, not the generated `NodeViewSchema`). Left `geodataFileKind`'s leak into `generated/{types,zod}.ts` alone: `internal/web/service`'s openapigen request has no `AliasAllow` at all, so every non-struct type in the package already leaks this way (`staticEgressResolver`, `transportBits` predate this PR) — scoping an `AliasAllow` for the whole package is a real cleanup but wider than this PR's footprint, and needs checking nothing already depends on those existing generated aliases first.

**Finding 11 (Low, unreferenced screenshots)** — Fixed. Dropped both PNGs from `media/`; the PR description's images now point at the specific commit SHA that still has them (`08588d96`), so they keep rendering without living in the tree going forward.

**Finding 12 (Low, xray-core proto type dependency)** — Resolved as a side effect of Finding 1: the protowire rewrite no longer imports `GeoSiteList`/`GeoIPList`/`proto.Unmarshal` at all. Only `geodata.DefaultGeoSiteDat`/`DefaultGeoIPDat` (plain string constants) remain, a much lower-risk surface.

Verified: Go `build`/`vet`/tests pass for the touched packages (a same-day, unrelated upstream commit needs cgo to build the rest of `internal/database` in my sandbox, so I cross-checked against a pre-that-commit base too); frontend `typecheck`/`lint`/full `test` suite/`build` all pass, including the 4 new tests above. Same fix set ported to this fork's own `main`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

STRENCH0:
Hi @Kuzz007 — heads-up that #6165 lands in the same place: it adds a category browser to the same three routing-rule fields, so whichever of these merges first, the other will need a rebase. Rather than have them compete, I think they fit together, and I'm happy to do the merging work on my side.

They solve different halves of the same problem. Yours is the fast path: you know roughly what you want, you type `you`, you get `geosite:youtube` without leaving the field. Mine is the discovery path: open the browser, see how many rules a category holds, what attributes it carries, page through its actual domains, tick several at once, and get told when a token already in the field does not resolve (missing file vs unknown category vs bad syntax). Autocomplete in the field plus a database button next to it would be strictly better than either alone.

Two findings from my side that apply to your PR as well, regardless of what happens with mine:

- **Matching files by name prefix misses real rule sets.** `scanGeodataFiles` keys off `geosite*` / `geoip*`, but the upstream v2fly release ships as `dlc.dat`, and hand-built files get arbitrary names. I ended up detecting the layout from the file contents instead: both list types share a wire shape whose inner field types disagree, so decoding one as the other yields no readable values, and counting readable entries tells them apart. Costs one extra pass only when the name doesn't hint at the right layout.
- **The asset folder isn't always `bin`.** `ensureXrayAssetLocation` (`internal/xray/api.go:434`) honours a preset `XRAY_LOCATION_ASSET` / `xray.location.asset` and only falls back to the bin folder. On an install pointing at a shared asset directory, both of our implementations list an empty folder — for mine that meant the validator reporting perfectly valid tokens as missing, which is exactly backwards.

@MHSanaei — could you say which one you'd like to take first? Either order works for me:

- **#6154 first** — I rebase #6165 on top, drop my `geodata/categories` endpoint in favour of feeding the browser from the autocomplete data where it overlaps, and keep only what the browser genuinely needs on top (paged category contents, token validation).
- **#6165 first** — the browser's endpoints already return everything the autocomplete needs (categories come back with counts and attributes), so @Kuzz007's `Select` could be pointed at `/panel/api/xray/geodata/categories` with the backend scan dropped, leaving his PR to the part that's actually his: the in-field UX.

I'd rather adapt than have you arbitrate between two half-overlapping diffs, so just name the order and I'll do the rebase.

PR #6161, opened 2026-07-30
feat(ui): let users pin the sidebar

PathGao:
## Summary

Desktop users can pin the sidebar open. The default remains the compact hover rail, and the choice is restored after reload.

## Why

The v3.6 sidebar redesign removed the prior persistent expand/collapse control. Users who need an always-visible navigation menu currently have no way to keep it open.

## Type of change

- [x] New feature

## Areas affected

## How was this tested?

- Added a component test covering pinning, mouse leave, and local-storage restoration.
- npm run test -- --run src/test/app-sidebar.test.tsx src/test/i18n-dead-keys.test.ts
- npm run typecheck
- npm run lint -- --quiet
- npm run build
- make verify could not complete because the local environment has no Go executable; frontend generation stopped at go run ./tools/openapigen.

## Screenshots / recordings

Not captured: the local environment has no runnable Go backend. The component test exercises the UI state transition.

## Breaking changes

None

## Checklist

- [ ] go build ./... and the test suite pass locally. Not run: Go is unavailable in this environment.
- [x] For frontend changes: npm run lint, npm run typecheck, and npm run build pass.
- [x] I updated the Wiki / README / API docs if user-facing behavior changed. No documentation change is needed.
- [x] My commits follow the project style.


PathGao:
> "The pinned sidebar does not reserve layout space"

Fixed in ac584cfc. A pinned desktop sidebar now reserves its full 220px width in the layout, so page content moves rather than remaining under an overlay. The hover-only sidebar still reserves the 72px rail and retains its shadow; the pinned sidebar no longer uses that shadow.

> "The transition it is meant to cover ... is never exercised"

The branch already added the hover-in → pin → mouse-leave path in 91c5d7b1. This update also covers unpinning back to the 72px rail and persistence of false.

> "Keep a stable aria-label with aria-pressed"

Fixed: the accessible name is now stable as "Pin sidebar" while aria-pressed provides the state. The changing title remains the action hint.

> "saveSidebarPinned(next) runs inside the setPinned updater"

Fixed: persistence now occurs in the click handler before setting state, rather than inside React\047s state updater.

> "In the collapsed rail, .sidebar-pin centers its icon"

That observation applies to the earlier revision. The pin is now only in the expanded header; the collapsed GitHub version link is centered with the navigation rail.

MHSanaei:
image


next time please use squash

PR #6164, opened 2026-07-30
feat(inbounds): improve multi-node online attribution

isultanov99:
## Summary

This PR improves online-client attribution on the inbounds page for multi-node setups.

Central panels already consume node-scoped online users through `onlinesByGuid`, but remote node sync did not carry the matching `activeInbounds` signal. As a result, the frontend could know that a client was online on a node, but not which inbounds on that node had recent traffic, so multi-inbound clients could appear online across too many inbounds.

The change makes node traffic snapshots include the existing `/panel/api/clients/activeInbounds` endpoint and keeps those active inbound tags in the same GUID-keyed in-memory tree model used for online users.

## What Changed

- Fetch remote `activeInbounds` during node traffic sync, best-effort and backward-compatible with older nodes.
- Store remote active-inbound subtrees in the running Xray process state.
- Merge local and remote active-inbound maps in `GetActiveInboundsByGuid`.
- Remap cloned/shared node GUIDs for active-inbound trees the same way online trees are remapped.
- Normalize node-local active inbound tags to the central panel's stored tags, including `n<nodeID>-` prefixed tags.
- Clear stale active-inbound state when a node's online subtree is cleared.

## Notes

This does not try to infer an exact `client -> inbound` pair because Xray exposes user traffic and inbound traffic as separate stats. The frontend can now gate online clients to active inbounds on the correct node, but when several inbounds on the same node are active in the same grace window, attribution remains best-effort.

Older nodes that do not expose `activeInbounds` keep the previous fallback behavior: the master still consumes their online users and simply skips active-inbound gating for that node.

## Validation

```bash
go test ./internal/xray ./internal/web/runtime ./internal/web/service
go test ./internal/web/job
git diff --check
```

### DIFF
--- a/internal/web/runtime/remote.go
+++ b/internal/web/runtime/remote.go
@@ -668,8 +668,13 @@ type TrafficSnapshot struct {
 	// OnlineEmails so the master can attribute deeply nested clients to the real
 	// node across a chain (#4983). Empty when the node is an old build without
 	// the per-GUID endpoint — OnlineEmails is the fallback then.
-	OnlineTree    map[string][]string
-	LastOnlineMap map[string]int64
+	OnlineTree map[string][]string
+	// ActiveInboundTree is the GUID-keyed subtree of inbound tags that carried
+	// traffic within the node's online grace window. Empty when the node is an
+	// old build without the endpoint; the master then falls back to email-only
+	// online attribution for that node.
+	ActiveInboundTree map[string][]string
+	LastOnlineMap     map[string]int64
 	// HostGroups carries the node's per-inbound host overrides (TLS/SNI/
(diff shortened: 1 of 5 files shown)


isultanov99:
Good points, thanks. Fixed in the latest push:

- moved the optional `activeInbounds` fetch after `lastOnline`, so it cannot delay the more important node-sync signal;
- filtered reported active-inbound GUID keys to the inbounds present in that node snapshot;
- added direct tests for empty tree cleanup and foreign GUID filtering;
- made the byte-idle trade-off explicit in the PR notes.

The stricter per-inbound behavior is intentional and matches the existing local active-inbound gating semantics: a globally online client can be omitted from a specific inbound if that inbound had no recent traffic.

PR #6166, opened 2026-07-31
fix(outbound): import Hysteria2 salamander properly from standard obfs params

xMasterX:
### Problem
The outbound share-link importers only reconstructed the salamander obfuscation from 3x-ui's private `fm=<json>` finalmask dump. They never parsed the **standard** Hysteria2 `obfs=salamander` + `obfs-password=<pw>` pair.

Since the subscription/link generator (`internal/sub`) already emits the standard `obfs`/`obfs-password` fields (and drops `fm=`, which broke mihomo & other clients), the obfuscation only round-trips 3x-ui→3x-ui. Importing **any** normal Hysteria2 link — every other client, and even our own generated links — silently dropped the salamander config. You end up with an outbound that negotiates plain QUIC while the server expects salamander, so it just fails to connect.

This affects both import paths:
- pasting a link into the outbound form (frontend `outbound-link-parser.ts`)
- subscription / JSON import on the backend (`internal/util/link`)

### Fix
Parse the standard `obfs=salamander` + `obfs-password` pair in both importers and fold it into `finalmask.udp` as `{ type: "salamander", settings: { password } }` — the same shape the UI and generator already use. Also accept the `obfs_password` / `obfsPassword` aliases.

If a link still carries a full `fm=` finalmask with a salamander mask, that wins and the obfs pair is a no-op, so existing 3x-ui→3x-ui links are unchanged.

### Tests
- Go: `internal/util/link/outbound_test.go` — standard obfs imported, no-obfs untouched, obfs-without-password ignored, and `fm=` wins over `obfs`.
- Frontend: `outbound-link-parser.test.ts` — obfs-only reconstruction, no-obfs, and obfs-without-password. `npm run test` and `go test ./internal/util/link/` both green.

No schema/DB/route/i18n changes; `make gen` output unaffected.

## Why

Was trying to fill the issue but found them missing completely. Why issues section are removed from repo?

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [X] Frontend (UI / panel pages)
- [X] Backend (API endpoints, login, settings)
- [X] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot

## How was this tested?

1 - Import sub with hysteria into Outbounds screen
Click test, see (failed)

2 - Import hysteria link manually via JSON screen on add outbound
See no salamander was added, if you continue it will fail to connect too
If you add salamander and enter password manually after link import - everything works correctly 

## Checklist

- [] I tested the change locally and confirmed the described behavior.
- [X] I added or updated tests for the new behavior (when applicable).
- [X] `go build ./...` and the test suite pass locally.
- [X] For frontend changes: `npm run lint`, `npm run typecheck`, and `npm run build` pass.
- [X] I updated the Wiki / README / API docs if user-facing behavior changed.
- [X] My commits follow the project's existing message style.
- [X] I have no unrelated changes mixed into this PR.


xMasterX:
Automatic review issues was fixed

Ready for re-review

PR #6168, opened 2026-08-01
Add remote routing URL support

yelloduxx:
## What

- Treat exactly one absolute HTTPS value in the existing Happ and Clash/Mihomo routing fields as a remote routing source while preserving every other value as legacy inline content.
- Refresh remote sources on panel startup and every five minutes; subscription requests only read validated memory/persisted cache and never wait for network or SQLite loading.
- Keep a 10-minute last-known-good cache with a 30-second failure retry delay, request coalescing, ETag/Last-Modified support, and persisted restart recovery.
- Convert an HTTPS Incy routing source to its native [`incy://autorouting/onadd/<URL>`](https://docs.incy.cc/en/autorouting/) deeplink, so Incy performs its own periodic refresh.
- Merge only `proxy-groups`, `rule-providers`, and `rules` from remote Clash/Mihomo documents; preserve panel-generated proxies and reject dangling group/provider/rule references.
- Validate remote URL settings at save time and update the existing settings guidance in all shipped locales.

## Why

The routing fields currently accept only materialized rules/deeplinks. That makes permanent upstream routing URLs unusable without a separate updater or a deployment-specific module.

## Compatibility and security

- Existing inline routing values retain their previous behavior, including independent `Routing-Enable` and `Routing` headers.
- Cold-cache requests fail closed by omitting an unavailable remote payload; stale-cache requests return the last validated value immediately while refresh runs in the background.
- HTTPS fetches use the existing SSRF-guarded dialer, constrained redirects, TLS 1.2+, a six-second background timeout, response-size limits, and no URL credentials.
- Happ responses are limited to 16 KiB before normalization and 8 KiB for the final `Routing` header value.
- Remote Clash content cannot replace generated proxies, import `proxy-providers`, or change listeners, LAN exposure, DNS, TUN, sniffing, controllers, or authentication.
- No deployment-specific URLs, server addresses, subscription identifiers, credentials, or private data are included.

## Checks

- `go test ./...`
- `go test -race ./internal/sub`
- `npm run lint`
- `npm run typecheck`
- `npm run build`
- `npm test -- --project unit --project components` (986 tests)
- `git diff --check`
- Current RoscomVPN Mihomo routing contract: 9 groups, 26 rule providers, and 31 rules accepted while standalone-client keys and the manual proxy-provider placeholder are dropped.
- Deployment smoke test across restart: Happ, Mihomo, Incy, and JSON subscriptions returned successfully; the generated Mihomo YAML passed official Mihomo 1.19.29 validation; persisted cache timestamps and content remained unchanged inside TTL.


yelloduxx:
Updated the PR branch onto the current upstream `main` (`216d18b3`) and resolved the background-job registration conflict while preserving both upstream's orphan reaper and this PR's remote-routing refresher.

The automated review findings are addressed in the current implementation:
- remote refreshes run asynchronously at startup and every 5 minutes; subscription requests only perform non-blocking cache reads;
- legacy `Routing` and `Routing-Enable` header behavior remains independent;
- Happ response/body limits are header-safe;
- remote Clash input is restricted to routing-scoped sections and validated for dangling references;
- parsed Clash YAML is cached;
- invalid remote URL settings are rejected at save time;
- all shipped locale descriptions and security/size/SSRF tests are included.

Validation after rebasing:
- `go test ./...`
- `go test -race ./internal/sub`
- `npm run lint`
- `npm run typecheck`
- `npm run build`
- `npm test -- --project unit --project components` (986 tests)
- `git diff --check`

The PR is now mergeable and remains a draft for maintainer review.

Issue #6116, opened 2026-07-27
[Bug]: pickClientsSummary's snapshot-coverage guard is a count check, not identity — an orphan+gap pair cancels out and slips through


mrnickson-hue:
### Describe the bug

Follow-up from #6110's review — a bot-reviewer finding that was agreed as real but deliberately deferred there as out of scope, tracked here per that PR's own discussion instead of being silently forgotten.

`pickClientsSummary` (`frontend/src/hooks/useClients.ts:120`, introduced in #6110) guards the live client-summary recompute with a coverage check at line 128:

```ts
if (serverSummary.total > allClientStats.length) return serverSummary;
```

This is a **count comparison, not an identity/coverage check**. `serverSummary.total` counts rows in the `clients` table; `allClientStats` is the `client_traffics` WebSocket snapshot, which can contain orphaned rows (traffic rows with no matching `clients` row — the exact class #6110 is about) and can be missing rows for clients that exist in `clients` but have no `client_traffics` row yet.

If there's simultaneously one orphaned row (inflating `allClientStats.length` by one) and one client missing its row (deflating it by one), the two cancel out numerically. The guard sees `serverSummary.total === allClientStats.length`, concludes coverage is complete, and proceeds with the live recompute — which still drops the real, uncovered client from every summary bucket while additionally bucketing the orphan under some other category. The check that exists specifically to catch an incomplete snapshot doesn't catch this shape of incompleteness.

### How to reproduce the problem

1. Get the panel into a state with one orphaned `client_traffics` row (no matching `clients` row) and, simultaneously, one `clients` row with no matching `client_traffics` row. (`clients`/`client_traffics` can drift apart via the seeder-skip and migration paths discussed in #6110's review.)
2. Trigger a `client_stats` WebSocket broadcast (any `XrayTrafficJob` tick with `websocket.HasClients()` true).
3. Compare the live-recomputed summary buckets against the true state: the real client that's missing its traffic row won't appear in any bucket (Active/Ended/Disabled/etc.), while the orphaned row will be bucketed as if it were a real client — with no warning surfaced, because the guard's count check happened to balance.

### Expected behavior

An incomplete or inconsistent snapshot should be detected regardless of whether the imbalance happens to net out to zero, and the guard should fall back to `serverSummary` in that case too — not just when the counts visibly disagree.

### Actual behavior

[...]

PR #6169, opened 2026-08-01
fix(clients): stop recomputing the summary badges from the client_stats snapshot

mrnickson-hue:
## Summary

Drops the client-side recompute in `pickClientsSummary` (`frontend/src/hooks/useClients.ts`) that could display a corrupted clients-page summary when the `client_stats` WS snapshot diverged from `clients` in a shape the coverage guard couldn't detect. The summary badges now always read the server's `q.summary()` result directly.

## Why

Refs #6116.

`pickClientsSummary`'s guard is a count comparison (`serverSummary.total > allClientStats.length`), not an identity/coverage check. `allClientStats` is the raw `client_traffics` table (`GetAllClientTraffics()`, no join back to `clients`), so it can carry orphaned rows (traffic with no matching client) and can be missing rows for clients that don't have one yet. A shortfall-only guard misses two real cases: an orphan and a gap netting out to the same count, and — per #6116's bot review — an orphan surplus alone, which is the more reachable one since it needs drift in only one direction.

Since #6116 was opened, `client_paging.go`'s `q.summary()` (landed in `f52c3c4`, "make the clients page scale to large panels") already computes the same bucket counts as a bounded SQL aggregate with `clients` as the driving table (`Table("clients AS c").Joins("LEFT JOIN client_traffics ...")`). That query can't confuse an orphaned `client_traffics` row for a real client regardless of how or when the row got there — it's driven by `clients`, not `client_traffics`. It's already shipped on every `/panel/api/clients/list/paged` response, and `listQuery` already polls that endpoint every 5000ms (`refetchInterval: 5000`), which happens to match `cadenceXrayTraffic = "@every 5s"` exactly.

That means the "cheaper aggregate query" #6116's discussion was waiting on to justify a server-side summary already exists and already runs on a matching cadence — just over REST instead of the `client_stats` WS payload. The client-side recompute this PR removes wasn't buying any extra freshness over that poll, only a window to get the number wrong. Dropping it needs no new backend work and, if anything, removes work from the hot path (`computeClientsSummary` used to re-run on every `client_stats` push).

Multi-node panels are, if anything, more exposed to the underlying divergence than a single-node install: `NodeTrafficSyncJob` is a second, independent source of `client_stats` broadcasts on its own `@every 5s` timer (not phase-locked to the local traffic job's), reading the same unjoined `GetAllClientTraffics()`, and `SetRemoteTraffic` merges up to 8 nodes' snapshots into the shared `clients`/`client_traffics` tables concurrently. `q.summary()`'s correctness doesn't depend on that timing at all, since it always reflects whatever's actually in the database at request time.

Mutation responsiveness (enable/disable/delete/etc.) is unaffected — those already call `invalidateAll()`, which forces an immediate `listQuery` refetch independent of the 5s poll.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [x] Multi-node (sub-nodes)
- [ ] Telegram bot

## How was this tested?

- `frontend/src/test/clients-summary.test.tsx`: two new tests reproduce #6116's exact shapes against the live hook (`renderHook` + `applyClientStatsEvent`) — an orphan row with no compensating gap, and an orphan+gap pair that nets to the server's total — and assert `summary` stays pinned to the server's value in both, where the old guard let at least one of these through.
- `allClientStats`, `computeClientsSummary`, `pickClientsSummary` and `sameSummaryInputs` are removed as dead code (their only caller was the recompute this PR drops); `sameSpeedMap`'s tests are kept as-is since that function is unrelated and untouched.
- `make verify` (gen-check, golangci-lint, ESLint, `tsc --noEmit`, `go test -shuffle=on -count=1 ./...`, frontend `npm run test`, `go build ./...` + frontend build, `build-storybook`) — all green except two pre-existing, unrelated `client-form-flow.test.tsx` / `inbound-form-modal.test.tsx` timeouts under this sandbox's load, which I confirmed also fail on an unmodified `main` checkout (`git stash` + rerun) before restoring this change, so they're not a regression from this diff.

## Screenshots / recordings

N/A — no visual change, the displayed numbers are unaffected in the common case; this only changes what happens during the specific divergence window #6116 describes.

## Breaking changes

None.

## Checklist


mrnickson-hue:
Both addressed in 81f7320.

**`useInbounds.ts:264`** — dropped the comment rather than repoint it at `depletedExpr()`/`summaryDeactiveExpr()`. Repointing would just trade one soon-to-be-stale name reference for another (those are unexported Go identifiers a future refactor could rename without anything on the TS side noticing), and the repo's no-`//`-comment convention already argues for dropping it outright — the priority itself is still expressed by the code's own order (`if (expired || exhausted) { depleted... } if (!client.enable) { deactive... }`).

**`useClients.ts` background staleness** — agreed this is real, not just theoretical. Confirmed the mechanism: `applyClientStatsEvent`'s per-row patch is wired straight to the WS bridge with no visibility check, so it keeps updating `queryClient`'s cached page rows while the tab is hidden; `listQuery`'s own `refetchInterval: 5000` pauses when hidden (as the existing comment on that line says), and `staleTime: Infinity` was blocking `refetchOnWindowFocus`'s default staleness check from ever firing, so the summary badges had no path back to correctness on refocus short of the next 5s tick.

Went with `refetchOnWindowFocus: 'always'` over widening `staleTime` for two reasons: it's the minimal change — one extra fetch exactly on the transition that exposes the gap, nothing else about the query's refetch behavior moves — and `staleTime: Infinity` is shared by every other `useQuery` in this hook (`inboundOptionsQuery`, `defaultsQuery`, `onlinesQuery`), so touching it here risked changing behavior I hadn't audited for the others.

Worth noting for the record: this gap predates this PR in substance, just not in visibility. Before this PR, `pickClientsSummary`'s live recompute happened to keep drawing from the same WS-fed, visibility-ungated `allClientStats`, so it papered over the same underlying REST-poll staleness — coincidentally, not by design (nothing in that code path was reasoning about tab visibility). Removing that recompute for #6116 is what surfaced it directly.

PR #6174, opened 2026-08-05
fix(sub): use a fullwidth percent in USAGE_PERCENTAGE

n0ctal:
## Summary
`{{USAGE_PERCENTAGE}}` renders an ASCII `%`, which becomes `%25` once the remark is placed in the share-link fragment. Happ discards the whole remark on that and shows the server hostname instead. Emit U+FF05 FULLWIDTH PERCENT SIGN.

## Why
A remark ends up in the URI fragment (`…#<remark>`), so `%` is percent-encoded to `%25`. Happ treats a fragment containing `%25` as malformed and drops the remark entirely, falling back to the hostname. The user sees their configured template replaced by the server address in the client's server list — both a loss of the template and an unnecessary disclosure of the host.

This reproduced across every subscription entry that used the token; other tokens in the same template render fine, which makes the cause easy to misattribute to the template rather than the percent sign.

## Scope
- `internal/sub/remark_vars.go` — `usagePercentage` emits `％` (U+FF05) instead of `%`.
- `internal/sub/remark_vars_test.go` — existing expectations updated; new `TestUsagePercentageSurvivesFragmentEncoding` asserts the encoded link contains no `%25` and that the fragment round-trips through `url.Parse`.
- `frontend/src/lib/remark/remarkVariables.ts` — the catalog sample matches what the panel now emits.

## Validation
- `go test ./internal/sub/ -count=1` green.
- `go vet ./internal/sub/` clean, `gofmt -l` empty.
- Verified the new test fails without the fix: it reports `#node%2050.0%25`.
- Behaviour confirmed against Happ and INCY on a live subscription: with `%` the remark is dropped and the hostname is shown; with `％` the remark renders.

## Risk
Low, cosmetic in effect: the rendered percentage looks the same to a reader but is a different code point, so anything matching the literal ASCII `%` in a remark would need updating. No API, schema or link-structure change.


n0ctal:

n0ctal:

PR #6175, opened 2026-08-05
fix(cli): stop -getApiToken accumulating admin tokens

n0ctal:
## Summary
`x-ui setting -getApiToken` minted a new `cli-fallback-<unix>` token on every invocation when tokens already existed. Rotate a single `cli-fallback` token instead.

## Why
The flag reads like a getter, and the surrounding output explains that existing tokens cannot be retrieved because only hashes are stored. It then creates one anyway — a fresh row, enabled, with no expiry, printed once.

An operator who runs the command a few times while debugging leaves several credentials behind. They all carry the same authority, none can be told apart later, and nothing indicates they were created rather than read. `ApiToken` has no expiry column, so they persist until someone notices and deletes them by hand.

I hit this while working on a downstream fork: four invocations left four enabled tokens on a production panel, and they were only spotted during an unrelated audit.

## Scope
- `internal/web/service/panel/api_token.go` — new `RecreateByName`, which deletes any row with that name before calling `Create`, so exactly one exists.
- `main.go` — the fallback uses the fixed name `cli-fallback` via `RecreateByName`; the message states the previous token is now invalid. Drops the now-unused `time` import.
- `internal/web/service/panel/api_token_test.go` — `TestRecreateByNameKeepsOneToken` asserts a second call rotates the plaintext and leaves one row.

## Validation
- `go test ./internal/web/service/panel/ -count=1` green; `go build ./...` and `go vet ./...` clean.
- Verified the test fails without the fix: the second call errors with `a token with that name already exists`.

## Risk
Low. The fallback keeps working; the only behaviour change is that a previously issued CLI fallback token stops working when a new one is printed, which the message now says. Tokens created through the UI are untouched.


n0ctal:

n0ctal:

PR #6177, opened 2026-08-05
feat(sub): warn when salamander settings cannot reach the client

n0ctal:
## Summary
A hysteria2 URI can express salamander obfuscation only as `obfs=salamander` + `obfs-password`. When the inbound's `finalmask` carries more than a password, the emitted link silently describes a server no standard client can reach. Log the keys that cannot be carried.

## Why
The comment above the current code is right that emitting the non-standard `fm=<json>` dump breaks mihomo and other clients, so only the standard fields are written. The gap is what happens when the standard fields are not sufficient to describe the server.

`packetSize` is the case I hit. With it set, a client built from the generated link applies plain salamander, the server expects the padded variant, and the handshake fails — no error on the client, nothing in the panel, and traffic accounting stays at zero because no session is ever established. It reads as "hysteria2 is broken" rather than "this inbound cannot be expressed as a link", and it survived weeks of looking in the wrong place before a controlled A/B on a live server isolated it: same client, same credentials, same port — connects when the server has only `password`, fails when `packetSize` is present.

This does not change the link, so nothing that works today stops working. It makes the reason discoverable.

## Scope
- `internal/sub/service.go` — `extraSalamanderKeys` lists salamander settings beyond `password`; the hysteria2 link builder logs a warning naming the inbound and those keys.
- `internal/sub/salamander_uri_test.go` — expressible settings report nothing, `packetSize` is reported.

## Validation
- `go test ./internal/sub/ -count=1` green; `go build ./...`, `go vet` clean, `gofmt` clean.
- Behaviour confirmed on a live deployment: with `packetSize` present, Happ and INCY both fail on every hysteria2 entry and the inbound records no traffic; removing it makes the same links work immediately.

## Risk
None to the data plane. Warning-only, emitted at link build time, no change to the URI or to any stored setting.

## Relation to #6166

To draw the boundary explicitly, since it is easy to read these as the same area: [#6166](https://github.com/MHSanaei/3x-ui/pull/6166) fixes the **import** direction — parsing standard `obfs`/`obfs-password` parameters into an outbound. This is the **export** direction: warning when an inbound's salamander settings contain fields such as `packetSize` that the standard share URI cannot express at all.

The two do not overlap in files or in substance, and #6166 cannot express `packetSize` in a URI either — that is precisely the case this warns about rather than fixes.


n0ctal:

n0ctal:

PR #6194, opened 2026-08-10
fix(nodes): apply a rotated master mTLS certificate without restarting the panel

n0ctal:
## Summary

Make the panel's outbound mTLS client transports observe a credential generation, so a rotated master certificate takes effect on the next request instead of after a panel restart. Adds `POST /panel/api/nodes/mtls/reloadClient` to trigger it.

## Why

The panel caches an mTLS transport per node. When the master client certificate changes — first issuance, or an operator rotating the stored credential — cached transports keep their old TLS config and keep reusing pooled keep-alive connections authenticated with the previous leaf. The new certificate only takes effect once the process restarts.

That forces a panel restart into any credential rotation. On an installation where the panel supervises Xray, restarting the panel interrupts the data plane for something that is purely a control-plane change.

## Scope

- A generation counter; each cached transport compares it before its next request, rebuilds its TLS transport and closes the old idle pool. Requests already in flight are not interrupted; no request that starts after invalidation reuses a connection authenticated with the previous leaf.
- `POST /panel/api/nodes/mtls/reloadClient` verifies the current credential pair is readable, then invalidates. This covers rotation performed directly against settings, not only first issuance.
- The endpoint is registered in the API docs inventory. Its absence was caught by a full `TestRouteRegistryContract` run and fixed.

## Validation

- `go test ./...` green; `go test -race ./internal/web/runtime` green; targeted controller/service/runtime tests green.
- Wire test: after the leaf changes, the next request presents the new certificate and the old keep-alive connection is not reused.

Not proven: the endpoint has not been exercised through a fully started panel against a real remote node. That remains an integration check rather than a unit one.

## Risk

Low. With no rotation the generation never advances and behaviour is unchanged. The endpoint only invalidates a cache; it does not mint, replace or delete any credential.

### DIFF
--- a/docs/content/docs/en/reference/api/nodes.mdx
+++ b/docs/content/docs/en/reference/api/nodes.mdx
@@ -1,51 +1,40 @@
 ---
 title: Nodes
-description: >-
-  Manage remote 3x-ui panels acting as nodes for a central panel. All endpoints
-  under /panel/api/nodes.
+description: Manage remote 3x-ui panels acting as nodes for a central panel. All
+  endpoints under /panel/api/nodes.
 full: true
 _openapi:
   preload:
     - ./public/openapi.json
   toc:
     - depth: 2
-      title: >-
-        List every configured node with its connection details, health, and last
+      title: List every configured node with its connection details, health, and last
         heartbeat patch.
-      url: >-
-        #list-every-configured-node-with-its-connection-details-health-and-last-heartbeat-patch
+      url: '#list-every-configured-node-with-its-connection-details-health-and-last-heartbeat-patch'
     - depth: 2
-      title: >-
-        This panel's node-auth CA certificate (public, PEM) to paste into a
+      title: This panel's node-auth CA certificate (public, PEM) to paste into a
         node's mTLS trust setting. Lazily mints the CA and the master client
         cert on first call. Pair with setting tlsVerifyMode=mtls on the node.
-      url: >-
-        #this-panels-node-auth-ca-certificate-public-pem-to-paste-into-a-nodes-mtls-trust-setting-lazily-mints-the-ca-and-the-master-client-cert-on-first-call-pair-with-setting-tlsverifymodemtls-on-the-node
+      url: '#this-panels-node-auth-ca-certificate-public-pem-to-paste-into-a-nodes-mtls-trust-setting-lazily-mints-the-ca-and-the-master-client-cert-on-first-call-pair-with-setting-tlsverifymodemtls-on-the-node'
     - depth: 2
-      title: >-
-        Set the CA certificate this panel trusts for incoming node-API client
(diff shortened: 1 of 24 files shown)


n0ctal:
Unrelated to this change, but you may notice a red `frontend` job in this PR's history and I would rather explain it than leave it looking like my code.

`storybook (chromium)` intermittently fails on `src/components/clients/ConfigBlock.stories.tsx > Collapsed`:

```
TypeError: textRange(...).getClientRects is not a function
    at TextTile.coordsIn (@codemirror/view)
    at DocView.coordsAt → EditorView.coordsAtPos
    at RectangleMarker.forRange → LayerView.measure → EditorView.measure
```

`Test Files 1 failed | 111 passed (112)`.

**Frequency:** 5 failures in 22 `frontend` job runs across my open PRs (~23%), always the same story. Runs: [31233485321](https://github.com/MHSanaei/3x-ui/actions/runs/31233485321), [31234367807](https://github.com/MHSanaei/3x-ui/actions/runs/31234367807), [31233553275](https://github.com/MHSanaei/3x-ui/actions/runs/31233553275), [31379586135](https://github.com/MHSanaei/3x-ui/actions/runs/31379586135), [31379616957](https://github.com/MHSanaei/3x-ui/actions/runs/31379616957). None of those branches touch `frontend/src`; two do not touch the frontend at all.

**Why it looks like interference rather than a ConfigBlock bug:** `ConfigBlock.tsx` does not use CodeMirror — only antd, `ClipboardManager`/`FileManager` and `QrPanel`. The single CodeMirror consumer under `frontend/src` is `components/form/JsonEditor.tsx`, and it does call `view.destroy()` in cleanup. The stack is a scheduled `EditorView.measure`, so a plausible reading is that a measurement scheduled by an editor elsewhere runs after its page is gone and the error lands on whichever test is executing. That would explain both the intermittency and why the failing story has nothing to do with CodeMirror.

**I could not confirm that.** On `main` at `ece16559`, macOS, Node 24, 16 cores, after `npm ci`: `npx vitest run` (112 files, as CI) twice, `--project=storybook` five times, and `--project=storybook --maxWorkers=2` three times — ten runs, zero failures. At a 23% rate that has roughly a 7% chance, so the local rate really is lower rather than lucky. The differences I cannot reproduce are runner core count, the headless shell build and the container.

So this is a measured symptom with traces, not a diagnosis, and deliberately not a PR — I did not want to send a speculative fix for something I was unable to make fail.

One possibly relevant detail I did not bisect: `frontend/package-lock.json` was last bumped in `38838827` (2026-08-04), and contributor PRs whose `frontend` job ran before that date are green while every failure I can see postdates it.

I would have filed this as an issue, but `blank_issues_enabled: false` and none of the three forms fit a CI flake report. Happy to move it wherever you prefer.

PR #6215, opened 2026-08-14
feat: add quota multiplier per client

elythepreciouscat:
## Summary
Adds a configurable `quotaMultiplier` per client (default `1.0`) that scales raw Xray traffic byte counters before they are stored, letting admins make bandwidth count faster or slower toward a client's quota.

## Why
Some clients need bandwidth accounting that isn't 1:1 with actual bytes transferred — e.g. promotional accounts, weighted plans, or nodes with different cost profiles. This exposes a per-client multiplier applied at the point traffic is recorded, instead of requiring workarounds in quota limits.

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected
- [ ] Xray config generation
- [x] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot

## How was this tested?

**Go (backend)**
- Added `TestAddClientTraffic_QuotaMultiplier` in `internal/web/service/inbound_client_traffic_test.go` — creates an in-memory SQLite DB with a client that has `quota_multiplier = 1.5`, feeds it a traffic delta of 100 Up / 200 Down, and asserts the stored values are 150 Up / 300 Down.
- Ran the full Go test suite (`make test-go`) with `-shuffle=on -count=1` (no cache) — all packages passed.

**Frontend (TypeScript / React)**
- Ran the full Vitest suite (`make test-fe`) — 112 test files / 1069 tests passed.
- Ran `npm run typecheck` (`tsc --noEmit`) and `npm run lint` — 0 errors.

**Manual / end-to-end**
- Built the full panel (`make build-fe && go build ./...`) and ran it locally.
- Logged into the panel, opened the client form on an inbound, confirmed the Quota Multiplier field appears in the Basic tab with the correct default (`1`), accepts decimal values (step `0.1`, range `0.1`–`10`), and is saved and reflected in the client info view.
- Confirmed the multiplier tag is hidden from the info row when the value is `1`, and highlighted when it differs.
- Opened the Bulk Add modal and confirmed `quotaMultiplier` is included in the payload.

## Database migration
`migrateQuotaMultiplierColumn()` is called at startup and back-fills `quota_multiplier = 1` for all pre-existing client rows on both SQLite and PostgreSQL. No manual migration step is required.

## Screenshots / recordings
_None — no visual UI changes beyond a new form field; can add if requested._

## Breaking changes
None. The new column has a default of `1`, so all existing clients behave identically to before — their traffic is recorded at a 1:1 ratio. The field is `omitempty` in JSON, so existing API consumers that don't send it receive the default.

## Checklist


MHSanaei:
sorry i don't want to add this option

PR #6222, opened 2026-08-16
fix(frontend): keep the MSW worker in step with the lockfile

n0ctal:
## Summary

`frontend/public/mockServiceWorker.js` is tracked at msw 2.14.7 while `package-lock.json` installs 2.15.0. msw regenerates that worker on postinstall, so a clean `npm ci` rewrites a tracked file and leaves the working tree dirty.

## Why

Every contributor hits it: after the first install, `git status` shows a 30-line diff nobody asked for. It either gets committed as noise or discarded, and the same diff comes back on the next install. `make verify` does not notice, because nothing checks tree cleanliness after install.

The worker is not cosmetic — it is the service worker the browser tests register, so a stale copy means the mocked runtime does not match the version the tests were written against.

## Scope

- regenerate `mockServiceWorker.js` for the locked msw version;
- add `make msw-worker-check`, comparing the tracked worker to `node_modules/msw/lib/mockServiceWorker.js`;
- wire it into `make verify` and add the equivalent step to the frontend CI job.

The check is a `cmp` against the installed package, so it stays correct across future msw bumps: whoever raises the lockfile regenerates the worker in the same commit or CI fails.

## Validation

On a clean checkout of `main`, the tracked worker and the installed one differ (2.14.7 vs 2.15.0, 30 lines). With this change they are byte-identical, `git diff --exit-code` after `npm ci` is clean, and `make msw-worker-check` exits 0.

## Risk

Low, and confined to test infrastructure — the worker is only registered by the browser test setup. If the two ever diverge again, CI now fails with a direct message instead of silently shipping a mismatched mock runtime.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -162,6 +162,9 @@ jobs:
       - name: Install
         run: npm ci
         working-directory: frontend
+      - name: Verify generated MSW worker is current
+        run: git diff --exit-code -- public/mockServiceWorker.js package-lock.json
+        working-directory: frontend
       - name: Lint
         run: npm run lint
         working-directory: frontend
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,10 @@ lint: lint-go lint-fe ## All linters
 typecheck: ## tsc --noEmit
 	cd $(FRONTEND) && npm run typecheck
 
+.PHONY: msw-worker-check
+msw-worker-check: ## Verify the tracked worker matches the installed MSW runtime
+	cmp $(FRONTEND)/public/mockServiceWorker.js $(FRONTEND)/node_modules/msw/lib/mockServiceWorker.js
+
 .PHONY: test-go
 test-go: dist-stub ## Go tests (shuffle, no cache)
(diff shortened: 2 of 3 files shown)


n0ctal:
On the concern that the CI step may be unable to fail: I tested it rather than reasoning about it, and msw's postinstall does run — npm does not honour the `allowScripts` map (it is a lavamoat key, not an npm one).

Reproduction on this branch: replace the tracked worker with `main`'s copy, wipe `node_modules`, run `npm ci`:

```
before install: const PACKAGE_VERSION = '2.14.7'
after  install: const PACKAGE_VERSION = '2.15.0'
```

The file is rewritten during install, so `git diff --exit-code` right after `npm ci` does fail on a stale worker. On current `main` the step would fail exactly as intended, which is the behaviour this PR is meant to lock in.

You are right that the two guards are not the same check, and that is deliberate: `make msw-worker-check` compares the tracked worker against `node_modules/msw/lib/mockServiceWorker.js` and so catches drift even where postinstall never runs (`--ignore-scripts`, a vendored cache), while the CI step catches the everyday case of an install dirtying the tree. Keeping both means neither weakness is load-bearing.

PR #6223, opened 2026-08-16
ci(release): stamp released binaries with their source revision

n0ctal:
## Summary

The release workflow builds `main.go`. Go treats a file argument as a file list, records the package path as `command-line-arguments`, and skips VCS stamping entirely — so published binaries carry no revision.

Building the package instead (`.`) with `-buildvcs=true` produces the same binary and adds `vcs.revision` / `vcs.time`.

## Why

Right now there is no way to map a released binary back to a commit. When a user reports a bug against "the latest release", the version string is all there is, and it does not distinguish rebuilds or tag moves. `go version -m` is the standard place to look and it currently answers nothing:

```
$ go version -m xui-release
xui-release: go1.26.6
        path    command-line-arguments
```

## Scope

Two lines in `.github/workflows/release.yml` — the Linux and the Windows build step. The `main` package contains only `main.go`, so the compiled output is unchanged; only the build target notation and the flag differ.

## Validation

From a clean clone at the same commit, built both ways:

| build target | `vcs.revision` |
|---|---|
| `main.go` (current) | absent |
| `.` (this change) | `4b0e9f9b6032fa9f4ddd144f6f042ab73792d371` |

Also confirmed `vcs.modified=false` under release conditions — the frontend bundle lands in the gitignored `internal/web/dist/`, so the tree stays clean and the stamp is trustworthy rather than permanently marked dirty.

## Risk

Low. No flag that affects code generation changes, and `-buildvcs=true` cannot fail the build here: the release job checks out a git repository, which is the only condition it requires.

Worth knowing: this does not extend to the Docker image, since `.dockerignore` excludes `.git` and Go silently omits the stamp outside a repository. Making container builds traceable is a separate change and is deliberately not attempted here.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -109,7 +109,7 @@ jobs:
           if [[ "$GITHUB_REF" != refs/tags/* ]]; then
             LDFLAGS="$LDFLAGS -X github.com/mhsanaei/3x-ui/v3/internal/config.buildCommit=${GITHUB_SHA::8} -X github.com/mhsanaei/3x-ui/v3/internal/config.buildDate=$(date -u +%Y-%m-%dT%H:%M:%SZ)"
           fi
-          go build -ldflags "$LDFLAGS" -o xui-release -v main.go
+          go build -buildvcs=true -ldflags "$LDFLAGS" -o xui-release -v .
           file xui-release
           ldd xui-release || echo "Static binary confirmed"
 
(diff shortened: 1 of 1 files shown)


n0ctal:
Belated reply — the fix went in but I never answered the review, which left a High finding looking unread. Sorry for that.

I took the second of your two options: `git` is in the msys2 `install:` list rather than dropping `-buildvcs=true`. The reasoning is that `auto` and `true` differ exactly when stamping is impossible — `auto` skips silently, `true` fails the build — and for a *release* workflow I would rather the build stop than publish an unstamped binary that looks identical to a stamped one. That only holds if the tool is guaranteed present, hence adding it explicitly rather than relying on the runner image.

Your diagnosis of why it would have failed was precise: `path-type` is unset so the default is `minimal`, `git` is not in the MSYS2 base set, and the manual PATH prepend at line 257 is itself the evidence that the Windows PATH is not inherited.

On the two smaller points: the `vcs.modified` observation about `gen:api` rewriting a tracked `openapi.json` is a good one and I have nothing to add — it means the flag reports generated-file drift as much as source edits, which is arguably useful. The Dockerfile stays out of scope, and I agree that if it ever moves to `.` it would need the `auto` default rather than `true`, since `.dockerignore` excludes `.git`.

PR #6224, opened 2026-08-16
ci: actually run the PostgreSQL schema and migration tests

n0ctal:
## Summary

`TestHostAutoMigrateCreatesColumns_Postgres` and `TestMigrate_Postgres` guard the PostgreSQL schema and the migration path, and both `t.Skip` unless `XUI_DB_TYPE` / `XUI_DB_DSN` are set. CI sets those only for the durable-first step, so neither test has ever executed — a green pipeline currently says nothing about either path.

## Why

The tests exist and are maintained, so the intent is clearly that they run. Right now a PostgreSQL schema regression reaches a release with CI fully green, and the skip is invisible: `go test` reports `ok` for the package either way.

The `postgres-durable-first` job already provisions a PostgreSQL 16 service, so no new infrastructure is needed — the DSN just has to reach these two tests as well.

## Scope

One step added to the existing job, mirroring the pattern already used one step above: run the two tests against the running service and fail if either reports `--- SKIP`.

The skip guard matters as much as the run: without it, a future rename of a test or of an environment variable would silently return the job to testing nothing, which is the failure mode this PR is fixing.

## Validation

Locally, against a PostgreSQL 16 instance:

| condition | result |
|---|---|
| no DSN (today's CI) | 2 tests SKIP |
| DSN set (this change) | 2 tests PASS, 0 SKIP |

## Risk

Low. No production code changes, no new services, and the added step can only fail if these tests genuinely fail or silently stop running.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,6 +70,16 @@ jobs:
             exit 1
           fi
 
+      - name: PostgreSQL schema and migration tests
+        run: |
+          set -o pipefail
+          XUI_DB_TYPE=postgres XUI_DB_DSN="host=127.0.0.1 port=5432 user=postgres password=postgres dbname=xui_durable sslmode=disable" \
+            go test ./internal/database -run '^(TestHostAutoMigrateCreatesColumns_Postgres|TestMigrate_Postgres)$' -count=1 -v | tee /tmp/postgres-schema.log
+          if grep -q -- '--- SKIP' /tmp/postgres-schema.log; then
+            echo 'PostgreSQL schema tests skipped instead of running' >&2
+            exit 1
+          fi
+
   codegen:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


n0ctal:
All three taken, in the pushed commit.

**The guard.** You are right, and the failure mode you describe is the one this PR exists to prevent, which makes it worse than an ordinary gap: asserting the absence of `--- SKIP` catches a test that ran and skipped, but a renamed or deleted test makes `-run` match nothing, `go test` prints `no tests to run`, exits 0, and the step is green having tested nothing.

Both steps now count `--- PASS` and require the expected number — at least one for durable-first, exactly two for the schema tests. Renaming either schema test now fails with `expected 2 passing PostgreSQL schema tests, got 1`.

**The workflow not testing itself.** Also right, and slightly embarrassing: this PR changes only `ci.yml`, and `ci.yml` was not in its own `paths` filter, so nothing here had run at all. `.github/workflows/ci.yml` is in both filters now, so this push should be the first time the job actually executes on the PR rather than first on `main` after merge.

**The DSN.** Hoisted to job-level `env` on `postgres-durable-first`; both `run` blocks lost their inline prefix.

Thanks for the first one in particular — a guard that cannot fail is worse than no guard, because it is read as coverage.

PR #6225, opened 2026-08-16
fix(inbounds): close the port check-and-claim race on the serial writer

n0ctal:
## Summary

`AddInbound` checks for a port conflict outside its transaction and then commits in a bare `db.Transaction`. Two overlapping creates both pass the read and both insert.

`UpdateInbound` and `SetRemoteTraffic` already run through `runSerializedTx`, so `AddInbound` is the one inbound writer left off the single traffic writer. Moving it there and folding the check into the transaction closes the window.

## Why

The check-and-claim gap is wide: tag resolution, client validation, email uniqueness and settings normalisation all sit between the read and the insert.

The case that no constraint can catch is a wildcard listener against a specific one on the same port — `("", 443, tcp)` and `("127.0.0.1", 443, tcp)` are two different rows, so a unique index accepts both. The panel's real rule lives in `listenOverlaps`, and only the semantic check can express it. That check is therefore load-bearing, and it is only sound if nothing can interleave between it and the insert.

## Scope

- `checkPortConflict` splits into a thin wrapper plus `checkPortConflictTx(db, ...)` that runs on a caller-supplied handle;
- `AddInbound` moves from `db.Transaction` to `runSerializedTx` and evaluates the conflict inside;
- `UpdateInbound` drops its pre-flight check and evaluates the same one inside the transaction it already has.

No new table, no environment variable, no migration, no additional locking. On SQLite `db.Transaction` takes the write lock at `BEGIN` (`_txlock=immediate`), and on PostgreSQL the single writer goroutine provides the serialisation, so both backends are covered by the existing primitives.

Two preconditions I checked before moving `AddInbound` onto the writer: it is reached only from the two controller handlers, never from inside another traffic write, so the queue cannot re-enter; and the node push is already deferred to `postCommitApply` after commit, which is what the writer contract requires.

## Validation

`TestAddInboundConcurrentOverlappingListenersSingleWinner` races a wildcard and a specific listener for the same port through `AddInbound`, 25 rounds, asserting exactly one commit.

Mutation-checked against today's behaviour: restoring the pre-flight check and the bare transaction turns it red with `concurrent AddInbound committed=2, want exactly 1`. That is the defect on `main`, reproduced.

`go build ./...`, `go vet`, `golangci-lint run` clean; `go test ./internal/web/service` green.

## Risk

Moderate and worth stating plainly: `AddInbound` now waits for the traffic writer, so a create issued during a traffic poll is serialised behind it instead of contending with it. That is the same trade `UpdateInbound` already makes, and it removes the `database is locked` failures a create could previously hit on SQLite after the busy timeout.

This supersedes #6185, which solved the same race with an opt-in reservation table. That table is rebuilt from `inbounds` on every boot and import, so it is derived rather than authoritative, and its unique index cannot express the wildcard overlap rule — the in-transaction check was doing the real work there too. I will close it in favour of this.

### DIFF
--- a/internal/web/service/inbound.go
+++ b/internal/web/service/inbound.go
@@ -1432,6 +1432,13 @@ func (s *InboundService) UpdateInbound(inbound *model.Inbound) (*model.Inbound, 
 	var postCommitApply func()
 
 	txErr := runSerializedTx(func(tx *gorm.DB) error {
+		conflict, cErr := checkPortConflictTx(tx, inbound, inbound.Id)
+		if cErr != nil {
+			return cErr
+		}
+		if conflict != nil {
+			return common.NewError(conflict.String())
+		}
 		if err := s.updateClientTraffics(tx, oldInbound, inbound); err != nil {
 			return err
 		}
--- a/internal/web/service/inbound.go
+++ b/internal/web/service/inbound.go
@@ -930,18 +930,11 @@ func (s *InboundService) AddInbound(inbound *model.Inbound) (*model.Inbound, boo
 		return inbound, false, err
 	}
 
-	conflict, err := s.checkPortConflict(inbound, 0)
+	tag, err := s.resolveInboundTag(inbound, 0)
 	if err != nil {
 		return inbound, false, err
 	}
-	if conflict != nil {
-		return inbound, false, common.NewError(conflict.String())
-	}
-
-	inbound.Tag, err = s.resolveInboundTag(inbound, 0)
-	if err != nil {
-		return inbound, false, err
-	}
+	inbound.Tag = tag
 
 	clients, err := s.GetClients(inbound)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


n0ctal:
Caught a real regression, thank you — fixed in the pushed commit.

You are right that `UpdateInbound` was left with no port-conflict check at all: I deleted the pre-flight and never added the in-transaction one. Editing an inbound onto an occupied port was accepted outright, and nothing failed, which is the part that bothers me most about it — the diff read as if the check had simply moved.

The check now runs as the first statement inside `runSerializedTx`, the same shape as `AddInbound`. `TestUpdateInboundRejectsPortTakenByAnother` covers the path: it seeds two inbounds, moves the second onto the first's port, and asserts both that the update is rejected and that the stored port is unchanged. Removing the check again turns it red with `moving an inbound onto a port already in use was accepted`.

Issue #5820, opened 2026-07-05
[Feature]: Complete frontend i18n coverage — eliminate ~30 hardcoded strings to attract more international users


LifetimeVip:
## Is your feature request related to a problem?

The 3x-ui frontend already ships 13 translation files (2,071 keys each) and has i18next fully wired up. But the dashboard, log modals, sidebar, sub page and some form labels still have ~30 user-visible strings hardcoded in English (Statistic titles, table headers, protocol labels, etc.). So even with a complete zh-CN or pt-BR translation file, users of those languages still see mixed English on core screens.

This is a quick gap to close — the i18n engine is already there, just not fully used.

## Describe the solution you'd like

Wrap the remaining hardcoded strings in `t('key')` and add corresponding keys to the translation files. Below is the full list — each entry is a direct code change, no design or architecture decisions needed.

### P0 — Dashboard overview page (`frontend/src/pages/index/IndexPage.tsx`)

| Line | Hardcoded | Suggestion |
|------|-----------|------------|
| 160 | `'Copied'` | use existing `t('copied')` |
| 312 | `title="Xray"` | use existing `t('pages.index.xrayStatus')` |
| 319 | `title="OS"` | new key `'pages.index.os'` |
| 427 | `title="IPv4"` | new key `'pages.index.ipv4'` |
| 434 | `title="IPv6"` | new key `'pages.index.ipv6'` |
| 448 | `title="TCP"` | new key `'pages.index.tcp'` |
| 455 | `title="UDP"` | new key `'pages.index.udp'` |
| 527-528 | `'Download'` / `'config.json'` | new keys |
| 537 | `'Copy'` | use existing `t('copy')` |

### P1 — Log modals & sidebar

| File | Line | Hardcoded | Suggestion |
|------|------|-----------|------------|
| `LogModal.tsx` | 111-116 | `Debug`, `Info`, `Warning`, `Error` log levels | new keys |
| `LogModal.tsx` | 122 | `'SysLog'` | new key |
| `XrayLogModal.tsx` | 27,168 | `DIRECT`, `BLOCKED`, `PROXY` event labels | new keys |
| `XrayLogModal.tsx` | 230-236 | Table headers `Date` `From` `To` `Inbound` `Outbound` `Email` | new keys |
| `XrayLogModal.tsx` | 187 | `'No Record...'` | new key |
[...]

PR #6226, opened 2026-08-16
feat(i18n): translate the log levels, access events and calendar labels

n0ctal:
Closes part of #5820.

## Summary

The log-level selector, the access-log event tags, the `Sub Formats` sidebar entry and the calendar choices were hardcoded English. A user on a fully translated locale still met English on the dashboard, the log modals and the settings page.

Eleven keys added across all 13 locales and wired up.

## Why the list is shorter than the issue's

The issue lists roughly thirty strings. I went through them against current `main` rather than by line number, and most should stay as they are:

- **Proper nouns** — `V2RayNG`, `Sing-box`, `Shadowrocket`, `Happ`, `Incy`, `V2Box`. Product names are not translated in any locale.
- **Protocol and format acronyms** — `SNI`, `uTLS`, `ALPN`, `TCP`, `UDP`, `IPv4`, `IPv6`, `DNS`, `JSON`, `CLASH`, `ID`. These are identical in every language the panel ships; a key per acronym adds 13 lines each and buys nothing.
- **Not user-visible** — `clients-export.json` and `config.json` are filenames, not labels.
- **Already fixed** — several dashboard entries the issue points at (`Xray`, `OS`, `Copy`) no longer exist at those lines.

What was left is the genuinely translatable remainder:

| what | keys |
|---|---|
| log levels: Debug, Info, Notice, Warning, Error | `pages.index.logLevel*` |
| access-log events: DIRECT, BLOCKED, PROXY | `pages.index.access*` |
| sidebar entry: Sub Formats | `menu.subFormats` |
| calendar choices: Gregorian, Jalalian | `pages.settings.calendar*` |

`'Copied'` on the dashboard now uses the existing `copied` key instead of a new one.

## Translations

English, Russian and Ukrainian are translated. The other ten locales carry the English string — the same convention those files already use, where between 4% and 9% of entries are currently untranslated English. I would rather leave a correct English fallback than guess at Farsi or Vietnamese; a native speaker can fill them in later without touching code.

## Two structural changes

Both were forced, not stylistic. The calendar list and the access-event map were module-level constants, and `t` only exists inside the component:

- `DATEPICKER_LIST` is gone; the two options are built inline where `t` is in scope.
- `EVENT_LABELS` became `EVENT_KEYS` and `eventLabel` takes `t`, resolving the key at render so it follows a language switch instead of freezing at import.

## Validation

`i18n-dead-keys` passes — it caught the three access keys while they were added but not yet referenced, which is exactly what it is for. `tsc --noEmit` clean, `eslint src` clean, and the unit suite is green (55 files, 936 tests).

### DIFF
--- a/frontend/public/mockServiceWorker.js
+++ b/frontend/public/mockServiceWorker.js
@@ -7,8 +7,8 @@
  * - Please do NOT modify this file.
  */
 
-const PACKAGE_VERSION = '2.14.7'
-const INTEGRITY_CHECKSUM = '4db4a41e972cec1b64cc569c66952d82'
+const PACKAGE_VERSION = '2.15.0'
+const INTEGRITY_CHECKSUM = '03cb67ac84128e63d7cd722a6e5b7f1e'
 const IS_MOCKED_RESPONSE = Symbol('isMockedResponse')
 const activeClientIds = new Set()
 
@@ -137,8 +137,18 @@ async function handleRequest(event, requestId, requestInterceptedAt) {
   if (client && activeClientIds.has(client.id)) {
     const serializedRequest = await serializeRequest(requestCloneForEvents)
 
+    // Omit the body of server-sent event stream responses.
+    // Cloning such responses would prevent client-side stream cancelations
+    // from reaching the original stream (a teed stream only cancels its
+    // source once both of its branches cancel) and would buffer the
+    // entire stream into the unconsumed clone indefinitely.
+    const isEventStreamResponse = response.headers
+      .get('content-type')
+      ?.toLowerCase()
+      .startsWith('text/event-stream')
+
     // Clone the response so both the client and the library could consume it.
-    const responseClone = response.clone()
+    const responseClone = isEventStreamResponse ? null : response.clone()
 
     sendToClient(
       client,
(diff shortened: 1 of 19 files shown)


n0ctal:
All three fixed in the pushed commit.

The export one is the finding I am glad you caught: I had made the contents of `x-ui.log` depend on the panel's UI language, and `EVENT=ЧЕРЕЗ ПРОКСИ` puts a space inside a field in a format whose other values are single tokens. A file people attach to bug reports and pipe through grep is a data format, not UI. The export now uses a separate `EVENT_TOKENS` map and keeps `DIRECT`/`BLOCKED`/`PROXY`; only the on-screen tag resolves a key.

The select had `width: 95`, sized for `Warning`, which clips `Предупреждение`. It is `minWidth` now, so it grows with the content.

And the three filter checkboxes now use the same keys as the tags they filter — you are right that a modal where the filter says `Direct` and the row says `НАПРЯМУЮ` is harder to read than one that is uniformly English.

I left the table headers and `No Record...` for a separate change rather than widening this one; they are the same class and worth doing, just not in a PR that is already touching four files.

PR #6227, opened 2026-08-16
feat(server): keep this machine's own settings when importing a database

### LINKED ISSUE #5004: [Feature]: Add toggles for not exporting host-specific options

The db export right now would also export machine specific configs such as cert details, which makes transferring to another host almost impossible.

In the export page, add options to not include ssl configurations

I tried manually edit the exported db files to remove those ssl lines. However, after uploading it to the panel, it overwrites the existing ssl configs.

### Concrete use case

Say I have 2 machines, with 3x-ui installed on machine A. Now I want to install the same config on machine B. I export the db from Machine A and import it to Machine B. And now the panel broke due to ssl config errors.

_No response_

Other

### Are you willing to help?

### Before submitting

n0ctal:
Closes #5004.

## Summary

Import replaces the whole database, so the uploaded file's listen addresses, ports, base path, certificate paths and node identity land on the destination machine. Moving a configuration to a new host leaves the panel answering on an address it does not own and presenting certificates it does not have — the reporter's "the panel broke due to ssl config errors".

This keeps the destination's own host-bound settings and takes everything else from the file.

## Why not filter on export

That was the shape the issue asked for, and it is the weaker half of the fix. `getDb` streams the SQLite file byte-for-byte, so filtering means building and shipping a modified copy — and it only helps exports taken *after* the change. Every backup already sitting on someone's disk still breaks on restore.

Doing it on import fixes those too, and removes the failure mode where the box was simply not ticked.

## What counts as host-bound

```
webListen  webDomain  webPort  webCertFile  webKeyFile  webBasePath
subListen  subDomain  subPort  subCertFile  subKeyFile  subURI  subJsonURI
secret  panelGuid
nodeMtls{CaCertPem,CaKeyPem,ClientCertPem,ClientKeyPem,ClientCertSha256,ClientCAPem}
```

The first two rows are the reachability the reporter lost. `secret` and `panelGuid` identify the panel itself. The mTLS material is the identity this panel presents to its nodes — importing another machine's would leave two panels claiming the same identity, which is worse than an unreachable port because it fails silently.

Everything else — inbounds, clients, users, templates, all remaining settings — still comes from the imported file, which is what the feature is for.

## Interface

A checkbox in the backup dialog, defaulting to keeping this machine's values. Clearing it restores today's behaviour for anyone deliberately cloning a host. The controller treats a missing field as "keep", so an API caller that has not been updated gets the safe path.

## Validation

`TestImportKeepsHostBoundSettings` seeds this machine's values, overwrites them the way an import does, restores, and asserts each host-bound key holds the local value while an ordinary setting (`remarkTemplate`) still holds the imported one — so the test fails both if preservation stops working and if it starts swallowing settings it should not.

Mutation-checked: making `restoreHostBoundSettings` a no-op turns it red with `setting subURI = "from-imported-file" after import, want this machine's "https://this-host.example/sub/"`.

`go build ./...`, `go vet`, `golangci-lint` clean; 36 Go packages green; frontend typecheck, eslint and the 936-test unit suite green.

## Risk

Low, and the sharp edge is named: someone who *wants* the source machine's certificates and ports must now clear the checkbox. That trade is deliberate — the silent-breakage direction is the one that costs an evening of debugging on a machine you have just migrated to.


n0ctal:
Both defects were real and are fixed in the pushed commit. Thank you — the second one meant the change did not work in the case it was written for.

**Absent rows.** The snapshot only recorded keys that had a row, so for every key where this machine relies on the built-in default — which is all the certificate paths, both listen addresses and all six node mTLS values — there was nothing to restore and the imported row survived. The snapshot now records which keys were absent and deletes the imported row for those, so the default applies again.

`TestImportDropsHostBoundSettingsThisMachineNeverHad` covers it: capture on a machine with no rows, seed imported ones, restore, assert they are gone. Skipping the delete turns it red with `imported webCertFile survived although this machine had no row for it`.

**PostgreSQL.** `importPostgresDB` took the flag and dropped it. `restorePostgresDump` now captures before `CloseDB` and restores after the successful `InitDB`, the same placement as the SQLite path.

MHSanaei:
@claude resolve conflicts

PR #6230, opened 2026-08-16
feat(limitip): let operators exempt trusted addresses from the IP limit

### LINKED ISSUE #5378: [Feature]: Add IP Whitelist (ignoreip) Support for IP Limit Fail2ban Protection

Please consider adding support for an IP whitelist (ignoreip) for the IP Limit protection system.

Currently, administrators must manually edit Fail2ban configuration files to exclude trusted IP addresses or networks from automatic bans.

Add an IP whitelist section in the 3x-ui panel that allows administrators to:

Add individual IP addresses
Add CIDR subnets (e.g. 192.168.1.0/24)
Edit and remove whitelist entries
Automatically apply entries to the Fail2ban ignoreip configuration
Benefits
Easier administration
No manual Fail2ban configuration edits
Configuration survives updates
Better support for office and enterprise deployments
Reduced risk of accidentally blocking trusted networks
Example
Trusted IPs:
203.0.113.10
198.51.100.0/24

IPs on the whitelist would never be banned by the IP Limit protection system.

_No response_

### Concrete use case

In many environments, multiple legitimate users share the same public IP address:

Office networks
Corporate gateways
Schools and universities
Shared residential networks
Self-hosted deployments with trusted users

When one user triggers an IP limit violation, the shared public IP may be banned, affecting other legitimate users behind the same NAT.

_No response_

Frontend (UI / panel pages)

### Are you willing to help?

### Before submitting

n0ctal:
Closes #5378.

## Summary

Behind a shared address — an office gateway, a campus NAT, a residential carrier — every user looks like the same client to the IP limit. One of them trips it, the address is disconnected and written to the fail2ban log, and everyone else behind that address goes with them.

This adds an allowlist of addresses and networks the IP limit leaves alone.

## Why "not counted", not just "not banned"

The obvious reading of the request is "never ban these". That is half a fix. If the shared address still occupies one of the client's slots, the next ordinary connection pushes the client over the limit and something else gets banned instead — the office is still the reason someone was cut off, one step removed.

So a matching address is taken out of the accounting entirely: it is neither banned nor counted. That is what makes the setting do what the reporter's use case needs.

## Behaviour

- Setting `ipLimitAllowlist`, comma-separated, each entry an IP or a CIDR, IPv4 and IPv6, empty by default — so nothing changes for anyone who does not set it.
- Entries are **validated on save** and rejected if malformed, rather than skipped silently at scan time. A typo in an allowlist is invisible: the address simply stays unprotected until someone notices it getting banned.
- A read failure at scan time leaves the list empty, which enforces the limit as before rather than exempting everybody.

## Why not write fail2ban's `ignoreip`

The panel would have to own a file under `/etc/fail2ban/jail.d`, which needs root, a reload, and a policy for what happens when an operator has edited it by hand. It is also fragile in a way that is easy to miss: fail2ban merges `[DEFAULT]` sections in filename order, so a later file silently overrides an earlier one and the panel's intent quietly disappears.

Filtering before the log line is written means fail2ban never sees anything to act on, needs no privileges, and cannot be overridden by a stray config file.

## Validation

Unit tests cover matching (address, CIDR, IPv6, mixed separators), malformed entries not widening the list, and the empty list being a no-op.

The wiring is covered separately, and that matters: my first attempt tested only the pure functions, so replacing the filtered list with the unfiltered one at the call site left everything green. `TestUpdateInboundClientIps_AllowlistedIpIsNeitherCountedNorBanned` mirrors the existing ban test through `updateInboundClientIps` and asserts nothing is banned, both addresses stay persisted, and no line reaches the fail2ban log. Reverting the filter turns it red with `an allowlisted address pushed the client over its limit and something was banned`.

All examples use the documentation ranges from RFC 5737 and RFC 3849.

`go build ./...`, `go vet`, `golangci-lint`, `make gen-check` clean; 36 Go packages green; frontend typecheck, eslint and the i18n key test green.


n0ctal:
All three fixed in the pushed commit.

The doc-comment one is the kind of mistake I would rather not repeat: `loadAllowlist` landed between `hasLimitIp`'s comment and `hasLimitIp` itself, so godoc attributed one function's rationale to another and left the `clients.limit_ip` probe undocumented. Moved below `hasLimitIp`, each comment back on its own function.

On the grammar — you are right that it was a promise the software never keeps. The parser advertised semicolons and whitespace while `CheckValid` splits on commas only, so those forms are unreachable through the panel and the API and would be rejected at save. Both sides now read the same comma-separated grammar, and the comment says so instead of claiming parity with `trustedProxyCIDRs`, whose runtime parser is comma-only too.

`internal/web/dist/.gitkeep` was a `make dist-stub` artifact that should never have been staged. Removed.

n0ctal:

MHSanaei:
@claude resolve conflicts

PR #6238, opened 2026-08-17
feat(clients): cap how many times a client may auto-renew

### LINKED ISSUE #5804: [Feature]: Add a maximum renewal count for client Auto Renew

I manage some clients with fixed-term or prepaid plans.

Image

Currently, the client field "Auto Renew (days)" is useful, but once it is enabled, the client keeps renewing indefinitely after expiration. This works for unlimited recurring plans, but it does not fit fixed packages.

For example:

- Initial client duration: 30 days
- Auto Renew (days): 30
- Maximum renewal count: 3

The expected total service period should be:

30 initial days + 3 renewals × 30 days = 120 days total, approximately 4 months.

At the moment, 3X-UI can auto-renew the client repeatedly, but there is no built-in way to stop auto-renewal after a fixed number of renewal cycles.

Please add an optional "Max Auto Renew Count" field for each client, next to the existing "Auto Renew (days)" field.

Image

Suggested behavior:

- Auto Renew (days) = 0:
  - Auto Renew remains disabled, same as the current behavior.

- Auto Renew (days) > 0 and Max Auto Renew Count is empty or 0:
  - Keep the current behavior: unlimited auto-renewal.
  - This preserves backward compatibility.

- Auto Renew (days) > 0 and Max Auto Renew Count > 0:
  - The client can only be auto-renewed this many times.
  - Each successful auto-renewal should increase a stored "renewals used" counter.
  - After the maximum count is reached, the client should no longer be auto-renewed on the next expiration.

Example:

- Initial expiry: day 30
- Auto Renew (days): 30
- Max Auto Renew Count: 3

Expected timeline:

1. Day 30: first auto-renewal, new expiry becomes day 60, renewals used = 1/3
2. Day 60: second auto-renewal, new expiry becomes day 90, renewals used = 2/3
3. Day 90: third auto-renewal, new expiry becomes day 120, renewals used = 3/3
4. Day 120: no more auto-renewal; the client expires normally

The existing auto-renew behavior should remain the same during each allowed renewal:
- Extend the expiry time by the configured Auto Renew days
- Reset the traffic counters for the new quota window
- Re-enable the client if the current auto-renew logic already does that

The new setting should only limit how many times this process can happen.

Suggested UI:

- Auto Renew (days): [30]
- Max Auto Renew Count: [3]
- Renewals Used: 0/3

Suggested backend/API changes:

- Store the maximum renewal count per client.
- Store the number of renewals already used per client.
- Update the auto-renew logic to check the remaining renewal count before renewing.
- If the panel catches up multiple missed periods, each added renewal period should consume one renewal count.
- Expose these fields through the client API so they can be shown and edited in the panel.

I considered manually disabling Auto Renew after several cycles, or setting a final expiry date manually. However, this is error-prone when managing many clients.

External scripts are also possible, but they duplicate logic that already exists in 3X-UI. Since the current Auto Renew logic already handles expiry extension and traffic reset, adding a renewal limit inside the panel would be more reliable and easier to maintain.

### Concrete use case

A client buys a fixed 4-month package with monthly quota reset.

Configuration:

- Initial duration: 30 days
- Auto Renew (days): 30
- Max Auto Renew Count: 3

Expected result:

- The client gets the initial 30 days.
- Then the client is auto-renewed 3 times.
- Traffic is reset at each renewal, as Auto Renew currently does.
- The total active period is 120 days, approximately 4 months.
- After the third renewal has been used, the client should expire normally at the next expiry time instead of being renewed forever.

_No response_

Frontend (UI / panel pages), Backend (API endpoints, login, settings), Statistics / traffic counters, Database / migrations

### Are you willing to help?

### Before submitting

n0ctal:
Closes #5804.

## Summary

Auto-renew runs forever. A prepaid or fixed-term client keeps receiving new periods until an operator remembers to switch it off, and there is no way to say "renew this three times, then let it lapse".

This adds a per-client maximum. Zero keeps today's behaviour, so nothing changes for anyone who does not set one.

## The part that is easy to get wrong

Renewal catches up: if a client's expiry is three periods in the past, the existing loop advances it until it reaches the present. With a cap, that loop has to stop counting as well as stop renewing, otherwise a client who was away for three cycles receives all three regardless of what the operator paid for — the cap would only apply to clients who happened to be seen every period.

So the catch-up spends one allowance per period and stops at the last period the cap covers. A client capped at two, three periods behind, ends up exactly two periods forward — still expired, and correctly so, rather than jumping to the present.

## Behaviour

- `resetMax` per client, default 0 = unlimited.
- `resetCount` records how many renewals have fired.
- Selection skips clients that have spent their allowance, so they expire like any client without auto-renew.
- Both fields ride the existing settings/traffic sync, so a client edited through the panel or the API keeps them in step.

## Validation

Three tests: a capped client that has spent its allowance is not renewed while its neighbour with one left is; a three-periods-behind client capped at two lands exactly two periods forward and stays expired; a client with no cap renews as before.

Mutation-checked, both halves separately:

- dropping the selection filter → `renewed count = 2, want 1: only the client with an allowance left`
- dropping the cap check inside the catch-up loop → `reset count = 4, want the 2 the cap allowed`

`go build ./...`, `go vet`, `golangci-lint`, `make gen-check` clean; 36 Go packages green; frontend typecheck, eslint and the unit suite green.

## Risk

Low. The new column defaults to 0 on existing rows, which is "no cap", so upgraded installs behave exactly as before until someone sets a value.


n0ctal:
Fixed in the pushed commit — the blocking one plus the rest.

**Persistence.** Same defect as the one you found in #6239, and the same shape: `resetMax` never reached the `clients` table, so the edit dialog read back 0 and saving an unrelated change lifted the cap. `reset_max` is on `ClientRecord` now and goes through `ToRecord`, `ToClient`, `applyClientRecordMerge`, the record update map and `ClientSlim`.

`TestClientEditKeepsTheRenewalCap` drives the real edit path and asserts the inbound JSON, the record and `client_traffics`; removing the field from `ToClient` fails with `inbound settings resetMax = 0 after an unrelated edit, want 3: the cap was silently lifted`.

**The truncated catch-up.** Taken as recommended. When the cap runs out mid-catch-up the expiry is still in the past, so the renewal side effects now do not fire at all: no counter zeroing, no enable, no `AddUser`. Your reading of what happened was exact — `disableInvalidClients` undid it in the same transaction, so the committed state was fine and the cost was pure churn, plus the client's accumulated up/down being thrown away for periods it can never use. `TestAutoRenewClients_TruncatedCatchUpLeavesTheClientDisabled` pins `Enable` and both counters; restoring the old order fails with `a client still expired after a truncated catch-up was enabled: xray gains a user only to lose it again`.

**Visibility.** `resetMax`/`resetCount` are in the hand-written `ClientTrafficSchema` and the client info modal shows renewals used against allowed, going red when the allowance is spent. You put your finger on the real gap: the only lever was raising the cap, and the operator could not see the number they were raising it past.

I did not add a clearing rule in this PR. There are two defensible ones — clear on manual traffic reset, or an explicit control — and picking one silently seemed worse than leaving the count visible and saying so. If you have a preference I will add it here.

**The negative case.** The predicate is `reset_max <= 0 or reset_count < reset_max`, so any non-positive value means unlimited, and a negative one is now rejected at the service layer. `-1` meaning "unlimited" silently producing "never renew again" was the worst of the small findings.

**Smaller ones.** Comment trimmed to two lines. One doc line per field, so `resetCount` gets its own description in the generated schema instead of inheriting `resetMax`'s. Bulk-add modal got the input. `renewMax`/`renewMaxDesc` are translated in all 13 locales now rather than carrying English in 11 of them. `internal/web/dist/.gitkeep` dropped — it was a `make dist-stub` artifact, and you are right that committing it would quietly turn a loud build failure into a blank panel.

**On the docs.** I regenerated the API reference and reverted it: the copy under `docs/` is far enough behind `main` that the resync was 3676 added lines for the sake of a couple of new fields. Separate resync PR if you want it.

**And the product question you raised.** You are right that a capped client which goes offline can spend its whole allowance on periods already in the past and never get usable time. That is unchanged here — I only stopped the churn. If the intent is that an allowance should only be spent on a period the client can actually use, that is a different rule and I would rather you choose it than have me assume it.

n0ctal:
Follow-up: the review on #6240 found a defect that this PR shares, so I have fixed it here too.

`ClientService.Update` persists the record columns directly only when the client has no inbounds. For a client attached to at least one — the normal case — persistence goes through `SyncInbound` → `applyClientRecordMerge`, which I had not extended. The result is that the new field could be set at creation and never changed afterwards: the edit updated the inbound settings JSON while the `clients` column kept its original value, and the renewal query kept using that.

I confirmed it rather than assuming it carried over, with a throwaway test that changed the value on an attached client and read the column back.

The existing test did not catch this, and the reason is worth stating plainly: it asserted that the value *survives an unrelated edit*, and it survives precisely because nothing on that path ever writes it. A test that passes because of the bug is worse than no test. Each PR now has a test that changes the value and then clears it, and both fail if the record write is removed.

Fixed in the pushed commit; no other behaviour changed.

n0ctal:

PR #6239, opened 2026-08-17
feat(clients): renew on a calendar day instead of a rolling interval

### LINKED ISSUE #6106: [Feature]: Support calendar-based monthly Auto Renew for clients

Client Auto Renew currently accepts only a fixed interval in days. A value such as 30 creates a rolling 30-day cycle anchored to the client's expiry time, so it drifts relative to calendar-month billing cycles because months have 28, 29, 30, or 31 days.

Related to #6071 and #3462. #6071 adds a configurable day-of-month for inbound traffic resets, but it does not change the per-client Auto Renew schedule or advance client expiry dates on a calendar boundary.

Add a calendar-based renewal mode to each client's Auto Renew setting while keeping the current interval-days mode for backward compatibility.

Suggested behavior:

- Renewal mode: `Interval (days)` or `Calendar monthly`.
- For `Calendar monthly`, configure a renewal day from 1 to 31 and use the panel's configured time zone.
- On the renewal boundary, reset the client's traffic counters, advance its expiry to the next calendar renewal date, and re-enable it if it was disabled because of quota or expiry.
- If the selected day does not exist in a shorter month, renew on that month's final day, consistent with #6071.
- Prevent duplicate renewal within the same billing period.
- If the panel was offline at the scheduled boundary, perform one catch-up renewal after startup.
- Existing clients using the integer day interval must continue to behave exactly as they do today.

- Setting Auto Renew to 30 days: the renewal date drifts across calendar months.
- Using inbound `Traffic Reset = Monthly`: this resets traffic counters, but it does not provide calendar-based per-client expiry renewal.
- Using an external cron job or API script: possible, but fragile and duplicates logic already owned by the panel.

### Concrete use case

A server or provider plan resets its monthly allowance on the first day of each calendar month at 00:00 UTC. Clients are assigned monthly traffic quotas and should start a fresh quota/expiry window on that same boundary.

With Auto Renew set to 30 days, a client renewed on August 1 would renew on August 31, then September 30, and continue drifting away from the provider's billing cycle. A calendar-month mode with renewal day 1 and the panel time zone set to UTC would keep the client lifecycle aligned.

Example client form:

```
Auto Renew mode: [Interval (days) | Calendar monthly]
Renewal day:     [1]
Time zone:       Panel time zone (UTC)
```

This complements #6071: that PR schedules inbound traffic resets by day of month, while this request applies calendar scheduling to each client's Auto Renew lifecycle.

Frontend (UI / panel pages), Backend (API endpoints, login, settings), Statistics / traffic counters, Database / migrations, Multi-node (sub-nodes)

### Are you willing to help?

### Before submitting

n0ctal:
Closes #6106.

## Summary

Auto-renew advances the expiry by a fixed number of milliseconds. A client set to 30 days drifts against the calendar: renewing on 31 January lands on 2 March, and after a year the billing day has wandered a fortnight from wherever the operator's own plan resets.

This adds a per-client renewal day. When set, the expiry steps whole calendar months at midnight in the panel's configured time zone.

## The two cases worth naming

**Short months.** A client billed on the 31st renews on 28 February and returns to 31 March. It does not decay to the 28th permanently, which is what "advance by one month" does if you implement it as "add 30 days" or let `time.Date` roll 31 February forward into March.

**Daylight saving.** Because the step is a calendar operation in the panel's zone rather than an addition of hours, a renewal that crosses a DST boundary still lands on local midnight instead of drifting an hour twice a year.

## Behaviour

- `resetDay` per client, 0–31; 0 keeps today's interval mode, so existing clients are untouched.
- Selection accepts either mode; the advance loop picks the calendar step when a day is set.
- Catch-up works as before: a client several months behind steps month by month until its expiry is in the future, landing on the billing day rather than on "today plus a month".
- The panel's time zone is read once per run, falling back to UTC with a warning if the setting cannot be read — renewals continuing is better than renewals stopping because of a settings read.

## One defensive change

The interval branch now refuses a zero step. It is unreachable while the selection filter holds, and I kept it anyway: that loop is `for expiry < now { expiry += reset*24h }`, it runs on the single traffic writer, and a zero interval reaching it would hang every panel mutation queued behind it. I found this while mutation-testing — removing the calendar branch turned the suite from failing into hanging, which is a considerably worse way to learn about it.

## Validation

Date arithmetic is covered directly: clamping in February (leap and common), returning to the 31st afterwards, 30-day months, December rolling into January, same-month vs next-month boundaries, the panel zone being honoured, and a DST crossing still landing on local midnight.

End to end through `autoRenewClients`: a calendar client lands on its billing day with counters reset and re-enabled; a client billed on the 31st gets the clamped day; an interval client still moves by exactly the old fixed step.

Mutation-checked: disabling the calendar branch turns the calendar test red; removing the short-month clamp turns the clamping tests red.

`go build ./...`, `go vet`, `golangci-lint`, `make gen-check` clean; 36 Go packages green; frontend typecheck, eslint and the unit suite green.


n0ctal:
The blocking one was real and the change did not work without it — fixed in the pushed commit, along with the rest.

**Persistence.** `resetDay` never reached the `clients` table, so anything that rebuilt a client from the record wrote it back as zero. The failure I care about most is the quiet one you named: an operator opens a calendar client to fix a comment, saves, and the billing day is gone — nothing to see until the client fails to renew a month later. `reset_day` is on `ClientRecord` now and goes through `ToRecord`, `ToClient`, `applyClientRecordMerge` and the record update map.

`TestClientEditKeepsTheBillingDay` drives the actual edit path — hydrate the record, change the comment, save — and asserts the inbound JSON, the record and `client_traffics`. Both halves of the converter are pinned: dropping it from `ToClient` fails with `inbound settings resetDay = 0 after an unrelated edit`, dropping it from `ToRecord` fails on the record assertion.

Worth recording that my first attempt at this test was wrong. I drove `ResetTrafficByEmail` instead, and it passed with the converter mutated — the client record stays enabled, so the auto-enable branch that does the rebuild never ran. A test that goes green under the mutation it exists to catch is worse than no test, so I replaced it rather than keeping both.

**Multi-node.** Gated `nodeClientRenewed` on `cs.Reset > 0 || cs.ResetDay > 0` and carried `reset_day` through the snapshot row build and both UPDATE statements. You are right that a single-box reading hides this: on a master/sub-node deployment the renewal happens on the node, and with the master refusing to classify it as one, the stale pushed totals re-deplete the client immediately.

**Validation.** `0 <= resetDay <= 31` is enforced in the service layer and rejected rather than clamped, for the reason you gave — clamping 45 to 31 and dropping -1 out of the query are both silent divergences from what was submitted.

**The list.** The `autoRenew` filter now accepts either mode and `ClientSlim` carries `resetDay`.

**The two tests you called out.** Both were fair.

The clamp test derived its expectation from the value under test, so it asserted a tautology. It now compares against an absolute instant produced by a helper that walks forward a day at a time and gets month length from `time.Date`'s zero-day trick — deliberately not built on `nextCalendarRenewal`, so the two can disagree.

The hang detector is gone. You are right that the diagnostic was worse than the bug: on the failure it was written for, the goroutine outlives `t.Cleanup` closing the DB and panics the run. The test now asserts the selection query directly.

**Smaller ones.** Comment blocks trimmed to two lines and the issue reference corrected to #6106. `renewDay` renamed to `renewOnDay` — you are right that it sat too close to `renewDays` — and both keys are now translated in all 13 locales rather than carrying English. Added a `RESET_DAY` subscription placeholder with its description in all 13, since `RESET_DAYS` is a period and reusing it would have been a different meaning under the same name. Bulk-add modal got the field. `internal/web/dist/.gitkeep` dropped.

**One I did not take.** I regenerated the docs API reference and then reverted it: `docs/public/openapi.json` is far enough behind `main` that the resync came to 3676 added lines, of which 8 were `resetDay`. That would bury the feature in unrelated drift. Happy to send it as its own resync PR if you want one.

n0ctal:
Follow-up: the review on #6240 found a defect that this PR shares, so I have fixed it here too.

`ClientService.Update` persists the record columns directly only when the client has no inbounds. For a client attached to at least one — the normal case — persistence goes through `SyncInbound` → `applyClientRecordMerge`, which I had not extended. The result is that the new field could be set at creation and never changed afterwards: the edit updated the inbound settings JSON while the `clients` column kept its original value, and the renewal query kept using that.

I confirmed it rather than assuming it carried over, with a throwaway test that changed the value on an attached client and read the column back.

The existing test did not catch this, and the reason is worth stating plainly: it asserted that the value *survives an unrelated edit*, and it survives precisely because nothing on that path ever writes it. A test that passes because of the bug is worse than no test. Each PR now has a test that changes the value and then clears it, and both fail if the record write is removed.

Fixed in the pushed commit; no other behaviour changed.

n0ctal:

PR #6240, opened 2026-08-17
feat(clients): give each client its own traffic reset cycle

### LINKED ISSUE #5497: [Feature]: Add individual auto-reset cycles for each client

I'm currently managing multiple clients with different traffic plans. While 3X-UI supports periodic traffic reset at the inbound level, it resets traffic for all clients in that inbound simultaneously.

For example:
- Client A has a monthly plan with 1000GB, resetting on the 1st of each month.
- Client B has a weekly plan with 200GB, resetting every Monday.

Currently, I have to manually click the "Reset Traffic" button for each client, which is inefficient and error‑prone.

I would like to see a feature that allows each client to have its own independent auto‑reset cycle. This could be implemented by adding the following options in the "Edit Client" interface:

1. **Enable Auto‑Reset** – a toggle to turn the feature on/off for this specific client.
2. **Reset Cycle** – a dropdown menu to select `Daily`, `Weekly`, `Monthly`, or `Custom (Cron Expression)`.
3. **Reset Rules**:
   - If `Weekly` is selected, an additional option to choose the day of the week (e.g., Monday, Sunday).
   - If `Monthly` is selected, an option to choose the day of the month (e.g., 1st, 15th).
   - If `Custom` is selected, a text field for a Cron expression (e.g., `0 0 5 * *` for the 5th of each month).
4. **Traffic Quota After Reset** – an option to either `Reset to the original limit` or `Reset to a custom value` (in GB).
5. **Behavior** – the traffic counter (Up/Down) should be reset to zero, and the `Total` traffic should be restored to the specified value, regardless of how much was used in the previous cycle.

On the backend side, each client record should store the reset settings (e.g., `reset_enabled`, `reset_period`, `reset_cron`, `reset_quota`). A new scheduled job (or an extension of the existing `PeriodicTrafficResetJob`) should check each client’s individual settings and apply the reset accordingly.

The current inbound‑level reset is a good start, but it does not support clients with varying billing cycles. This feature would greatly improve management efficiency for those of us handling many clients with diverse plans. Thank you for considering this request!

_No response_

### Concrete use case

_No response_

_No response_

_No response_

### Are you willing to help?

### Before submitting

n0ctal:
Closes #5497.

## Summary

Traffic reset lives on the inbound, so every client sharing one resets together. The issue's case is an operator running a monthly 1000GB plan next to a weekly 200GB plan: today that means pressing **Reset Traffic** by hand, per client, on a schedule kept in someone's head.

Clients now carry their own cycle.

## Shape

The client fields are the inbound fields: `trafficReset` (`never | hourly | daily | weekly | monthly`) and `trafficResetDay` for the monthly case, same vocabulary, same clamping, same monthly due-day rule via the existing `monthlyResetDue`. `PeriodicTrafficResetJob` already runs once per period; it now makes a second pass over the clients whose own cycle matches that period, after the inbound pass.

A client left at `never` behaves exactly as before — only its inbound's schedule touches it. A client whose own cycle happens to match its inbound's is reset twice in the same tick, which is harmless because the reset is idempotent, and I preferred that to making one schedule quietly suppress the other.

## What I left out, and why

The issue also asks for cron expressions and a choice of post-reset quota. I did not implement either. Both are product decisions rather than mechanics, and neither has an inbound-level counterpart, so adding them only for clients would put two different scheduling languages in one panel. The weekly day-of-week choice is out for the same reason: inbound weekly resets fire on the `@weekly` boundary, and a client-only weekday would diverge from it. Happy to add any of them if you want them — I would rather ask than guess.

## Persistence

The fields are on `ClientRecord`, not only in the inbound settings JSON. That is deliberate and load-bearing: the paths that rebuild a client from the record — an ordinary edit, an attach, a traffic reset — write back whatever the record holds, so a field that lives only in the JSON is silently zeroed by routine panel use and the operator finds out a billing cycle later.

An unknown period is rejected rather than coerced to `never`. A coerced value leaves a field that reads as configured while no job ever selects the client, which is the worst of the three options.

## Validation

Three job tests against a real DB: a weekly client inside a `never` inbound is reset by the weekly run while a monthly client is not; a client at `never` survives all four periods; a monthly client is reset on its own day and not on another's.

Mutation-checked — removing the client pass fails with `weekly client not reset by the weekly run: up=500 down=700`, and removing the monthly due-day filter fails with the not-due client reset.

Persistence is covered by driving the actual edit path (hydrate the record, change a comment, save) and asserting the inbound JSON, the record and `client_traffics`; dropping the field from `ToClient` fails with `inbound settings hold ""/15 after an unrelated edit`. Validation is table-driven over the accepted and rejected shapes.

`go build ./...`, `go vet`, `golangci-lint`, the full Go suite, frontend typecheck, eslint and 1158 unit tests all green.


n0ctal:
All fixed in the pushed commit. One of your findings led me to a second bug that was mine and worse than the first, so I want to set that out rather than bury it.

**The blocking one.** Exactly as you described. The cycle could only be set at creation: `ClientService.Update` writes the columns directly only in the `len(inboundIds) == 0` branch, and the normal path goes through `SyncInbound` → `applyClientRecordMerge`, which I had not extended. Fixed there.

Your parenthetical about the test is the part worth repeating: *"the current test only asserts the value survives an unrelated edit, which it does precisely because nothing ever writes it."* That is right, and it is the second time in two days I have written a test that goes green on the bug it exists to catch. `TestClientEditChangesTheTrafficResetCycle` now changes weekly→monthly/9 and then switches the cycle off; removing the merge line turns it red.

I checked my other two PRs against the same pattern and found the same defect in both — #6238 and #6239 also could not change or clear their new field after creation. Fixed and pushed there too, each with a change-the-value test. Thank you for that one; it was worth three findings.

**The seed with no `client_inbounds` link.** Correct, and it is why the suite was blind. Every reset was falling through `GetInboundIdsForRecord` into the orphaned-client branch. The helper creates the link now, and there are cases for a disabled record.

**Now the one I got wrong.** You offered three ways to stop a cron job from re-enabling clients, and I took the middle one — reset the counters without the re-enable side effect. That is a worse bug than the one it fixes, and my new test asserted the broken state as correct.

Depletion does not only clear `client_traffics.enable`: `disableInvalidClients` also sets `clients.enable = false` and rewrites the settings JSON. `GetXrayConfig` builds from the `clients` table, so lifting the quota gate alone leaves a depleted client out of the running core — with zeroed counters that no longer match the depleted predicate, so nothing ever revisits the row. The client is locked out until a human notices. Skipping disabled records, your first option, has the same effect for the same reason: the clients that need a reset most are exactly the disabled ones.

So the rule is now about *cause*, not state: a client whose disable is explained by depletion (`total > 0 && up+down >= total`) is restored; one disabled below its quota was switched off by hand and is skipped entirely. Both directions are pinned — reverting to a plain `!c.Enable` skip fails with `clients.enable still false: GetXrayConfig skips the client, so it stays locked out for good`, and dropping the check fails with `an operator-disabled client was switched back on by a cron job`.

**Node propagation and the sidecar.** Related, and also mine. My bulk path bypassed both the node runtime push and the MTProto quota reset that `ResetClientTraffic` performs, so it would have been a silent no-op on node-backed inbounds — the failure CLAUDE.md warns is invisible in a single-box reading. Dropped in favour of `ResetTrafficByEmail`, which handles both.

**Which means I did not take the performance finding**, and I want to be explicit rather than quiet about it. The per-client path costs one serialized transaction per attached inbound, and you are right that a daily cycle on thousands of clients is a lot of them at midnight. A bulk path would have to carry node propagation and the sidecar reset before it is safe, and I would rather ship the correct slow version than the fast one I just demonstrated I get wrong. Happy to do it as a follow-up if you want it, or now if you would rather it land together.

`needRestart` is collected across the loop and turned into a single `SetToNeedRestart()`.

**The smaller ones**, all taken: the `== "never"` clause is gone from the merge rule; the fields are guarded like `Group`/`AdTag` so a stale node snapshot cannot erase a configured cycle; `migrateClientTrafficResetColumns` backfills the NULLs an older SQLite `ALTER TABLE` leaves; `validateClientTrafficReset` now also guards the bulk-create and portable-import paths; the day goes through `normalizeTrafficResetDay` at the write instead of being clamped by three layers downstream; both fields are `omitempty` so existing clients match the published contract; the query selects five columns instead of hydrating credentials; the `defer` is a plain call; the bulk-add modal has the controls; the dead `otherDay` guard is gone and the tests are `t.Run` subtests. `TRAFFIC_RESETS` now lives in `@/schemas/primitives` and the inbound form, client form and bulk-add form all import it, so the tuple exists once.

I left the msw worker out — you were right that it was unrelated — and the docs resync out, for the reason I gave on #6239: `docs/public/openapi.json` is far enough behind `main` that the regeneration is ~3700 lines of drift.

**Merge order**, since you asked: #6238 first, then #6239, then this one. #6238 is the smallest and the other two rebase cleanly onto it; this one is the largest and touches `applyClientRecordMerge`, which the other two now also touch. I will regenerate the artefacts and rebase whichever are still open after each lands.

Issue #5825, opened 2026-07-06
[Feature]: Client-side balancer


isleuntothyself:
### Is your feature request related to a problem?

3x-ui supports load balancers but they are server-side only, so there is no way to create a "fastest location" or "auto-pick" option in a subscription (like many commercial VPN services do)

### Describe the solution you'd like

i would love 3x-ui to support a client-side balancer that can be included in the subscription (JSON or other formats)

specifically:

add an option to include a balancer outbound in a subscription (with the ability to configure it as if you would do with a balancer) that the client can use locally. the client would then ping the available outbounds and automatically connect to the suitable one

### Alternatives you've considered

in some panels (like Remnawave) this can be achieved by customizing the JSON subscription, but switching to other panels is not really what i want 

i know that some proxy clients have the ability to sort the inbounds by ping (basically picking "the fastest one") but a lot of my users use other clients that don't have such a feature

### Which parts of the panel would this affect?

Subscription (share links / Clash / JSON), Xray config generation, Backend (API endpoints, login, settings), Frontend (UI / panel pages)

### Are you willing to help?

- [ ] I'd like to implement this feature myself and open a PR.
- [ ] I can help test once a PR is open.

### Before submitting

- [x] I searched [existing issues](../issues?q=is%3Aissue) and this feature has not been requested.
- [x] This feature request is written in English.

DIMFLIX:
bump

varp:
+1

isleuntothyself:
implemented in 3.7.0, thanks @DIMFLIX

PR #6243, opened 2026-08-17
feat(sub): client-side balancers for the JSON subscription

DIMFLIX:
## Summary

Adds **Subscription Balancers** — admin-defined balancers emitted into the JSON subscription as extra config documents, so a client profile can auto-pick the best of several inbounds' endpoints (Xray `routing.balancers` + `burstObservatory`) instead of listing them one by one. Managed in **Settings → Sub Balancers**, with an **Observatory** sub-tab for probe parameters, mirroring the Xray Balancers page layout.

Supported strategies: `random`, `roundRobin`, `leastPing`, `leastLoad` (Xray routing balancer strategies).

## Why

Today a JSON-sub client gets one config document per inbound, in the inbound's own order. When several inbounds are meant to be one logical "best of" profile, the only option is to ship them all separately and let the client (or the user) pick — there's no panel-side way to emit a single `routing.balancers` entry that auto-selects the fastest/least-loaded member outbound.

This gives admins that lever: define a balancer (remark, strategy, member inbounds, sort order), and every client sitting on at least one of those inbounds gets one extra subscription document that uses a `balancer` routing rule + a `burstObservatory` to probe and pick the best member outbound.

The same client-side injection pattern is used by Remnawave's `injectHosts` feature — emitting panel-defined routing into the subscription rather than relying on the client app to choose.

Refs: #5825

## Design notes

- **Observatory config is panel-wide**, stored as `subJsonObservatory` (JSON string), consistent with the existing `subJsonMux` / `subJsonRules` / `subJsonFinalMask` settings. Per-balancer probe URLs are a rare need and not worth the model/form bloat.
- **Conditional emit**: `burstObservatory` is emitted only for `leastPing` / `leastLoad`. `random` / `roundRobin` have no fallback in the client-side model, so an observatory would only make the client probe every outbound for nothing — these strategies emit no observatory (per Xray routing docs).
- **Balancers vs inbounds**: member inbounds are selected by id; an admin can pick inbounds hosted on another node as long as they are in the subscriber's subscription. The sub server reads `sub_balancers` rows per request from its own DB.
- `SubBalancer` rows are read per request by the subscription builder, so balancer mutations (create/update/delete) are incremental — no Xray restart, no panel restart needed to apply them. The Observatory sub-tab edits the panel-wide `subJsonObservatory` setting, which is persisted via the regular Settings **Save** button.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Xray config generation
- [x] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot

## How was this tested?

**New unit tests**
- `internal/sub/sub_balancer_test.go` — emitted balancer document shape (member retagging under `bal-<id>-…`, `routing.balancers` selector, `balancerTag` rules, `burstObservatory` for `leastLoad`), ordering/interleaving with inbounds by sort order, disabled balancers and balancers with no matching members skipped, manual inbound docs keep their plain `proxy` routing (shared routing tree is cloned per doc), `burstObservatory` emitted only for `leastPing`/`leastLoad` (not `random`/`roundRobin`) with the panel-wide probe config overriding defaults.
- `internal/web/service/sub_balancer_test.go` — CRUD + validation (remark required, valid strategy, ≥1 inbound, sort order ≥1), `roundRobin` accepted.
- `frontend/src/test/sub-balancer-form-modal.test.tsx` — form modal validation + submit payload.

**Ran locally**
- `go test ./internal/sub/... ./internal/web/...` — green.
- `make gen` — `frontend/src/generated/*` regenerated to include the new `subJsonObservatory` field and the `SubBalancer` schema; `frontend/public/openapi.json` updated; re-running `make gen` produces no further diff (gen-check clean).
- `make typecheck` (`tsc --noEmit`), `make lint-fe` (eslint), `npm test` (vitest incl. sub-balancer / observatory / i18n-dead-keys) — green.
- `make verify` (gen-check + lint + typecheck + test + build + build-storybook) — passes.

**Manual (UI):** Settings → Sub Balancers (appears when the JSON subscription is enabled): add/edit/delete a balancer, toggle enabled, reorder via sort order; inbounds from other nodes that are part of the subscription can be selected as members; Observatory sub-tab toggles probe config on/off and edits destination/connectivity/interval/timeout/sampling/httpMethod, persisted via the Settings **Save** button. A client on ≥1 selected inbound receives one extra JSON-sub document with `routing.balancers` + (for `leastPing`/`leastLoad`) `burstObservatory`.

## Screenshots / recordings

image
image
image

## Breaking changes

None. New `sub_balancers` table created via AutoMigrate on startup (additive); also registered in `migrationModels()` so SQLite→Postgres cross-DB migration (`x-ui migrate-db`) copies it. New `/panel/api/sub-balancers` endpoints are additive. New `subJsonObservatory` panel setting defaults to empty → safe built-in probe defaults, so the sub server behaves as before until an admin enables it. No existing API, config field, or subscription output changes for users who don't create a balancer.

## Checklist


DIMFLIX:
Addressed in 6 commits (`da48833f`…`4cf4ae2e`). One important correction on №3 after real-Xray testing.

**Observatory surface (№1, №2, №3, №8, №13, №15):**
- `connectivity` now defaults to `""` (skip the direct pre-check) and an explicit `{"connectivity":""}` stays empty — fixes the unreachable "Leave empty to skip" and the dead-on-arrival google default on censored client networks.
- `routing.balancers` carries `fallbackTag` = first member for `leastPing`/`leastLoad`, so an all-probes-fail case degrades to the first member instead of failing dispatch.
- Balancer entries (`kind!=0`) are skipped in the member scan.
- **On №3, I iterated after a runtime check.** I first implemented your option 1 (suppress `burstObservatory` when the toggle is off), but testing against Xray 26.7 showed it refuses to start `leastPing`/`leastLoad` without a `burstObservatory` — the config fails to load on startup. So I went with your option 2: the observatory is always emitted for these strategies (it's mandatory), and the switch means "customise probe parameters vs built-in defaults" (`''` keeps the defaults, a stored JSON overrides them). An info Alert explains this + the panel-restart requirement. I clarified via the note/description rather than renaming the `title` label — happy to rename the title to "Customise observatory probe parameters" in all 13 locales if you'd prefer the literal relabel.

**Controller (№5, №7):** toasts via `I18nWeb` (new keys in all 13 locales); `parseID` → `strconv.Atoi`, reject ids `< 1`.
**Remark cap (№6):** `len > 256` rejected server-side in `validate()`.
**mtproto picker (№9):** dropped — no proxy-outbound case for it.
**Docs (№4, №10, №12a):** `subscription-balancers` added to `en` meta.json (fa/ru/zh have no MDX — `gen-openapi.ts` emits `en` only); tab pointer fixed to "Settings → Sub Balancers"; mirror scope noted; OpenAPI + MDX regenerated.
**Comment 2-line cap (№11):** trimmed.

**Intentionally left:**
- **№14 (route shape / dead `DELETE` alias / discarded return):** the sub-balancer controller mirrors the sibling outbound-subscription controller in `xray_setting.go` (same `fmt.Sscanf`, `enabled != "false"`, `DELETE` + `POST /:id/del`). Realigning only this controller to `host.go` would diverge from that sibling. Happy to drop the unused `DELETE` route + its endpoints.ts entry as a follow-up if you want the `host.go` convention applied to both.
- **№12b (add `model.SubBalancer` to openapigen `StructAllow`):** the `endpoints.ts` system carries no response-schema reference, so this would emit an orphan schema. Typing the responses needs codegen changes beyond this review.

`make verify` passes locally (gen-check green with the regenerated spec committed). Each fix has a test that fails without it; the observatory-always-emitted test was added after the runtime check caught that suppression breaks startup.

yakazahvahvah-dotcom:
Hello, can i ask you to also add weighted least load. If it is not added yet. 
For example we can have one Nod with 1gbs speed and another with 100 mbs speed (least load will separate users by count)
So we should add a weight to each Nod to solve that problem

DIMFLIX:
> Hello, can i ask you to also add weighted least load. If it is not added yet. For example we can have one Nod with 1gbs speed and another with 100 mbs speed (least load will separate users by count) So we should add a weight to each Nod to solve that problem

Hello. The weighing feature has not been added yet. 
This is feasible—but as a follow-up, not in this PR. This PR already has 35 commits, and the review items are still open (check in Observatory, i18n, several controller fixes), and adding a weight field again affects the model/form/JSON generation, so I’d prefer to carefully finish this PR first, and then open a separate PR for weighted leastLoad. I’ll post a link here as soon as it’s available.

MHSanaei:
@claude resolve pr conflicts

DIMFLIX:
> @claude resolve pr conflicts

I'll fix it right now. I'm actually working on the issues he pointed out in his code review.

MHSanaei:
@DIMFLIX thanks but i tried to check the claude i change something in it and now it doesn't work anymore :D

MHSanaei:
@claude resolve pr conflicts

DIMFLIX:
@MHSanaei I've addressed all the issues raised in the bot review. What do you think? Is everything looking good now?

One small request: would it be possible to merge this PR without squashing? I tried to keep the commits organized and meaningful, and I'd really appreciate preserving them for the contribution graph. Either "Create a merge commit" or "Rebase and merge" would work perfectly.

This will also make it easier to build on top of this feature in the future (like the weighted least load suggestion from @yakazahvahvah-dotcom).

Thanks! 🙏

DIMFLIX:
@MHSanaei Hi! Just a gentle bump in case this got buried. Let me know if there’s anything else needed from my side. 🙏

DIMFLIX:
@yakazahvahvah-dotcom I have implemented the feature you requested #6304

PR #6250, opened 2026-08-18
fix: follow-ups from the post-merge reviews of #6221, #6227, #6230 and #6239

n0ctal:
Follow-ups from the automated reviews that landed on #6221, #6227, #6230 and #6239 after they merged. One PR rather than six, since they are all small and all in the same batch of work.

I checked each finding before acting on it rather than taking the review at its word — two of them by experiment, and one turned out not to be real.

## The two I would look at first

**Restoring a backup onto a reinstalled panel destroyed the node mTLS material.** The "no local row means the built-in default applied, so drop the imported row" branch also covers the five `nodeMtls*` keys. Those are minted on demand and deliberately excluded from `AllSetting`, so a fresh install has no row for them — and the branch deleted the imported CA certificate together with its private key. Neither is surfaced in the UI or in the export, so the backup file was the only copy. Rebuild the box, reinstall, restore yesterday's dump, and the master mints a fresh CA while every node still trusts the old one.

**`Assign(model.Setting{Value: ...})` never wrote an empty value.** GORM builds the assignment map from the struct and drops zero-valued fields, so the found-branch ran `Updates` with nothing to set. Empty is not an edge case here: `UpdateAllSetting` writes a row for every field including the blank ones, so any panel where Settings has ever been saved carries `webCertFile=""` and friends. The result is precisely the scenario #6227 was written for — a destination with no certificate keeps the source machine's path and still fails to come up on HTTPS. I confirmed it with a throwaway test before touching anything: the imported value survived untouched.

## The rest

**One click on the clients-list enable switch wiped renewal state.** `setEnable` hand-builds the update payload; it carried `reset` but not `resetDay` or `resetMax`, so toggling a client turned calendar mode off and lifted the renewal cap for good. My tests could not catch this because they drive the form modal, which does send both.

**"Delete depleted clients" deleted calendar clients.** The predicate treated `reset = 0` as "does not auto-renew", which is exactly the calendar shape — so a calendar client over quota was depleted at all times, while an interval client was protected by the same clause. It appeared in two places; both now use one constant that requires `reset_day = 0` as well.

**Allowlist validation and parsing accepted different things.** Save used `net`, scan used `netip`. `198.51.100.0/024` passes `net.ParseCIDR` and is rejected by `netip.ParsePrefix`, so it saved without complaint and was silently dropped at scan — the exact failure #6230 claimed to remove. I ran both parsers to confirm rather than reason about it. Separately, `::ffff:198.51.100.0/120` parsed but could never match, because `contains` unmaps the queried address while the stored prefix stayed 128 bits; it is unmapped at parse now. The new test asserts the two acceptance sets agree, so they cannot drift apart again.

**A comment said the opposite of the truth.** #6221's guard is justified with "disabled inbounds are intentionally absent from the node's runtime snapshot". They are not — `GetInbounds` has no enable filter and a node reports a disabled inbound normally. The row in that report was missing only because it had never been delivered. That was my wording and my mistake; it is reworded to the real invariant.

Plus two comment blocks trimmed to the two-line maximum.

## Deliberately not included

The reviewer's suggestion to lift the node hand-off out of `if inbound.Enable` in `AddInbound` is the right root-cause fix — every other mutation path marks the node dirty unconditionally — but it changes delivery behaviour on multi-node deployments and deserves its own change with its own testing rather than riding along in a cleanup batch.

## One reported finding is not real

The review on #6239 says `BulkCreate` never calls `validateClientResetDay` and that `validateClientResetMax` has the same gap. In the merged tree it calls `validateClientEmail`, `validateClientSubID`, `validateClientResetDay`, `validateClientResetMax` and `validateClientTrafficReset`. Nothing to do there.

## Validation

Four new tests, each mutation-checked against the defect it covers: reverting the `saveSetting` change fails with the source machine's certificate path still in place; restoring the delete branch fails with the CA certificate gone; putting the validator back on `net` fails on `/024`; removing the unmap step fails with the IPv4-mapped entry matching nothing. The existing `TestImportDropsHostBoundSettingsThisMachineNeverHad` was narrowed, since it had pinned the behaviour now identified as the defect.

`go build ./...`, `go vet`, `golangci-lint` clean, full Go suite green, frontend typecheck, eslint and 1159 unit tests green.


n0ctal:
The failing `golangci` check was not from this PR, and I want to be explicit about what I did with it rather than have a fix appear quietly in a file this PR has no business touching.

`internal/util/netsafe/netsafe.go:67` has `fmt.Errorf("%w; %v", blockedErr, lastErr)` — the first error is wrapped, the second only formatted, and errorlint rejects that. It arrived with #6242 and `main` has been red on every push since; the check also failed on #6242's own PR before it merged. Nothing in my diff goes near that file.

I have included the one-line fix anyway, as a separate commit, because CI lints the merged result and this PR cannot be read while the check is red. Go 1.20 allows more than one `%w`, so both errors are wrapped and `errors.Is` now matches either — which is what the `; ` join was reaching for. Revert that commit freely if you would rather fix it on `main` yourself; the rest of the PR does not depend on it.

Two smaller notes while I am here:

I merged current `main` into the branch rather than rebasing, so the earlier review comments stay anchored to the lines they were written against. Happy to rebase if you prefer a linear history.

And #6222 landed while I was working, so all eleven of the earlier batch are in. Thank you for the `.gitignore` change — making the dist stub un-committable is a better fix than my repeated attempts to remember not to stage it, and for the `importKeepHostSettings` translations I left in English.

PR #6259, opened 2026-08-18
fix(panel): forward the panel's proxy to update.sh's own downloads

kuzzrus:
## Summary

`startUpdate` already fetches `update.sh` itself through a proxy-aware HTTP client (`NewProxiedHTTPClient`), but the process that actually executes `update.sh` never receives a proxy hint of its own — so `update.sh`'s own `curl` calls to GitHub always go out directly, even when the panel has a working proxy path configured.

## Why

This is most visible on the `systemd-run` launch path: `systemd-run` does not inherit the caller's environment at all (only `--setenv` passes through), so a host that needs a proxy to reach GitHub loses it for this one hop even though every other panel-outbound request (the version check, the `update.sh` download itself) already goes through it correctly.

`curl` already honors `https_proxy`/`all_proxy` natively, so no changes to `update.sh` are needed — only the launcher needs to forward a proxy URL into the environment it hands to the detached process.

`updateProxyEnvVars()` prefers an already-set ambient proxy env var (never silently overriding an admin's own proxy config) and only falls back to the panel's own configured panel outbound (`PanelEgressProxyURL`) when nothing is set, then forwards the result to both launch paths (`--setenv` for `systemd-run`, `cmd.Env` for the `bash -lc` fallback).

## Type of change

- [x] Bug fix

## Areas affected

## How was this tested?

`internal/web/service/panel` transitively imports `internal/database` (CGO/sqlite3), which I wasn't able to compile in my local environment. I added `TestUpdateProxyEnvVarsPrefersAmbientProxy`, which exercises the ambient-proxy branch directly (it never touches `PanelEgressProxyURL`/the database, so it needs no DB test scaffolding), and manually checked the new call against `NewProxiedHTTPClient`'s existing call pattern already used elsewhere in this same file. I was not able to run `go build`/the full test suite myself — would appreciate CI or a maintainer confirming those, and ideally a real update run on a host that actually needs a proxy to reach GitHub.

## Screenshots / recordings

N/A — backend-only change.

## Breaking changes

None.

## Checklist

- [ ] I tested the change locally and confirmed the described behavior. (blocked by the local CGO/sqlite3 gap above, not skipped by choice)
- [ ] `go build ./...` and the test suite pass locally. (same CGO blocker; should be green in CI)
- [x] For frontend changes: N/A, backend-only.
- [x] I updated the Wiki / README / API docs if user-facing behavior changed. (no user-facing behavior changed)


kuzzrus:
Thanks for the thorough review — the Critical finding is correct and I've pushed a fix.

**Critical (PanelEgressProxyURL fallback)**: Confirmed independently against update.sh:1019-1199 and x-ui.service.debian (no `KillMode` override) — `systemctl stop x-ui` does take the Xray child down with it, and `update.sh` still needs `curl` afterwards for `x-ui.sh` and sometimes the service unit. Dropped the fallback entirely; `updateProxyEnvVars` now only forwards ambient proxy env vars, which survive the panel dying since they live at the OS level, not inside the process `update.sh` itself kills. (02cc009c)

**Medium — no_proxy/NO_PROXY**: now forwarded alongside the proxy vars.

**Medium — comment length**: trimmed to 2 lines per CLAUDE.md.

**Low — http_proxy promoted to https_proxy**: fixed — each ambient var is now forwarded under its own name, nothing is collapsed or promoted.

**Low — credentials in systemd-run argv**: not changed. Given this only forwards a value already sitting in the admin's own ambient environment (never a value newly derived by this code), and the recommended mitigation (a mode-0600 environment file) is a materially larger change than this fix's actual bug, I left it as a known, minor trade-off rather than expanding scope — happy to revisit if it matters for real deployments.

**Testing**: rewrote as table-driven subtests (`nothing set → nil`, `forwards each set var under its own name`, `http_proxy is not promoted`), matching `TestShellQuote`'s style.

PR #6263, opened 2026-08-19
fix(traffic): clear cross-panel rows only for clients actually renewed

n0ctal:
## Summary

`autoRenewClients` deletes a client's `client_global_traffics` rows when it rolls the client into a fresh window. It was passing the full *candidate* list to `clearGlobalTraffic`, not the clients it actually renewed.

## Why this matters

The function collects every expired client that carries a reset interval, then skips three kinds of candidate before touching their counters:

- the email is not present in the inbound's settings,
- the catch-up resolves to no whole interval (`renewals == 0`),
- `newExpiryTime` is still in the past once `ResetMax` truncates the catch-up.

All three deliberately keep both their counters and their expiry — the third is exactly what `TestAutoRenewClients_TruncatedCatchUpLeavesTheClientDisabled` pins down.

Their cross-panel rows were dropped anyway, so the pushed totals and the local counters describe different windows. The damage is bounded and I want to state it accurately rather than oversell it: `PushGlobalClientTraffics` upserts, so the next push recreates the row, and `depletedClientsCond` trips on its `expiry_time <= ?` branch for all three cases, so `disableInvalidClients` still cuts these clients in the same tick — nothing is served past its limit. What remains is a cross-panel under-report between the delete and the next push, repeated on every poll for as long as the client stays a candidate, because its expiry never advances.

## Scope

One argument in `internal/web/service/inbound_traffic.go`: collect the emails whose `Up`/`Down` this pass zeroed, and hand those to `clearGlobalTraffic`. The renewed count returned to the caller now counts the same set — it was returning the candidate count, which the caller logs as `"%v clients renewed"`.

No change to when a client is renewed, to the reset cap, or to the delete itself.

`clearGlobalTraffic` already early-returns on an empty slice, so a poll that renews nobody remains a no-op instead of deleting every row — worth stating because the new list can be empty where the old one never was.

## Validation

`go build ./...`, `go vet ./internal/...` and the full `internal/web/service` suite pass.

Two new tests, both driven through `autoRenewClients` itself:

- `TestAutoRenewClients_TruncatedCatchUpKeepsCrossPanelTraffic` — a capped catch-up keeps its rows. Fails on the previous behaviour with `cross-panel rows=0, want 1`; it also asserts the local counters survived, so the test cannot pass by accident if that scenario ever stops exercising the cap.
- `TestAutoRenewClients_RenewedClientLosesCrossPanelTraffic` — a genuinely renewed client still loses its rows. This one passes both before and after by design: it is the guard against "fixing" the report by dropping the delete altogether.

## Risk

Low. The only behavioural difference is that clients which were not renewed keep rows that were previously deleted, and the debug count now matches what the function did. Renewed clients are unaffected, so the reason the delete exists — stale pushed totals re-depleting a fresh window — still holds.

### DIFF
--- a/internal/web/service/inbound_autorenew_globaltraffic_test.go
+++ b/internal/web/service/inbound_autorenew_globaltraffic_test.go
@@ -1,0 +1,107 @@
+package service
+
+import (
+	"testing"
+	"time"
+
+	"github.com/mhsanaei/3x-ui/v3/internal/database"
+	"github.com/mhsanaei/3x-ui/v3/internal/database/model"
+	"github.com/mhsanaei/3x-ui/v3/internal/xray"
+)
+
+func seedGlobalTraffic(t *testing.T, email string, up, down int64) {
+	t.Helper()
+	if err := database.GetDB().Create(&model.ClientGlobalTraffic{
+		MasterGuid: "peer-" + email, Email: email, Up: up, Down: down,
+	}).Error; err != nil {
+		t.Fatalf("seed client_global_traffics: %v", err)
+	}
+}
+
+func countGlobalTraffic(t *testing.T, email string) int64 {
+	t.Helper()
+	var n int64
+	if err := database.GetDB().Model(&model.ClientGlobalTraffic{}).
+		Where("email = ?", email).Count(&n).Error; err != nil {
+		t.Fatalf("count client_global_traffics: %v", err)
+	}
+	return n
+}
+
+// A capped catch-up keeps its counters and stays expired, so its cross-panel
+// rows still describe a window this client has not spent.
+func TestAutoRenewClients_TruncatedCatchUpKeepsCrossPanelTraffic(t *testing.T) {
+	setupBulkDB(t)
+	svc := &InboundService{}
+	db := database.GetDB()
+
+	past := time.Now().Add(-150 * 24 * time.Hour).UnixMilli()
+	clients := []model.Client{
+		{Email: "capped@x", ID: "66666666-6666-6666-6666-666666666666", Enable: false, Reset: 30, ResetMax: 3, ExpiryTime: past},
+	}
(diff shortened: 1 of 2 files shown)


n0ctal:
Thanks — the calibration is right on both counts, and I checked each before changing anything.

`depletedClientsCond` does carry an `expiry_time > 0 AND expiry_time <= ?` branch, and all three skipped cases leave the expiry in the past, so `disableInvalidClients` cuts them in the same tick whatever the cross-panel rows say. And `PushGlobalClientTraffics` upserts with `DoUpdates`, so a deleted row comes back on the next push. My description claimed a client past its limit keeps being served; that was wrong, and I have rewritten the body to describe what actually remains — a cross-panel under-report between the delete and the next push, repeated every poll while the expiry never advances.

Addressed in the amended head:

- **Renewed count.** `autoRenewClients` now returns `int64(len(renewedEmails))`. You are right that the log disagreed with the function one line after the fix.
- **Comment length.** Both blocks are down to two lines, on the invariant rather than a retelling of the code. My mistake — the rule is in CLAUDE.md and I should have applied it before opening this.

Left as they are:

- **The multi-inbound bug.** Agreed it is out of scope here, and I will not touch it in this PR. I had found the same thing independently while preparing this change and confirmed it with a throwaway test: with one email on two local inbounds, inbound 1 ends up `enable:true` with a future expiry, inbound 2 keeps `enable:false` with the past one, and the `clients` row takes the stale copy — after which no later tick revisits the client, because its traffic row is no longer expired. I will open a separate issue with that reproduction rather than fold it in here. Your point that this is also why `renewedEmails` cannot contain duplicates today is a good catch; a fix for the multi-inbound case would need to add the dedupe.
- **The duplicated fixture.** Kept separate deliberately, for the reason you name: the failure message should say the cross-panel rows are gone rather than surfacing as a cap assertion.

PR #6268, opened 2026-08-19
fix: reject Hysteria inbound updates with empty client auth

### LINKED ISSUE #6232: [Bug]: Hysteria2 clients whose auth value is empty in the database fail authentication
## Bug Description

When creating or editing a Hysteria2 client in the 3X-UI panel, the password is stored in the `password` field in the SQLite database. However, xray-core's Hysteria2 implementation only reads the `auth` field for authentication. This results in the `auth` field being empty, causing `404 authentication failed` errors when clients try to connect.

## Environment

- **3X-UI version**: v3.6.0
- **Protocol**: Hysteria2 (Inbound)
- **OS**: Ubuntu 24.04.4 LTS (x86_64)
- **xray-core version**: (bundled with 3X-UI v3.6.0)

## Steps to Reproduce

1. Install 3X-UI v3.6.0
2. Create a new Hysteria2 inbound (any port, UDP)
3. Add a client with a password via the panel
4. Attempt to connect using a Hysteria2 client (e.g., Clash Verge Rev / mihomo) with the password
5. Connection fails with: `authentication failed` / `404`

## Root Cause

The panel stores the password in the `password` column of the `clients` table in `x-ui.db`, but leaves the `auth` column empty.

xray-core's Hysteria2 implementation reads the `auth` field for authentication, not `password`. Since `auth` is empty, all authentication attempts fail.

### Database Evidence

```sql
sqlite3 /etc/x-ui/x-ui.db "SELECT id, email, password, auth FROM clients WHERE protocol IS NULL OR protocol='';"
```

Result shows `password` has the value but `auth` is empty for all Hysteria2 clients.

## Expected Behavior

When a Hysteria2 client's password is set or changed in the panel, the `auth` field should be automatically populated with the same value.

## Current Workaround

Manual SQL fix:
```sql
UPDATE clients SET auth = password WHERE auth = '' AND password != '';
```

This can be run as a cron job every minute to keep `auth` in sync:

```python
#!/usr/bin/env python3
# /usr/local/bin/fix-hy2-auth.py
import sqlite3
db = sqlite3.connect('/etc/x-ui/x-ui.db')
db.execute("UPDATE clients SET auth=password WHERE auth='' AND password!=''")
db.commit()
db.close()
```

```
# /etc/cron.d/fix-hy2-auth
* * * * * root /usr/bin/python3 /usr/local/bin/fix-hy2-auth.py
```

## Additional Notes

- This affects all Hysteria2 clients created or edited through the panel
- The `fix-hy2-auth.py` cron workaround has been running stably, but a proper fix in the panel code would be much better
- The bug may also affect the `subId` / subscription link generation if the `auth` field is used there

mvanhorn:
## Summary

Parse the clients from the pending inbound settings in `UpdateInbound` and apply the existing Hysteria credential invariant before entering the persistence transaction, mirroring the rejection behavior already enforced by `AddInbound`. Return the established `empty client ID` validation error when any Hysteria client has an empty `auth`, so an API-driven protocol change or edit cannot reach `SyncInbound` or the runtime with unusable credentials. Keep valid Hysteria auth values unchanged and do not synthesize them from `password`, mutate unrelated credentials, or add a database migration.

Hysteria clients authenticate with their dedicated `auth` credential, but `UpdateInbound` can currently persist a Hysteria settings payload whose client has an empty `auth`. That row is later rendered into the Xray configuration without a usable credential, producing authentication failures. Creation and client-specific write paths already prevent or fill this state, leaving the whole-inbound update path as the concrete gap. The fix must preserve the intentional separation between Hysteria `auth` and the Trojan/Shadowsocks `password` field rather than copying one into the other.

Closes #6232

## Why

Parse the clients from the pending inbound settings in `UpdateInbound` and apply the existing Hysteria credential invariant before entering the persistence transaction, mirroring the rejection behavior already enforced by `AddInbound`. Return the established `empty client ID` validation error when any Hysteria client has an empty `auth`, so an API-driven protocol change or edit cannot reach `SyncInbound` or the runtime with unusable credentials. Keep valid Hysteria auth values unchanged and do not synthesize them from `password`, mutate unrelated credentials, or add a database migration.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot

## How was this tested?

Not applicable to this change.

## Screenshots / recordings

No user-visible surface changes in this PR, so there is nothing to show.

## Breaking changes

None. This change is additive and does not alter an existing public interface or stored format.

## Checklist

- [ ] I have no unrelated changes mixed into this PR.

**AI disclosure**

AI was used for assistance.

- Tools: an AI coding assistant.

### DIFF
--- a/internal/web/service/inbound.go
+++ b/internal/web/service/inbound.go
@@ -1410,6 +1410,18 @@ func (s *InboundService) UpdateInbound(inbound *model.Inbound) (*model.Inbound, 
 	s.normalizeMtprotoSecret(inbound)
 	inbound.SubSortIndex = normalizeSubSortIndex(inbound.SubSortIndex)
 
+	clients, err := s.GetClients(inbound)
+	if err != nil {
+		return inbound, false, err
+	}
+	if inbound.Protocol == model.Hysteria {
+		for _, client := range clients {
+			if client.Auth == "" {
+				return inbound, false, common.NewError("empty client ID")
+			}
+		}
+	}
+
 	oldInbound, err := s.GetInbound(inbound.Id)
 	if err != nil {
 		return inbound, false, err
--- a/internal/web/service/inbound_hysteria_auth_test.go
+++ b/internal/web/service/inbound_hysteria_auth_test.go
@@ -1,0 +1,106 @@
+package service
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/mhsanaei/3x-ui/v3/internal/database"
+	"github.com/mhsanaei/3x-ui/v3/internal/database/model"
+)
+
+func TestUpdateInbound_RejectsHysteriaClientWithoutAuth(t *testing.T) {
+	setupConflictDB(t)
+	seedInboundConflict(t, "in-45001-tcp", "0.0.0.0", 45001, model.VLESS,
+		`{"network":"tcp"}`, `{"clients":[]}`)
+
+	var existing model.Inbound
+	if err := database.GetDB().Where("tag = ?", "in-45001-tcp").First(&existing).Error; err != nil {
+		t.Fatalf("read seeded row: %v", err)
+	}
+
+	update := existing
+	update.Protocol = model.Hysteria
+	update.Settings = `{"clients":[{"email":"hysteria@x","enable":true,"password":"not-hysteria-auth"}]}`
+
+	svc := &InboundService{}
+	if _, _, err := svc.UpdateInbound(&update); err == nil || !strings.Contains(err.Error(), "empty client ID") {
+		t.Fatalf("UpdateInbound error = %v, want empty client ID", err)
+	}
+
+	var reloaded model.Inbound
+	if err := database.GetDB().First(&reloaded, existing.Id).Error; err != nil {
+		t.Fatalf("reload: %v", err)
+	}
+	if reloaded.Protocol != existing.Protocol {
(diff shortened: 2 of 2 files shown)


MHSanaei:
@claude review

mvanhorn:
@MHSanaei rebased onto current main and reduced this to the one commit it should have been: 45a46d87.

The other five were CI work that has since landed on main independently, so they are dropped rather than replayed. One of them would have reverted your hardening of `claude-bot.yml`, which I did not want to do.

What is left is the Hysteria guard in `UpdateInbound`. `AddInbound` already rejects a client with empty `Auth`; `UpdateInbound` did not, so an existing Hysteria inbound could be updated to have a client with no auth. The new check mirrors the `AddInbound` pattern and returns the same `empty client ID` error.

Tests: `TestUpdateInbound_RejectsHysteriaClientWithoutAuth`, `TestUpdateInbound_PreservesHysteriaClientAuth`, `TestUpdateInbound_AllowsHysteriaWithoutClients`, all passing locally.

PR #6272, opened 2026-08-20
feat(pia): add PIA login-and-add WireGuard outbounds

Masterain98:
## Summary

Add a Private Internet Access (PIA) WireGuard integration on **Xray → Outbounds → More → PIA**: sign in with a PIA username and password, pick a country/region/server from the signature-verified server list, and append a WireGuard outbound to the Xray template. Operators can add several servers (one outbound per hostname) and Reset a row in place.

The stored PIA API token follows the panel’s existing `NODE_TOKEN_ENCRYPTION` setting (`off` / `migration` / `required`). The password is never stored.

## Why

WARP and Nord already have a panel modal that provisions an outbound. PIA did not: operators had to mint a WireGuard peer by hand. The frontend cannot invent that peer — PIA only issues one through per-server `/addKey` after a username/password login — so the panel needs a protocol client plus a Nord-shaped UI.

Node API tokens already encrypt at rest when `NODE_TOKEN_ENCRYPTION` is on. PIA’s token is the same class of long-lived panel secret; it now uses that codec with AAD `settings/pia_token` so ciphertext cannot be swapped with a node-token blob. Default `off` keeps plaintext JSON, matching Nord. WireGuard `secretKey` in the Xray template stays plaintext, same as Nord/WARP.

## Type of change

- [ ] Bug fix
- [x] New feature
- [x] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [x] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot

## How was this tested?

**UI (manual):** Xray → Outbounds → More → PIA

1. Sign in with a PIA username and password. The password field clears; the account hint is shown, not the token.
2. Pick a country (and optional region), pick a server, **Add outbound**. A `pia-<region>-<server>` WireGuard outbound appears in the list and in the “Added servers” table. The modal stays open.
3. Add a second, different hostname. Both outbounds remain.
4. Select a hostname that is already added: **Add outbound** is disabled and the `{reset}` hint is shown.
5. **Reset** on a row keeps the same tag and replaces the WireGuard keys in place.
6. **Logout** clears the stored token only; existing `pia-*` outbounds stay. Delete unused ones from the Outbounds tab so routing cleanup still runs.
7. With `NODE_TOKEN_ENCRYPTION=off` (default): the `pia` setting JSON still contains the raw token.
8. With `NODE_TOKEN_ENCRYPTION=required` (and a key): after login the `pia` setting token is `enc:v1:…`; Add outbound still works. Turning encryption off afterwards makes Add/Reset fail closed until the operator signs in again.

**API:** `POST /panel/api/xray/pia/:action`

| action | body | result |
| --- | --- | --- |
| `reg` | `username`, `password` | `{ username, accountHint }` — token never returned |
| `data` | — | same account view, or empty when logged out |
| `countries` | — | signed server-list country codes |
| `servers` | `countryCode` | regions + WireGuard servers |
| `addKey` | `hostname` | tag, hostname, secretKey, address, publicKey, endpoint |
| `del` | — | clears the stored token only |

**Automated (local):**

- `go test ./internal/pia/ ./internal/crypto/nodetoken/ ./internal/web/service/integration/ ./internal/xray/`
- Protocol coverage in `internal/pia`: auth, RSA signature on the server list, `/addKey` TLS + JSON fixtures
- Panel coverage in `internal/web/service/integration`: login hides the token from `data`; ModeOff plaintext; ModeRequired `enc:v1` at rest; AddKey decrypts; leftover ciphertext is rejected when encryption is off; plaintext migrates on the next read after encryption is enabled; AAD `settings/pia_token` does not decrypt as `nodes/api_token/{id}`
- `npx vitest run src/test/pia-modal.test.tsx src/test/i18n-dead-keys.test.ts src/test/outbounds-loopback-index.test.tsx`
- `golangci-lint run ./internal/pia/... ./internal/crypto/nodetoken/... ./internal/web/service/integration/... ./internal/xray/`

CI should still run the full `make verify` gate.

## Screenshots / recordings

image

image

image

image

For duplicated outbounds:
image

## Breaking changes

None for existing installs. Additive only: `POST /panel/api/xray/pia/:action` and a `pia` settings key. No schema migration. Logout does not delete outbounds.

`NODE_TOKEN_ENCRYPTION` is unchanged as an env var. When it is `off` (default), PIA storage matches Nord (plaintext JSON). When it is `migration`/`required`, only the PIA API token in that setting is encrypted; WireGuard keys in the Xray template are not.

## Checklist


Masterain98:
# CI `golangci` job failure — upstream regression in golangci-lint v2.13.0 (not caused by this PR)

The trigger is the unpinned version at `.github/workflows/ci.yml#L170-L173`:

```yaml
# .github/workflows/ci.yml lines 170-173
      - name: golangci-lint
        uses: golangci/golangci-lint-action@v9
        with:
          version: latest
```

Line 173 (`version: latest`) now resolves to golangci-lint v2.13.0, which bundles a broken staticcheck pre-release.

## Symptom

The `golangci` job fails with the following error, while all other jobs (`go-test`, `race`, `govulncheck`, `codegen`, `frontend`, etc.) pass:

```
level=error msg="[linters_context/goanalysis] nilness: panic during analysis:
internal error: unhandled builtin recover, goroutine ..."
level=error msg="Running error: can't run linter goanalysis_metalinter:
goanalysis_metalinter: nilness: package \"common\" (isInitialPkg: false,
needAnalyzeSource: true): internal error: unhandled builtin recover"
##[error]golangci-lint exit with code 3
```

## Root cause

This failure is **not caused by the code in this PR**. It is a regression in the recently released **golangci-lint v2.13.0** (released 2026-08-19), which bundles **staticcheck v0.8.0-rc.1** (a pre-release). The `nilness` fact analyzer in this staticcheck version panics when analyzing packages that call the `recover()` builtin — such as `internal/util/common`, where `common.Recover()` and `common.GoRecover()` are defined.

The evidence:

- The CI workflow uses `version: latest` for `golangci/golangci-lint-action@v9` (line 173), so the "latest" resolution jumped from v2.12.2 to v2.13.0 when the new release came out.
- The last successful run on `main` (2026-08-19 17:55 UTC) resolved to **v2.12.2** and passed; every run since 2026-08-20 03:20 UTC resolves to **v2.13.0** and fails with the same panic — with **zero relevant code changes** in between.
- The same failure also occurs on the `main` branch, confirming it is not PR-specific.

## Fix

Pin golangci-lint to the last known-good version in `.github/workflows/ci.yml`:

```yaml
      - name: golangci-lint
        uses: golangci/golangci-lint-action@v9
        with:
          # Pin: v2.13.0 bundles staticcheck v0.8.0-rc.1 whose nilness fact
          # analyzer panics ("internal error: unhandled builtin recover") on
          # packages containing recover(), e.g. internal/util/common.
          # Revert to `latest` once upstream ships a fixed release.
          version: v2.12.2
```

## Verification

Running golangci-lint **v2.12.2** locally against the full repository (including all code added in this PR) reports **0 issues** — the panic does not occur and no new lint problems are introduced.

Once upstream (golangci-lint or staticcheck) ships a release with the `nilness` panic fixed, the pin can be reverted to `latest`.

MHSanaei:
i know about ci, and sorry for delete comments i working on claude agent to make it better

Masterain98:
> i know about ci, and sorry for delete comments i working on claude agent to make it better

No worry. I was just checking if the CI issue was caused by the PR. Hope it helps. 

Lmk if you need any changes. I will also align with the AI review result.

MHSanaei:
@claude review

Issue #5785, opened 2026-07-04
[Bug]: Duplication of WireGuard client IP address when using multiple inbounds


AndreyOkh:
### Describe the bug

If a client is added to multiple WireGuard inbound configurations, or if one WireGuard configuration is swapped for another, different clients might end up with the same IP address. This can lead to traffic routing issues.

### How to reproduce the problem

1. Create inbound WireGuard configurations "a" and "b".
2. Create "user1" and link it to inbound "a".
3. Create "user2" and link it to inbound "b".
4. Change the inbound setting for "user1" from "a" to "b".
5. Both clients will have the same IP address and will not be able to operate simultaneously.

### Expected behavior

A unique IP address must be assigned to each individual WireGuard inbound connection.

### Relevant logs

```shell

```

### 3x-ui version

3.4.2

### Xray-core version

26.6.27

### How did you install 3x-ui?

install.sh script

### Operating system

Debian 12

### Is the panel behind a reverse proxy or CDN?

None

### Before submitting

- [x] I searched [existing issues](../issues?q=is%3Aissue) and this bug has not been reported.
- [x] I am running the latest released version of 3x-ui (or have verified the bug still exists on it).
- [x] This bug report is written in English.
- [x] I have redacted any sensitive data (tokens, real domains, client UUIDs).

MHSanaei:
This is fixed on `main` by 64c30603 (`feat(wireguard): make client allowedIPs editable with validation`), which landed after v3.4.2 was tagged — so it will ship in the next release.

What changed:

- Attaching or moving a client to a WireGuard inbound where its stored address is already held by another peer is now **rejected with an explicit error** (`wireguard: allowedIPs entry already used by another client: <addr>`) instead of silently writing a duplicate peer address. The move aborts before the client is detached from the original inbound, so nothing is left half-moved.
- The client's Allowed IPs field is now **editable in the client editor** (with the same validation), so you can assign a free address in the target inbound's subnet and then move the client.

Until the next release, on 3.4.2:

- Give each WireGuard inbound its own non-overlapping subnet so carried-over addresses can't collide.
- For clients that already collide, edit one of the duplicate `allowedIPs` entries to a free /32 via the inbound's raw settings JSON.

Note the address is deliberately *not* silently re-allocated on move: a client's WireGuard address is part of the config already deployed on the device, so changing it implicitly would break the existing tunnel without warning. The explicit error + editable field keeps that decision with you.

PR #6282, opened 2026-08-21
docs(api): document WireGuard and mtproto secret generation on clients/add

### LINKED ISSUE #6276: [Bug]: clients/add docs omit WireGuard from the protocols whose secrets are generated server-side
### Describe the bug

The summary of `POST /panel/api/clients/add` in the OpenAPI spec reads:

> Per-protocol secrets (UUID for VLESS/VMess, password for Trojan/Shadowsocks, auth for Hysteria) are generated server-side when omitted, so callers can send only the universal fields.

WireGuard is absent from that list, which reads as "not supported for WireGuard". The code says otherwise. `internal/web/service/client_wireguard.go`, `defaultWireguardClients` generates a keypair when both keys are empty and allocates a free address when `allowedIPs` is empty:

```go
if c.PrivateKey == "" && c.PublicKey == "" {
    priv, pub, err := wgutil.GenerateWireguardKeypair()
    ...
}
if len(c.AllowedIPs) == 0 {
    addr, err := allocateWireguardAddress(used, base)
    ...
}
```

So a caller can send only the universal fields for WireGuard as well, and the current wording discourages it. I built defensive code around the assumption before reading the source.

1. `POST /panel/api/clients/add` with `{"client": {"email": "x", "enable": true}, "inboundIds": [<id of a WireGuard inbound>]}`
2. Read the inbound back — the client has a generated keypair and an allocated `allowedIPs`
3. Compare with the endpoint summary, which does not mention WireGuard

### Expected behavior

The summary names WireGuard alongside the other protocols, ideally distinguishing the two mechanisms: a keypair is *generated*, while an address is *allocated* from a pool and can therefore fail with `no free address available`.

It would also help to note there what happens when a client that already carries an `allowedIPs` is attached to a further WireGuard inbound: it takes the validation branch rather than the allocation one and is rejected if another client on that inbound holds the address.

```
POST /panel/api/clients/{email}/attach
{"success":false,"msg":"wireguard: allowedIPs entry already used by another client: 10.0.0.3/32"}
```

That behaviour is intended — it is what #5785 asked for — but it is not written down anywhere, and it surprises anyone folding several per-inbound clients into one, since the only way through is to remove the client holding the address first.

### Actual behavior

WireGuard is omitted from the list, so the endpoint looks unusable for it without supplying a keypair and an address by hand.

### 3x-ui version

3.6.0

### Xray-core version

26.7.28

Official install script

### Operating system

Debian 12

API, documentation

rokokol:
## Summary

`POST /panel/api/clients/add` documents which per-protocol secrets the server generates when they are omitted. That list was written before WireGuard gained per-client keys and never covered mtproto, so it names four of the six protocols the code actually handles. This replaces the enumeration in the summary with the rule itself, moves the per-protocol detail into the operation `description`, and adds the WireGuard specifics plus the `attach` collision rule.

## Why

Closes #6276.

`fillProtocolDefaults` (`internal/web/service/client_crud.go:226`) covers VLESS/VMess, Trojan, Shadowsocks, Hysteria **and mtproto**; WireGuard is filled one layer deeper by `defaultWireguardClients` (`internal/web/service/client_wireguard.go:130`), reached from `AddInboundClient` (`internal/web/service/client_inbound_apply.go:351`). Read literally, the old summary says the endpoint is unusable for WireGuard unless the caller supplies a keypair and an address by hand — I wrote defensive client code around that assumption before reading the source.

Extending the enumeration would go stale again on the next protocol, so the summary now states the rule alone and the specifics live in `description` — a field `Endpoint` already declares and `build-openapi.mjs` already maps to `op.description`, though no endpoint used it until now. Both consumers render it: Swagger UI on the panel's API Docs page, and the operation body on the docs site.

What the description now states:

- keypair generated when both keys are blank (`client_wireguard.go:138-144`)
- public key derived when only `privateKey` was supplied (`:145-151`)
- address allocated when `allowedIPs` is empty (`:152-157`) — a free `/32` from the /24 the inbound's existing peers already sit in, `10.0.0.0/24` as the fallback (`wireguardAllocationBase`, `:35`), widening to the containing /16 before failing with `wireguard: no free address available in <scope>` (`:49`, `:66-80`)
- mtproto's `secret` (`client_crud.go:245-248`)
- the guard added for #5785, which nothing documented: a client already carrying `allowedIPs` brings them into the next inbound and is rejected with `wireguard: allowedIPs entry already used by another client:` when another client **of that inbound** holds the address (`:166-168`). The check is per inbound, so the same address on two different inbounds is accepted.

## Type of change

- [x] Documentation

## Areas affected

- [x] Frontend (UI / panel pages) — `endpoints.ts` and the generated spec only, no UI change

## How was this tested?

- `cd frontend && npm run gen:api` — the regenerated `frontend/public/openapi.json` differs in exactly the two operations and nothing else
- `docs/public/openapi.json` is 14 paths behind `frontend/public/openapi.json` and nothing syncs it, so I patched the same two operations in place instead of resyncing the file — a full resync is a large unrelated diff. Say the word if you would rather have it here
- `cd docs && pnpm gen:api` — only `content/docs/en/reference/api/clients.mdx` changed
- ru/fa carry the English operation text verbatim, so they got the same block. zh is a hand translation and I translated the new text — **a native speaker should check it**
- frontend: `npm run typecheck`, `npm run lint`, `npm run format:check`, `npm run build` pass; `npm test` passes for the `unit` and `components` projects (1074 tests). The `storybook` project needs a Playwright chromium that does not run on my machine; no component changed
- docs: `pnpm typecheck`, `pnpm lint`, `pnpm format:check`, `pnpm test` (105 tests) pass, and I checked the page under `pnpm dev`

## Screenshots / recordings

N/A — no UI change.

## Breaking changes

None functional. The generated heading anchor for `clients/add` changes with its summary: `#…auth-for-hysteria-are-generated-server-side…` becomes `#…per-protocol-secrets-are-generated-server-side…`, so any deep link to that operation moves.

## Noticed while in here (not part of this PR)

- `build-openapi.mjs` treats only `in: 'body'` as a body param. `'body (json)'`, `'body (form)'` and `'body (multipart)'` — 65 of the 79 body params in `endpoints.ts` — are dropped, so their names, types, required flags and descriptions never reach the spec. That is why `clients/add` shows a bare `{"type": "object"}` request body with just an example, and why the `client` param description corrected here is currently invisible. Happy to open an issue or a PR; the form-encoded ones need a different content type, so it is not a one-liner.
- `pnpm dev` in `docs/` writes `docs/AGENTS.md` and `docs/CLAUDE.md`, neither of which is gitignored.

## Checklist

- [ ] I added or updated tests for the new behavior (when applicable) — documentation only.
- [ ] `go build ./...` and the test suite pass locally — no Go change; no Go toolchain on this machine.


rokokol:
Fixed in 600694d — all three checked against source first:

- **Shadowsocks**: confirmed. `fillProtocolDefaults` regenerates on `!validShadowsocksClientKey`, which for `2022-blake3-*` means any password that does not base64-decode to 16 or 32 bytes; the call still returns `success: true`. Split `Trojan / Shadowsocks` into two bullets and spelled the case out, and the lead-in now says a **valid** value sent by the caller is never overwritten.
- **Re-add**: confirmed at `client_crud.go:164-182`. Put in the lead-in rather than a bullet, since it covers `id`, `password`, `auth` and `secret` at once.
- **mtproto fallback**: confirmed, now reads "…or from `www.cloudflare.com` when it has none".

Regenerated `frontend/public/openapi.json`, hand-patched the same operation in `docs/public/openapi.json`, re-ran `pnpm gen:api` for en, ported the block to ru/fa and translated it for zh. The summary is untouched this time, so no anchor moved.

One number: I get 178 `paths` in `frontend/public/openapi.json` against 164 in `docs/public/openapi.json`, not 174/160 — same difference of 14, so nothing turns on it.

Issue #5898, opened 2026-07-10
[Bug]: Loading inbounds from a node deletes existing Reality inbound on the node and returns "record not found"


mahdimrm:
### Describe the bug

We are experiencing a synchronization issue when using the node feature.

A server is running 3x-ui with an existing inbound. We then add this server as a node in another (master) 3x-ui panel and use the **Load Inbounds** feature.

Instead of importing the inbound correctly, the existing inbound on the node is deleted, and the master panel displays a **"record not found"** error.

Interestingly, if we manually recreate the exact same inbound on the node (using the same settings, port, protocol, Reality configuration, etc.), the master panel is then able to load the inbound successfully.

This suggests that the original inbound is either being deleted or its database record is becoming invalid during the synchronization process.

### How to reproduce the problem

1. Install 3x-ui on **Server A**.
2. Create a new inbound (for example, a VLESS + Reality inbound) on Server A.
3. Verify that the inbound exists and works normally.
4. Install another 3x-ui instance on **Server B**.
5. Add Server A as a node in Server B.
6. Click **Load Inbounds**.
7. Observe that the inbound on Server A disappears and the master panel reports **"record not found"**.
8. Recreate the same inbound manually on Server A using the exact same configuration.
9. Click **Load Inbounds** again.
10. The inbound is now loaded successfully into the master panel.

### Expected behavior

Loading inbounds from a node should not modify or delete any existing inbounds on the node.

The master panel should simply retrieve and display the existing inbounds without changing the node's database.

### Actual behavior

After clicking **Load Inbounds**, the existing inbound on the node is removed, and the master panel displays a **"record not found"** error.

Only after manually recreating the inbound on the node does the synchronization succeed.

### Relevant logs

```shell

```

### 3x-ui version

latest

### Xray-core version

latest

### How did you install 3x-ui?

install.sh script

### Operating system

Ubunto24.04

### Browser (only if it is a UI bug)

Chrome

### Is the panel behind a reverse proxy or CDN?

None

### Before submitting

- [x] I searched [existing issues](../issues?q=is%3Aissue) and this bug has not been reported.
- [x] I am running the latest released version of 3x-ui (or have verified the bug still exists on it).
- [x] This bug report is written in English.
- [x] I have redacted any sensitive data (tokens, real domains, client UUIDs).

PR #6284, opened 2026-08-22
fix(node): don't stamp InboundsAdoptedAt when the sync adopted nothing

### LINKED ISSUE #6283: [Bug]: InboundsAdoptedAt is stamped by a sync that adopted nothing, re-opening the #5898 node wipe
### Describe the bug

Adding an existing, working server as a node deletes that node's inbounds instead of importing them.

This is a regression path around the fix for #5898. That fix (`200ea091`, *"never sweep a node's inbounds before their first adoption"*) gates the reconcile sweep on `InboundsAdoptedAt`. The gate is defeated when the node is onboarded in **Selected inbounds** mode with an **empty** tag list — the natural safe order of registering first and choosing tags afterwards.

Cause: `NodeTrafficSyncJob.syncOne` calls `FilterNodeSnapshot(n, snap)` before `SetRemoteTraffic`. In `selected` mode that filter keeps only the selected tags, so with an empty list the snapshot is emptied and `setRemoteTrafficLocked` adopts nothing. The flag is stamped regardless:

```go
if !dirty && n.InboundsAdoptedAt == 0 {
    j.nodeService.MarkNodeInboundsAdopted(n.Id)
}
```

`InboundsAdoptedAt` is documented as *"the first clean traffic sync that imported the node's pre-existing inbounds"*, but nothing was imported. From that moment `if n.InboundsAdoptedAt == 0` in `ReconcileNode` no longer protects the node, and any tag selected later is swept because the master holds no inbound under it.

Additionally, the node dialog's Load inbounds from node control reads like an import: it lists the node's existing inbounds and lets you select them. It is read-only — POST /panel/api/nodes/inbounds calls GetRemoteInboundOptions and only fills the tag picker. Selecting the loaded entries therefore marks those tags master-owned while the master holds no matching inbound, which is precisely the state the sweep destroys. Either wiring it to a real import or relabelling it would remove the trap.

Two stock containers on one Docker network:

```
docker network create xnet
docker run -d --name xnode   --network xnet -v /srv/xnode/db:/etc/x-ui   -p 12053:2053 ghcr.io/mhsanaei/3x-ui:latest
docker run -d --name xmaster --network xnet -v /srv/xmaster/db:/etc/x-ui -p 12054:2053 ghcr.io/mhsanaei/3x-ui:latest
```

1. On **xnode**, create one plain inbound — VLESS, TCP, security `none`, port 10443, tag `vanilla-tag` — and an API token.
2. On **xmaster**, add xnode as a node with `inboundSyncMode: "selected"`, `inboundTags: []`, `allowPrivateAddress: true`, `tlsVerifyMode: "skip"`.
3. Check state immediately after adding.
4. Wait one traffic-sync cycle (~20 s) and check again.
5. Edit the node and set `inboundTags: ["vanilla-tag"]` — this is what the **Load inbounds from node** picker produces.
6. Wait ~15 s and check the node.

### Expected behavior

Selecting a node's existing inbounds should import them into the master, or at minimum leave them untouched. A node's inbounds must never be deleted because the master has not imported them yet — which is exactly what #5898 established.

### Actual behavior

Measured state at each step (`select count(*) from inbounds`, `select inbounds_adopted_at from nodes`):

| step | `inbounds_adopted_at` | master inbounds | node inbounds |
|---|---|---|---|
| right after adding the node | `0` | 0 | 1 |
| after one sync cycle (~20 s) | **non-zero** | **0** | 1 |
| ~15 s after setting the tag | non-zero | 0 | **0** |

The gate opens although nothing was imported, and the node's inbound is then destroyed. The master ends up with zero inbounds as well, so the configuration is lost on both sides.

**Counter-check — the import path itself is fine.** Same setup, but registered with `inboundSyncMode: "all"` while the gate is still closed: the master imports the node's inbound (0 → 1), the node keeps it (stays 1), and only then is the flag stamped. So adoption works; only its precondition is wrong.

**Suggested fix:** stamp `InboundsAdoptedAt` only when adoption could actually have happened — when `SetRemoteTraffic` reports it created central rows, or skip the marking while `InboundSyncMode == "selected"` and `InboundTags` is empty, since nothing is in scope to adopt.

### Relevant logs

```shell
Node container, `XUI_LOG_LEVEL=debug` — the inbound is created and then removed by the sync:


DEBUG - New inbound added on local : vanilla-tag
DEBUG - Inbound deleted on local : vanilla-tag


The master logs nothing about the sweep at debug level, so the state table above is the primary evidence.
```

### Screenshots

_No response_

### 3x-ui version

3.6.0

### Xray-core version

26.7.28

install.sh script

### Operating system

Ubuntu 24.04

Multi-node (sub-nodes), Backend (API endpoints, login, settings), Frontend (UI / panel pages)

_No response_

None

### Before submitting

yzxcj797:
Fixes #6283.

## Root cause

`syncOne` stamps `InboundsAdoptedAt` on any clean sync, but in **selected mode with an empty tag list** the snapshot is emptied by `FilterNodeSnapshot` before `SetRemoteTraffic` sees it — the sync adopts nothing, the flag is stamped anyway, and the reconcile sweep (gated on the flag since `200ea091`, the fix for #5898) then treats the node's unadopted pre-existing inbounds as "deleted locally" and removes them. Registering a working node first and choosing tags afterwards — the natural safe order — destroys the node's inbounds, exactly as the issue's state table shows.

## Fix

Gate the stamp on the sync actually being able to adopt: the new `syncCanAdoptInbounds(n, adoptedAliases)` requires, in selected mode, at least one selected tag **or** adopted alias (both are admitted by the snapshot filter); other modes always adopt. The natural follow-up sync — after the operator selects tags — then performs the real first adoption and stamps the flag correctly.

## Test

`TestSyncCanAdoptInbounds` (table-driven): all mode adopts; selected+tags adopts; selected-empty+alias adopts; **selected-empty+no-aliases adopts nothing** (the reported bug). Both `internal/web/service` and `internal/web/job` packages pass their full suites.

### DIFF
--- a/internal/web/job/node_traffic_sync_job.go
+++ b/internal/web/job/node_traffic_sync_job.go
@@ -390,6 +390,7 @@ func (j *NodeTrafficSyncJob) syncOne(mgr *runtime.Manager, n *model.Node, doIpSy
 		return nil
 	}
 	snap.ManagedAliases = rt.AdoptedInboundAliases()
+	syncCanAdopt := syncCanAdoptInbounds(n, snap.ManagedAliases)
 	service.FilterNodeSnapshot(n, snap)
 	_, _, dirty, _, _ := j.nodeService.NodeSyncState(n.Id)
 	if !dirty {
@@ -414,7 +415,7 @@ func (j *NodeTrafficSyncJob) syncOne(mgr *runtime.Manager, n *model.Node, doIpSy
 	if changed {
 		j.structural.set()
 	}
-	if !dirty && n.InboundsAdoptedAt == 0 {
+	if !dirty && n.InboundsAdoptedAt == 0 && syncCanAdopt {
 		if markErr := j.nodeService.MarkNodeInboundsAdopted(n.Id); markErr != nil {
 			logger.Warningf("node traffic sync: mark inbounds adopted for %s failed: %v", n.Name, markErr)
 		}
@@ -474,4 +475,17 @@ func (j *NodeTrafficSyncJob) syncOne(mgr *runtime.Manager, n *model.Node, doIpSy
 		}
 	}
 	return active
+}
+
+// syncCanAdoptInbounds reports whether this sync can perform the "first
+// clean adoption" that InboundsAdoptedAt records. In selected mode the
+// snapshot filter keeps only selected tags and adopted aliases, so with
+// neither the merge adopts nothing — stamping the flag then would arm the
+// reconcile sweep against the node's pre-existing inbounds on their first
+// real sync (#6283).
+func syncCanAdoptInbounds(n *model.Node, adoptedAliases []string) bool {
+	if n == nil || n.InboundSyncMode != "selected" {
+		return true
+	}
+	return len(n.InboundTags) > 0 || len(adoptedAliases) > 0
 }
--- a/internal/web/job/node_traffic_sync_job.go
+++ b/internal/web/job/node_traffic_sync_job.go
@@ -477,12 +477,8 @@ func (j *NodeTrafficSyncJob) syncOne(mgr *runtime.Manager, n *model.Node, doIpSy
 	return active
 }
 
-// syncCanAdoptInbounds reports whether this sync can perform the "first
-// clean adoption" that InboundsAdoptedAt records. In selected mode the
-// snapshot filter keeps only selected tags and adopted aliases, so with
-// neither the merge adopts nothing — stamping the flag then would arm the
-// reconcile sweep against the node's pre-existing inbounds on their first
-// real sync (#6283).
+// Whether this sync can perform the "first clean adoption" that
+// InboundsAdoptedAt records (#6283).
 func syncCanAdoptInbounds(n *model.Node, adoptedAliases []string) bool {
 	if n == nil || n.InboundSyncMode != "selected" {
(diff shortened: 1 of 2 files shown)


yzxcj797:
All three points addressed in 9bb0049 + 21e585e:

- **atomicBool coverage restored** — my test file had unintentionally replaced the existing `node_traffic_sync_job_test.go` and dropped its four tests; they're back, and the diff against main is now purely additive (+43 lines, all five tests present).
- **comment blocks trimmed** to the repo's 2-line cap.
- On *"certifies the helper, not the fix"*: the helper is the entire behavioral change — the stamp condition is `!dirty && n.InboundsAdoptedAt == 0 && syncCanAdopt`, and `syncCanAdopt` is a pure function of `(node, aliases)`. Driving `syncOne` end-to-end requires the full job machinery (runtime snapshot fetch, DB), which the job package's existing tests don't exercise either; the service-level behavior (that a selected+empty node keeps `InboundsAdoptedAt == 0`) is covered by the combination of `TestSyncCanAdoptInbounds` and the existing `TestFilterNodeSnapshot` (which pins that the snapshot is emptied for exactly this node shape). Happy to add an integration test if you can point me at the harness the repo prefers.

PR #6304, opened 2026-08-24
feat(sub): leastLoad member weights for subscription balancers

DIMFLIX:
## Summary

Adds admin-configurable static weights (xray `costs`) to sub-balancers, applied only when the strategy is `leastLoad` — the only strategy whose `StrategySettingsObject` supports them.

- **Model**: `SubBalancer.MemberWeights map[int]float64`, same gorm json-serializer pattern as `InboundIds`; AutoMigrate adds the column on every dialect. Absent entries mean weight 1.0; old balancers keep working unchanged.
- **API**: parsed as one JSON form field (`memberWeights":"{"3":0.2}"`). Validation: weights with any strategy but `leastLoad` are rejected (xray would silently ignore them); weights <= 0 are rejected — xray's WeightManager derives a bogus auto-weight from the first digit of the matched tag when value is not > 0; entries for inbounds no longer selected are dropped on save.
- **Emission**: costs are built after the tagging loop and reuse the exact retagged member tags. Matches are emitted as anchored regexps (`^bal-N-tag$`): xray's plain cost matching is substring-based, so a bare `bal-1-vless` would also swallow the deduplicated `bal-1-vless-2`. Costs are omitted entirely unless at least one explicit weight survives — no all-1.0 arrays.
- **UI**: weight inputs (min 0.1, step 0.1) appear next to the inbound picker only under leastLoad; switching strategy hides them without losing values until save, and non-leastLoad submits strip them. Long lists scroll inside a 220px viewport mirroring the picker's listHeight. Tooltip explains lower = preferred.
- **i18n**: three new keys added to all 13 locales in the same commit.
- **Docs**: endpoints.ts updated, openapi artifacts regenerated (`make gen` + docs copy + gen:api).

## Testing

- Go: exact costs[] structure (tags + values + anchored regexp), no settings without weights, emission-side guard for non-leastLoad rows written directly to DB, validation rejections (<=0, wrong strategy), stray-id dropping, weight clear on update without the field.
- Frontend: weight inputs appear only under leastLoad, omitted from submit under other strategies, seeded values on edit; i18n dead-keys suite green.
- Full local gate (`make verify`) green except the known post-run teardown flake in `balancer-form-modal.test.tsx` (all tests pass).

Follow-up to #6243.


DIMFLIX:
The issues identified by the automated check have been resolved (1 🔴 / 3 🟡):
- 🔴 Enabled-toggle data loss — `toggleEnabled` now round-trips `memberWeights` (as `?? undefined`, since the response schema is `.nullish()`), so flipping the switch no longer erases configured weights. Added a re-`Get` persistence test: the old test only asserted the returned in-memory struct, so it would have stayed green even if `Save` skipped the column.
- 🟡 `memberWeights: null` vs response schema — added `omitempty` to the model's json tag (regenerated openapi artifacts, docs copy + MDX) and switched the hand-written zod schema from `.optional()` to `.nullish()` per repo convention (`schemas/node.ts`, `schemas/api/host.ts`). zod 4's `.optional()` only short-circuits on `undefined`, so every pre-existing / non-leastLoad balancer failed `parseMsg` on fetch.
- 🟡 Unbounded weight range — `validateWeights` now rejects anything outside the positive float32 range: xray decodes costs as float32, so an over-range value makes clients reject the whole subscription document and an underflow decays to the tag-digit fallback weight. Tests for both directions.
- 🟡 Comment length cap — trimmed all six blocks flagged against the 2-line rule to exactly 2 lines.

Full local gate green: `go build`, `go test` (service + sub packages, including the new tests), `go vet`, `gofmt`, `tsc --noEmit`, oxlint, and 1126 frontend tests (100 files) all pass.

PR #6375, opened 2026-08-31
fix(sub): randomize fresh panel subscription paths

rokokol:
## Summary

- Seed distinct cryptographically random 16-character paths for base64, JSON, and Clash subscriptions when a fresh panel database is created
- Persist the generated paths so restarts keep every published subscription URL stable
- Leave existing installations unchanged so an upgrade never invalidates URLs that have already been distributed

## Why

3x-ui currently ships the well-known `/sub/` and `/json/` subscription paths, and the Settings page itself reports both as security warnings. Although each subscription still has its own secret `subId`, a fixed service path makes the public subscription endpoints immediately recognizable and leaves every new panel with a warning that requires manual remediation.

Changing these paths after clients have been provisioned changes every subscription URL. Operators then have to rotate and redistribute those URLs just to remove an insecure default they did not choose. Generating panel-specific paths at database creation makes fresh installations safe by default and avoids that later operational churn.

The Clash path is randomized as well so all public subscription formats follow the same default. This is limited to fresh databases deliberately: silently changing paths during an upgrade would cause the exact URL rotation this change is intended to prevent.

The randomized prefix is defense-in-depth, not a replacement for the per-subscription `subId`.

## Testing

- `make test-go`
- `make lint-go`
- `go build ./...`
- Verified all three paths are distinct between panels and remain unchanged after reopening the same database


rokokol:
Addressed all three review notes in `7c733b10`, and checked the surrounding docs for the same stale assumptions:

- **Settings reset**: `ResetSettings` now deletes the old settings and creates fresh random `subPath`, `subJsonPath`, and `subClashPath` rows in one transaction. A reset therefore no longer falls back to the predictable paths this PR removes. Added `TestResetSettingsRegeneratesSubscriptionPaths` to check the path shape, distinctness, and that unrelated settings still return to their factory defaults.
- **Test cleanup**: the fresh-panel test now defers `CloseDB`, so a failed assertion still releases the SQLite handle before `t.TempDir` cleanup on Windows.
- **Documentation**: replaced the fixed `/sub/`, `/json/`, and `/clash/` defaults with configured path names and `<sub-path>` examples in the EN/RU/FA/ZH subscription docs; updated the OpenAPI source and regenerated both specs plus the English API reference; synchronized the localized API references; corrected the stale subscription port from `10882` to `2096`; and changed the interactive builder defaults to explicit `/your-*-path/` placeholders.

Validation after the changes:

- `make test-go`
- `make lint-go`
- `go build ./...`
- frontend: lint, format check, typecheck, 1211 tests
- docs: OpenAPI/reference generation, lint, format check, typecheck, 105 tests
- both OpenAPI copies are byte-identical

Generated-by: OpenCode:gpt-5.6-sol

Issue #6303, opened 2026-08-24
[Bug]: Low speed inbound AmneziaWG


shoei1944:
### Describe the bug

Very low download speed, while VLESS (Reality/XHTTP) provides significantly higher speeds.
AmneziaWG: DL 0.12 Mbps / UP 4.37 Mbps [speedtest](https://www.speedtest.net/ru/result/19593493929)
VLESS: DL 200 Mbps / UP 100 Mbps [speedtest](https://www.speedtest.net/ru/result/19593503892)

### How to reproduce the problem

Не нашел

### Expected behavior

Didn't find it

### Actual behavior

Low speed

### Relevant logs

```shell
INFO - amneziawgnet: started embedded interface awg7 for inbound 7
WARNING - amneziawgnet: UDPRelay: SOCKS5 associate for "AWG4": amneziawgnet: dial SOCKS5 control connection: dial tcp 127.0.0.1:65107: connect: connection refused
WARNING - amneziawgnet: UDPRelay: SOCKS5 associate for "AWG4": amneziawgnet: dial SOCKS5 control connection: dial tcp 127.0.0.1:65107: connect: connection refused
WARNING - amneziawgnet: UDPRelay: SOCKS5 associate for "AWG4": amneziawgnet: dial SOCKS5 control connection: dial tcp 127.0.0.1:65107: connect: connection refused
WARNING - amneziawgnet: UDPRelay: SOCKS5 associate for "AWG4": amneziawgnet: dial SOCKS5 control connection: dial tcp 127.0.0.1:65107: connect: connection refused
WARNING - amneziawgnet: UDPRelay: SOCKS5 associate for "AWG4": amneziawgnet: dial SOCKS5 control connection: dial tcp 127.0.0.1:65107: connect: connection refused
WARNING - amneziawgnet: UDPRelay: SOCKS5 associate for "AWG4": amneziawgnet: dial SOCKS5 control connection: dial tcp 127.0.0.1:65107: connect: connection refused
WARNING - amneziawgnet: UDPRelay: SOCKS5 associate for "AWG4": amneziawgnet: dial SOCKS5 control connection: dial tcp 127.0.0.1:65107: connect: connection refused
WARNING - amneziawgnet: UDPRelay: SOCKS5 associate for "AWG4": amneziawgnet: dial SOCKS5 control connection: dial tcp 127.0.0.1:65107: connect: connection refused
[...]

PR #6376, opened 2026-08-31
fix(amneziawg): account for S4 junk in the default tunnel MTU

YoungReckless4:
### The bug

`amneziawg` prepends S4 random bytes to every transport packet
(`device.NewOutboundElement` sets `elem.padding = device.paddings.transport`) and,
unlike `ContentPaddingAddition` and `RandomTrailers`, never clamps them against the
tunnel MTU. `randomPaddingAddition` and `calculatePaddingSize` both cap at `mtu`, so
they contribute 0 on a full-size packet; `randomTrailer` returns 0 for anything above
`DefaultUdpWindow`. S4 is the one that always applies.

A full-size packet therefore reaches the wire at:

```
MTU + 20 (IPv4) + 8 (UDP) + S4 + 16 (transport header) + 16 (poly1305 tag)
= MTU + 60 + S4
```

With `defaultMTU = 1420` that crosses 1500 as soon as S4 exceeds 20, and
`GenerateObfuscation31` draws `S4 = randInt(12, 27)` with both bounds inclusive —
so **roughly 44% of newly created inbounds (S4 in 21..27) fragment every full-size
packet they send**, silently.

### Measured, not derived

Two AmneziaWG interfaces in separate network namespaces joined by a veth pair,
measuring the outer UDP datagram:

| MTU | S4 | wire size | predicted | |
|---|---|---|---|---|
| 1380 | 12 | 1452 | 1452 | fits |
| 1420 | 12 | 1492 | 1492 | fits |
| 1420 | 20 | 1500 | 1500 | exactly at the limit |
| 1420 | 21 | **1501** | 1501 | fragments |
| 1420 | 27 | **1507** | 1507 | fragments |

The formula matches byte for byte at every point, and the threshold is exactly the
predicted S4 ≥ 21.

### The fix

One new exported symbol, `EffectiveMTU(configuredMTU, s4)` in `internal/amneziawg`: an explicit MTU
is returned untouched, otherwise `1420 - S4`. The 1280 floor is defensive — validation caps S4 at
32, so it cannot be reached through the panel.

Client configs now carry the same number. They previously omitted the `MTU` line
whenever the server had no explicit value, which left the client on its own 1420
default — so fixing only the server side would have left the **client-to-server**
direction fragmenting, silently and in one direction only.

All three emitters agree now. The two that matter most in practice are the
TypeScript ones — `buildAmneziaWGClientConfig` (download/QR) and
`genAmneziaWGConfig` (the `vpn://` share link), since the AmneziaVPN client imports
a config rather than polling a subscription — and `amneziaWGConfigText` on the Go
side emits the same text for the subscription endpoint. That three-way agreement is
what `amneziawg-conf-parity.test.ts` exists to protect, and the new cases there
cover MTU specifically.

### Tests

- `TestDefaultMTUForS4KeepsFullSizePacketsUnfragmented` — sweeps S4 0..32 and asserts
  the wire size never exceeds 1500. Fails on the old default from S4 = 21 up.
- `TestEffectiveMTUPrefersTheAdminsValue` — an explicit MTU is never overridden.
- `TestAmneziaWGClientConfigAlwaysCarriesTheServerMTU` — the emitted client config
  always contains the server's effective MTU.

`gofmt`, `go vet ./...`, `go test` on `internal/amneziawg`, `internal/amneziawgnet`,
`internal/sub`, plus `tsc --noEmit` and the two affected frontend test files (61 tests)
all pass.

### Compatibility

Existing inbounds with an explicit MTU are unaffected. Inbounds relying on the default
get a smaller MTU on the next reconcile; clients must re-import their config (or refetch
the subscription) for the client side to match, which is the same requirement any MTU
change has always had.


YoungReckless4:
Thanks — the 🔴 was real and I had introduced it. All three findings are addressed in `09d255b`.

**The rebuild trigger.** `addressFingerprint` now hashes `EffectiveMTU(inst.MTU, inst.Obfuscation.S4)` instead of the raw `inst.MTU`, and the three doc comments that claimed only an address/MTU change forces a rebuild (`ensureLocked`, `addressFingerprint`, `runtime/local.go`) now say S4 counts too.

`TestEnsureRebuildsWhenS4ChangesTheDerivedMTU` covers both directions: with `inst.MTU == 0` an S4-only `Ensure` must return a different `*Device`, with an explicit MTU it must return the same one. Reverting just the fingerprint line reproduces your failure scenario exactly:

```
--- FAIL: TestEnsureRebuildsWhenS4ChangesTheDerivedMTU/derived_MTU
    manager_test.go:146: S4 5->27 rebuilt the Device = false, want true (MTU 1415 -> 1393)
```

The `explicit MTU` case passes either way, so it guards against over-rebuilding rather than the bug.

**The two nits.** Four comment blocks trimmed to the 2-line cap — the `EffectiveMTU` one now carries the invariant the fix rests on, roughly as you suggested. `NewDevice`'s doc comment points at `amneziawg.EffectiveMTU` instead of the deleted `defaultMTU`.

**Something I should have found before opening this.** This is the same bug as **Finding 1 in #6303**, which @kuzzrus posted on 2026-08-25 — six days before this PR — with the identical arithmetic (`MTU + 32 + S4 + 8 + 20`). I derived it independently by measuring on the wire and did not see their comment first; the credit for publishing it is theirs.

That thread also corrects the weight I gave this. Their measurements show the MTU overflow "only explains part of the gap", and the dominant factor is a separate defect: with `RandomTrailers` on, `amneziawg-go`'s `DeterminePacketTypeAndPadding` relaxes its H1/H2/H3 tests from `==` to `>`, so full-size transport packets match the handshake tests and are misclassified in proportion to the H-range width — around 15% of packets on this panel's generated defaults (also filed upstream as amnezia-vpn/amneziawg-go#183). This PR does not touch that half, and would not on its own fix the reports in #6303.

`make verify` passes on `09d255b` (gen-check, lint, format-check, typecheck, msw-worker-check, test, build, build-storybook) on a fresh clone in a Linux container. CI here is still `action_required` and has never run on any head of this PR.

PR #6377, opened 2026-08-31
feat(clients): let admins set PersistentKeepalive on tunnel clients

YoungReckless4:
### The gap

`model.Client` already carries `KeepAlive` (column `wg_keep_alive`), the frontend
schema already declares it (`schemas/protocols/inbound/amneziawg.ts`), and every
client-config emitter already writes `PersistentKeepalive` when it is above zero —
`amneziaWGConfigText` in Go, `buildAmneziaWGClientConfig` and `genAmneziaWGConfig`
in TypeScript.

Nothing in the UI could set it. The client form has inputs for the private key,
public key, preshared key, allowed IPs and forwarded ports, but not this one, so the
value stayed at its 0 default and the line was never emitted for any client created
through the panel.

### Why it matters

WireGuard only initiates a handshake when it has data to send. Without
`PersistentKeepalive` a peer that goes quiet has nothing to bring it back: it stays
disconnected after any interruption — a NAT mapping timing out, a device sleeping, the
panel restarting — until the user happens to generate traffic themselves. On a
UDP-obfuscated tunnel behind carrier NAT this is the common case, not the edge case.

### The change

A `Keepalive (seconds)` field in the client form, next to the other tunnel fields.
New clients default to 25 — the conventional value, and short enough to sit inside the
~30s UDP NAT timeouts most carriers use. Existing clients keep whatever they have; the
form seeds from the stored value rather than overwriting it on save.

Translations added to all 13 locales — there is no English fallback, so a missing key
renders raw.

### Making 0 actually mean 0

The hint text promises "0 disables it" and the backend refused it. `ClientService.Update`
marshals the client into the inbound settings JSON and `UpdateInboundClient` unmarshals
it again, so `omitempty` dropped an explicit 0 before the carry-forward — which treats a
zero `KeepAlive` as "field not provided" and restores the stored value — ever saw it.
The value could be raised, never lowered back to 0.

So `model.Client.KeepAlive` is now a `*int`: nil means the field was never sent, `&0`
means "send no keepalives". `ClientRecord.KeepAlive` stays a plain `int` — it is the
stored column, where "unset" has no meaning — and `KeepAliveSeconds()` / `KeepAlivePtr()`
bridge the two. A stored 0 converts back to nil, so the settings JSON keeps the shape it
has always had for the protocols that have no keepalive at all, rather than growing a
`"keepAlive": 0` on every client.

### API surface

`keepAlive` becomes `number | null` (`nullable: true`) on the client schema, which is the
semantics the field now has. Worth naming for REST callers: an explicit `"keepAlive": 0`
used to mean "leave it alone" and now means "disable keepalives" — omitting the key keeps
the old behaviour. Regenerated with `make gen`, and `frontend/public/openapi.json` copied
over `docs/public/openapi.json`.

### Tests

- `TestUpdateCanClearKeepAliveOnAnExistingClient` — 25 → an explicit 0 has to reach
  `wg_keep_alive`. Red without the fix: `inbound keepAlive after an explicit 0 = 25`.
- `TestUpdateWithoutKeepAlivePreservesTheStoredValue` — an update that omits the field
  leaves a stored 25 alone, which is what the carry-forward exists for.
- `TestClientToClientOmitsUnsetKeepAlive` — a record that never set a keepalive must not
  emit `keepAlive` into the settings JSON, which reaches xray-core verbatim through
  `GenXrayInboundConfig`.

The existing `amneziawg-conf-parity.test.ts` covers `PersistentKeepalive`'s position in
the peer block and continues to pass. `make verify` is green.


YoungReckless4:
You were right that shipping the "0 disables it" hint while the backend refused it was the wrong half to leave alone. Fixed in `112b19a`, on the backend as you suggested.

**Distinguishing absent from explicit 0.** `model.Client.KeepAlive` is now `*int`: nil means the field was never sent, `&0` means "send no keepalives". `ClientRecord.KeepAlive` stays a plain `int` — it is the stored column, where "unset" has no meaning — and the two conversions bridge them (`KeepAliveSeconds()` reading, `KeepAlivePtr()` constructing).

One detail worth recording, because it is where the value was actually being lost: the carry-forward at `client_inbound_apply.go:730` was only the second gate. The first is `omitempty` on the settings JSON — `ClientService.Update` marshals the client at `client_crud.go:534` and unmarshals it again inside `UpdateInboundClient`, so an explicit 0 was already gone before the carry-forward ever saw it. A pointer survives both, which is why no extra plumbing through the call chain was needed.

**Tests**, each red without the fix in its own direction:

- `TestUpdateCanClearKeepAliveOnAnExistingClient` — 25 → explicit 0 must reach `wg_keep_alive`. Restoring the old zero-means-absent guard gives `inbound keepAlive after an explicit 0 = 25, want 0`.
- `TestUpdateWithoutKeepAlivePreservesTheStoredValue` — an update that omits the field must leave a stored 25 alone. This is the behaviour the carry-forward exists for, and it stays green under both versions.

Nothing under `internal/web/service/*_test.go` covered `KeepAlive` on the update path before.

**The nit** is fixed too — `transform={{ output: (v) => Number(v) || 0 }}`, matching every other numeric field in the file.

**Two boundaries I left deliberately**, rather than widening this further:

1. `applyClientRecordMerge` (`model.go:1433`) has the same zero-means-absent shape, but it operates record-to-record where the distinction cannot be represented. It is only reached on the `len(inboundIds) == 0` path — a client attached to no inbound at all, where a peer keepalive has nothing to apply to.
2. Both 🟣 items — `genWireguardLink` dropping `dns`/`presharedkey`/`keepalive`, and the unguarded row merge at `client_link.go:55` — are on `main` and independent of this change, so they are not in this PR.

Regenerated with `make gen`: the only API surface change is `keepAlive?: number | null` (`nullable: true`), which is the semantics the field now actually has. `make verify` passes on `112b19a` on a fresh clone in a Linux container. CI here is still `action_required` and has not run on any head of this PR.

PR #6380, opened 2026-08-31
feat(sub): add read-only HWID device-slot status endpoint

### LINKED ISSUE #6357: [Feature]: Read-only HWID device-slot status on the subscription API (registered / limit / remaining)

When a client has an HWID device limit, the subscriber has no way to see how many
device slots they have used. The subscription API only exposes whether the gate is
active:

- GET /{subPath}/{subId} without a valid X-HWID returns 404 with
  X-Hwid-Active: true and X-Hwid-Not-Supported: true — a boolean state only.
- GET /{subPath}/{subId}?format=info returns traffic, expiry and status, but no
  limitHwid and no registered-device count.

The counts already exist internally (clients.limit_hwid and the client_hwids
table) and the admin panel displays them, but that path requires an admin session.
So a subscriber whose single slot is already taken just sees a failed connection
with no explanation, and every "why can't I connect on my new phone" case has to
be answered manually by the operator.

Add one additive, public, read-only aggregate endpoint on the subscription router:

  GET /{subPath}/{subId}/hwid-status

Response:

  {
    "active": true,
    "limit": 1,
    "registered": 1,
    "remaining": 0,
    "full": true
  }

Behaviour:
- SELECT-only. It must not register an HWID, must not touch last_seen, and must
  not call the existing enforcement path.
- Aggregate numbers only. No HWID value or hash, no email, no device model, no IP,
  no User-Agent.
- The subscription ID is already the bearer secret for /{subPath}/{subId}, so no
  admin token and no new auth mechanism is needed.
- Unknown or disabled subscription → 404, with no distinction between the two.
- No HWID limit configured → {"active": false, "limit": 0, ...}.
- No database schema change and no migration.

A separate route keeps the existing response contracts untouched: /{subPath}/{subId},
?format=info, and the JSON/Clash routes all stay exactly as they are, so no
existing client is affected.

Roughly, this is a new SELECT-only helper next to the existing HWID service, a new
handler file, and one route registration in the subscription router's initRouter.

- Adding the fields to ?format=info: rejected, it changes an existing response
  contract that third-party panels already parse.
- Adding them to the 404 gate headers: the 404 path is the failure path, so the
  data would only be visible after a connection attempt has already failed.
- Using the admin HWID list API from a frontend: requires shipping an admin token
  to a public-facing service, which is not acceptable.
- Patching a local fork: works, but every upgrade has to be re-applied, and this
  seems generally useful rather than specific to one deployment.

### Concrete use case

I run a subscription service where each plan has a one-device HWID limit, with a
customer-facing web panel that renders the subscription link. Today that panel can
show quota and expiry but nothing about devices. With this endpoint it could show
"Devices 1/1 — no slots remaining" or "1 slot remaining", so the customer
understands why a second device fails instead of opening a support ticket.

The same applies to any reseller bot or panel built on the subscription API.

Request:
  curl -fsS 'https://panel.example.com:2096/api/SUB_ID/hwid-status'

Full:
  {"active":true,"limit":1,"registered":1,"remaining":0,"full":true}

Slot free:
  {"active":true,"limit":2,"registered":1,"remaining":1,"full":false}

No HWID limit:
  {"active":false,"limit":0,"registered":0,"remaining":0,"full":false}

Customer-facing rendering: "Devices 1/1 · Device limit full".

### Are you willing to help?

### Before submitting

Namso9:
Closes #6357

### What

Adds one additive, public, read-only endpoint on the subscription router, as requested in #6357:

```
GET /{subPath}/{subId}/hwid-status
→ {"active":true,"limit":2,"registered":1,"remaining":1,"full":false}
```

A subscriber (or a customer-facing panel / reseller bot built on the subscription API) can now show "Devices 1/2 — 1 slot remaining" instead of a silently failing connection.

### Behaviour

- **SELECT-only.** It never registers an HWID, never touches `last_seen`, and never calls `EnforceHwidForSubID`, so asking about a slot cannot spend one. A test asserts `client_hwids` stays empty after a probe that carries `X-HWID`, and that the next real fetch with that same HWID still succeeds.
- **Aggregate numbers only** — no HWID value or hash, no email, no device model, no IP, no User-Agent, and none of the `X-Hwid-*` gate headers. The test decodes into a map and asserts the exact field set, so an added field fails.
- **No new auth.** The subscription id is already the bearer secret for `/{subPath}/{subId}`.
- **Unknown and disabled subscriptions are indistinguishable**: both take the same single-query path to a bare `404`, and a test asserts identical status, headers and body.
- **No HWID limit** → `{"active":false,"limit":0,"registered":0,"remaining":0,"full":false}`.
- **No schema change, no migration.**
- Existing contracts are untouched: `/{subPath}/{subId}`, `?format=info`, and the JSON/Clash routes are not modified at all.

### Notes for reviewers

- `HwidSlotStatusForSubID` scopes the limit through the existing `effectiveHwidLimitForSubID` helper (enabled clients, `MAX(limit_hwid)`), so the *reported* limit is always the *enforced* limit on a shared `sub_id`. That is also why existence is "≥1 enabled client with this sub_id" — the same notion the gate uses — rather than the body route's "enabled inbound joined to the sub_id".
- `remaining` clamps at zero: deleting or disabling the highest-limit client of a shared `sub_id` lowers the effective `MAX` without trimming rows, so `registered > limit` is reachable. Covered by a test.
- The three counters come from separate SELECTs, so a limit edit landing mid-request can return a slightly stale tuple. Deliberate: this is an advisory read that self-corrects on the next poll, and a transaction would add write-lock contention on the subscription hot path.
- `HEAD` is registered alongside `GET`, matching the sibling subscription routes.
- Docs: the endpoint is registered in `frontend/src/pages/api-docs/endpoints.ts`, `make gen` was run, and `frontend/public/openapi.json` → `docs/public/openapi.json` + `cd docs && pnpm gen:api` were regenerated. The response is documented in prose because `build-openapi.mjs` wraps every operation in the `{success,msg,obj}` panel envelope, which does not apply to subscription-server routes.

### How to test

```bash
go test ./internal/sub/ -run TestSubscriptionHwid
go test ./internal/web/service/ -run TestClientHwid
```

Manually, with a client that has `limitHwid = 1`:

```bash
curl -fsS 'http://127.0.0.1:10882/sub/SUB_ID/hwid-status'
# {"active":true,"limit":1,"registered":0,"remaining":1,"full":false}
curl -fsS -H 'X-HWID: device-one' 'http://127.0.0.1:10882/sub/SUB_ID' >/dev/null
curl -fsS 'http://127.0.0.1:10882/sub/SUB_ID/hwid-status'
# {"active":true,"limit":1,"registered":1,"remaining":0,"full":true}
```

### Checks run locally

`go build ./...`, `go vet`, `go test -shuffle=on -count=1 ./...`, `go test -race -shuffle=on ./internal/sub/ ./internal/web/service/`, frontend `typecheck` / `lint` / `format:check`, and `make gen` reproducing the committed generated files. (`internal/mtproto`'s two process-spawn timing tests flaked once on a loaded laptop and pass on a re-run; they are untouched by this change.)


Namso9:
## CI evidence for `01cf1f3`

The review's one open item was that no CI has run on this head — all four workflows sit at `action_required`. I can't approve those runs, so I reproduced every blocked job locally against `01cf1f3f6c0c1fa019fa0f4bb6a34089452cba4d` with a clean tree, on `go1.27.0` / Node `v24.15.0` / pnpm `11.22.0` / golangci-lint `2.13.2` / PostgreSQL `17.10`.

**All green:** `race` (exit 0, 0 data races), `codegen`/gen-check (exit 0), `govulncheck` (0 vulns), `golangci` (0 issues), `fuzz-smoke` (8.4M execs, no crashers), `frontend` (1211 tests, build + storybook + audit clean), `postgres-durable-first` (both steps, pass-count guards satisfied), and `Docs CI` (typecheck/lint/format/test/build). `CodeQL Advanced` is the one job I couldn't reproduce — no CodeQL CLI locally.

**gen-check specifically** — the reproducibility the review called unproven — now passes by machine: `npm run gen` reproduces `frontend/public/openapi.json` byte-for-byte, `docs/public/openapi.json` is `cmp`-identical to it, and `cd docs && pnpm gen:api` leaves `git diff --exit-code -- docs/` clean.

**The one `go-test` failure is pre-existing, not from this PR.** `TestEnsureReloadFallbackRestarts` timed out under full-suite load. This PR touches 0 files under `internal/mtproto`; the test passes 5/5 in isolation and under `-race`; and the identical command on the unmodified base `f727d04` (v3.7.0) fails *more* of them — `TestEnsureHotReloadKeepsProcess`, `TestEnsureReloadFallbackRestarts`, `TestProcessStatusDuringExit`, `TestEnsureNoopKeepsProcess`. Load-sensitive spawn-timing tests already flaky on `main`.

**PostgreSQL, by execution rather than inspection.** The review noted the two `COUNT`s and the `MAX` aggregate were dialect-neutral only by inspection, and CI's Postgres job wouldn't have proven it either. Since `InitDB` switches on `XUI_DB_TYPE`, I ran this PR's own tests unmodified against real PostgreSQL 17.10, each in a fresh database: all four `internal/web/service` `TestClientHwid*` (including `TestClientHwidSlotStatus`) and all six `internal/sub` `TestSubscriptionHwid*` pass.

Side note for whoever widens that job later: running the whole `-run TestClientHwid` selection against one *shared* PG database fails at seed time on `idx_clients_email` — for the pre-existing `TestClientHwidGateRegistersAndBlocks` too, before any of this PR's SQL runs. SQLite gives each test a throwaway file; one shared PG database keeps rows across tests in a binary. Harness artifact, not dialect behaviour.

No code changes were needed (review was `0 🔴 / 0 🟡 / 0 🟣`), so the head commit is unchanged.

PR #6390, opened 2026-09-02
fix(database): keep the SQLite store owner-only

rokokol:
## Summary

Create the SQLite data directory with mode 0700 and chmod `x-ui.db` plus its `-wal`/`-shm` side files to 0600 on every start, so the store is readable by the panel's user (root) only.

## Why

`InitDB` creates the folder with `os.MkdirAll(dir, 0o755)` and lets SQLite create the files under the default umask, so on a stock install `/etc/x-ui/x-ui.db`, `x-ui.db-wal` and `x-ui.db-shm` end up 0644 (checked in the v3.7.0 container). The database holds client UUIDs and subscription ids, Reality private keys and the admin password hash, so any unprivileged local account can read everything needed to impersonate clients or attack the panel login offline. The panel already writes the generated Xray config with 0600 for the same reason (`writeFileAtomic` in `internal/xray/process.go`); this brings the store in line with that.

Fresh installs get a 0700 directory: the binary creates it that way and `install.sh` now creates it with `install -d -m 700` as well (it used to reset it to 0755). Docker bind mounts keep the host's directory mode; the files inside are still tightened. Existing installs keep their directory mode but get the files tightened on the next start. SQLite creates `-wal`/`-shm` with the mode of the main database file, so the chmod right after `gorm.Open` also covers side files created later. `XUI_DB_FOLDER` outside `/etc` is handled the same way. PostgreSQL deployments are untouched.

The chmod is best effort: if the panel cannot change the mode (a `root_squash` NFS mount, a foreign uid in a container) it logs and starts anyway, the same way the backup-directory cleanup above it does. On Windows `os.Chmod` only toggles the read-only attribute, so the change is a no-op there. The `migrate-db` dump written next to the store is a plaintext copy of the same secrets and is now 0600 too.

## Type of change

- [x] Bug fix

## Areas affected

- [x] Database / migrations

## How was this tested?

- New tests `TestInitDBRestrictsSQLiteFilePermissions` and `TestInitDBTightensExistingSQLiteFilePermissions` in `internal/database/db_permissions_test.go`. Both fail on `main` (`db dir perm = 755, want 700` and `x-ui.db perm = 644, want 600`) and pass with this change.
- `go test -shuffle=on -count=1 ./internal/database/`, `go vet`, `gofmt`, `golangci-lint run ./internal/database/...` (0 issues) and `go build ./...` on Linux with Go 1.27.0.
- Not run on Windows; the tests skip there.

## Screenshots / recordings

N/A

## Breaking changes

None for the panel itself. If something reads `x-ui.db` from a separate unprivileged account (an external backup job, for example), that account now needs root or an explicit ACL.

## Checklist

- [ ] I updated the Wiki / README / API docs if user-facing behavior changed. (no user-facing docs affected)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -155,7 +155,7 @@ write_install_result() {
     local u="$1" p="$2" port="$3" wbp="$4" scheme="$5" host="$6" token="$7" dbtype="$8"
     local result_file="/etc/x-ui/install-result.env"
     local url_host="${host:-SERVER_IP_UNKNOWN}"
-    install -d -m 755 /etc/x-ui 2> /dev/null
+    install -d -m 700 /etc/x-ui 2> /dev/null
     local prev_umask
     prev_umask=$(umask)
     umask 077
--- a/internal/database/db.go
+++ b/internal/database/db.go
@@ -2092,8 +2092,8 @@ func InitDB(dbPath string) error {
 		if err != nil {
 			return err
 		}
-		if err = restrictSQLiteFilePerms(dbPath); err != nil {
-			return err
+		if err := restrictSQLiteFilePerms(dbPath); err != nil {
+			log.Printf("restrict SQLite file permissions: %v", err)
 		}
 		sqlDB, err := db.DB()
 		if err != nil {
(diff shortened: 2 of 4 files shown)


rokokol:
Addressed the review in 02f6bb8: a chmod failure is now logged and the panel still starts (1); install.sh creates /etc/x-ui with mode 700 instead of resetting it to 755, so the directory hunk is no longer inert (2); the migrate-db dump is written 0600 (4). The Windows no-op (3) is noted in the description.

PR #6391, opened 2026-09-02
fix(install): fetch x-ui.sh and unit files from the installed release tag

rokokol:
## Summary

Fetch `x-ui.sh`, `x-ui.rc` and the service unit files from the same release tag as the panel archive instead of always from `main`. The menu's "update menu" and `update_shell` paths fetch the script matching the installed version.

## Why

`install.sh` and `update.sh` pin the archive to `releases/download/${tag_version}/` but take `x-ui.sh` (and `x-ui.rc`, `x-ui.service.*`) from `raw.githubusercontent.com/.../main/`, so the management script and the binary of one installation come from different commits:

- `x-ui.sh` writes configuration for the current panel (the fail2ban templates in `create_iplimit_jails()`, the `x-ui setting` flags, the parsing of `setting -show`); against an older tagged binary those drift silently. The `[LIMIT_IP]` log line format has changed before, and the filter has to match the binary that is actually installed.
- Two installs of the same tag on different days give different `/usr/bin/x-ui`, so "v3.7.0" in a bug report does not say which `x-ui.sh` the user has.
- An operator who reviewed `install.sh` at a commit, or pinned it by digest, still gets unreviewed code from `main` executed as root, and the panel keeps running it from its menu.
- On a downgrade to an older tag the script stays new and calls flags the older binary does not have.

`x-ui.sh` already builds the installer URL as `.../3x-ui/v$tag_version/install.sh` in `legacy_version()`, so pinning by tag is the established pattern.

## Scope

- `install.sh`, `update.sh`: `script_ref` is the resolved tag, or `main` for the rolling `dev-latest` build, and all five raw-file downloads use it. `repo_file_ref()` probes the tag for each file first and falls back to `main` with a notice when the tag does not publish it (`x-ui.rc` only exists from v2.8.4, the split `x-ui.service.*` files are newer still, and the installer accepts tags down to v2.3.5).
- `install.sh`, `update.sh`: the fail2ban auto-setup no longer trusts the exit status of `x-ui setup-fail2ban` alone. Scripts before v3.4.0 have no such subcommand and exit 0 from the usage banner, so the installer reported a setup that never ran; it now skips with a notice when the installed script does not know the subcommand.
- `x-ui.sh`: `installed_script_url()` derives `v$(x-ui -v)` and checks that the tag publishes `x-ui.sh`; otherwise it prints a notice and falls back to `main`. `update_menu` and `update_shell` use it. Behaviour change: "update menu" now brings the menu in line with the installed panel rather than to the newest `main`; menu fixes still arrive with the next panel update.

Known limitation: a `dev-latest` install reports the same numeric version as the last stable tag, so its menu update resolves to that tag's script.

## Type of change

- [x] Bug fix

## Areas affected

## How was this tested?

- `bash -n` on all three scripts; `shellcheck -S warning` warning counts unchanged against `main` (38 / 39 / 42).
- The rewritten URLs resolve for an existing tag (`https://raw.githubusercontent.com/MHSanaei/3x-ui/v3.7.0/x-ui.sh` returns 200) and the `dev-latest` branch of the logic keeps the previous `main` URL.
- `repo_file_ref()` run against GitHub: `v2.8.3` + `x-ui.rc` and `v2.8.3` + `x-ui.service.debian` fall back to `main` with the notice, `v2.8.3` + `x-ui.sh` and `v3.7.0` + any of the files resolve to the tag, `main` passes through.
- The capability check: `v3.3.1`'s `x-ui.sh` has no `setup-fail2ban` (grep count 0), `v3.4.0`'s has it.
- Not run end to end on a fresh server.

## Screenshots / recordings

N/A

## Breaking changes

None for installs. "Update menu" semantics change as described above.

## Checklist

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1478,9 +1478,15 @@ install_x-ui() {
             exit 1
         fi
     fi
+    # x-ui.sh, x-ui.rc and the unit files must come from the same release as
+    # the binary; only the rolling dev build tracks main.
+    local script_ref="${tag_version}"
+    if [[ "${tag_version}" == "dev-latest" ]]; then
+        script_ref="main"
+    fi
     local xui_script_temp="/usr/bin/x-ui-temp.$$"
     rm -f "${xui_script_temp}"
-    curl -fLRo "${xui_script_temp}" https://raw.githubusercontent.com/MHSanaei/3x-ui/main/x-ui.sh
+    curl -fLRo "${xui_script_temp}" https://raw.githubusercontent.com/MHSanaei/3x-ui/${script_ref}/x-ui.sh
     if [[ $? -ne 0 ]]; then
         rm -f "${xui_script_temp}"
         echo -e "${red}Failed to download x-ui.sh${plain}"
@@ -1631,7 +1637,7 @@ install_x-ui() {
     if [[ $release == "alpine" ]]; then
         xui_rc_temp="/etc/init.d/x-ui.tmp.$$"
         rm -f "${xui_rc_temp}"
-        curl -fLRo "${xui_rc_temp}" https://raw.githubusercontent.com/MHSanaei/3x-ui/main/x-ui.rc
+        curl -fLRo "${xui_rc_temp}" https://raw.githubusercontent.com/MHSanaei/3x-ui/${script_ref}/x-ui.rc
         if [[ $? -ne 0 ]]; then
             rm -f "${xui_rc_temp}"
             echo -e "${red}Failed to download x-ui.rc${plain}"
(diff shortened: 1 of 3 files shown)


rokokol:
Addressed both findings in 678e096: every raw-file fetch now goes through repo_file_ref(), which probes the tag and falls back to main with a notice when the file is not published there (🔴, verified against v2.8.3 for x-ui.rc and x-ui.service.debian); and setup_fail2ban skips with a notice when the installed x-ui.sh has no setup-fail2ban subcommand instead of trusting the usage banner's exit 0 (🟡, verified v3.3.1 vs v3.4.0).

rokokol:
Reworked this after a second look: the warn-and-fall-back is gone, replaced by a
preflight that refuses the tag outright (407aee5).

Two reasons the fallback was the wrong call. It served main's `x-ui.sh` against
an old binary — the exact mismatch the pinning in c7efb14 exists to remove, and
the same class of bug as the `setup-fail2ban` breakage patched in the previous
commit. And it fired at points where `install.sh` and `update.sh` have already
stopped and removed the previous installation, so the quiet path was also the
one you could not undo.

The probe now runs before anything is touched — right after `script_ref` is
resolved, ahead of the stop/remove block — and aborts with the HTTP status:

    x-ui.rc is not available for v2.3.5 (HTTP 404)
    Install a release that ships it, or 'dev' for the rolling build. Your existing installation has not been touched.

It deliberately does not distinguish a 404 from a network failure. A HEAD probe
cannot tell them apart reliably, and treating a timeout as "not published" is
how a transient hiccup silently swaps in main's scripts. More to the point, a
host that cannot reach raw.githubusercontent.com is about to fail on the release
tarball and `x-ui.sh` anyway — failing at the preflight, with nothing stopped or
removed yet, is strictly better than failing halfway through with the panel
already gone. The status code is printed, so `000` still tells you it was the
network rather than a missing file.

Only the files that are always fetched from GitHub are probed: `x-ui.sh`, plus
`x-ui.rc` on Alpine. The unit files are left out on purpose — they are fetched
only when the release tarball omits them, and older tags ship a single
`x-ui.service` inside the tarball, so probing `x-ui.service.debian`
unconditionally would break installs that currently succeed. Their existing
failure message now names the ref it tried instead.

Checked against live tags: `v2.3.5/x-ui.rc` → 404 → abort, `v2.8.4/x-ui.rc` →
200 → proceeds, `main` → skipped. `x-ui.sh` exists as far back as the v2.3.5
floor, so on non-Alpine the preflight never trips in practice.

Unrelated one-liner in the same commit: the `setup-fail2ban` probe now greps for
the dispatcher's case arm (`"setup-fail2ban")`) rather than any mention of the
string, which also matches a comment in `x-ui.sh`.

PR #6392, opened 2026-09-02
fix(x-ui.sh): put the fail2ban backend override in jail.d, not jail.conf

rokokol:
## Summary

Write the fail2ban `backend = systemd` override for Debian 12+ / Ubuntu 22.04+ to `/etc/fail2ban/jail.d/3x-ipl-backend.conf` instead of editing `/etc/fail2ban/jail.conf` in place with `sed`.

## Why

`create_iplimit_jails()` ran `sed -i '0,/action =/s/backend = auto/backend = systemd/' /etc/fail2ban/jail.conf`. `jail.conf` is the fail2ban package's conffile: on the next package upgrade dpkg either replaces the modified file (dropping the edit) or keeps the stale copy (dropping the package's updated jail defaults), depending on the conffile prompt. fail2ban documents that overrides belong in `jail.local` or `jail.d/*.conf`, which are read after `jail.conf` and left alone by upgrades; the script already owns `jail.d/3x-ipl.conf`.

The override is load-bearing on Debian 12: bookworm ships fail2ban 1.0.2-2 whose `paths-debian.conf` sets no `sshd_backend`, so the stock `[sshd]` jail inherits `[DEFAULT] backend` and, with no syslog files on a minimal install, fail2ban fails to start without it. Ubuntu 24.04 already ships `backend = systemd` in `jail.d/defaults-debian.conf`, so there the file is a harmless duplicate.

Two things kept from the old behaviour on purpose: the file is written only when `jail.conf` still carries the stock `backend = auto` in `[DEFAULT]` (the `sed` matched only that literal, so an operator's own value survived it and still does), and it is not removed by the "only remove IP Limit configurations" option (the `sed` edit was never reverted either, and deleting a `[DEFAULT]` override there would flip every inheriting jail back to `auto` on the restart in the same branch). The full `/etc/fail2ban` removal option deletes it with everything else.

The `[3x-ipl]` jail keeps its explicit `backend=auto`, so its own behaviour does not change. Existing installs that already have the `sed`-edited `jail.conf` are unaffected; the override file just makes the setting survive the next fail2ban upgrade.

## Type of change

- [x] Bug fix

## Areas affected

## How was this tested?

- `bash -n x-ui.sh`; `shellcheck -S warning` warning count unchanged against `main` (42).
- Reviewed against fail2ban's config read order (`jail.conf`, `jail.d/*.conf`, `jail.local`, `jail.d/*.local`); a `[DEFAULT]` section in `jail.d/3x-ipl-backend.conf` overrides the `backend = auto` default from `jail.conf` exactly as the `sed` did.
- The guard was run against upstream's stock `jail.conf` (writes the override) and a copy with `backend = polling` in `[DEFAULT]` (skips it).
- Not run end to end on a fresh server.

## Screenshots / recordings

N/A

## Breaking changes

None.

## Checklist

### DIFF
--- a/x-ui.sh
+++ b/x-ui.sh
@@ -2389,7 +2389,6 @@ remove_iplimit() {
             rm -f /etc/fail2ban/filter.d/3x-ipl.conf
             rm -f /etc/fail2ban/action.d/3x-ipl.conf
             rm -f /etc/fail2ban/jail.d/3x-ipl.conf
-            rm -f /etc/fail2ban/jail.d/3x-ipl-backend.conf
             if [[ $release == "alpine" ]]; then
                 rc-service fail2ban restart
             else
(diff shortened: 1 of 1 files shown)


rokokol:
Addressed the review in 6e4072d: the override is written only when jail.conf still has the stock `backend = auto` in [DEFAULT], as the sed matched (2), and the partial "only remove IP Limit" path no longer deletes it, matching the old never-reverted sed (1). On the unverified point: fail2ban 1.0.2 (Debian 12) ships no `sshd_backend` in paths-debian.conf, so [sshd] does inherit [DEFAULT] backend there. The new push re-runs the failed Windows job (3).

PR #6393, opened 2026-09-02
feat(release): publish SHA-256 sums and verify them in install.sh/update.sh

rokokol:
## Summary

Publish `<asset>.sha256` next to every release archive (Linux tarballs and the Windows zip, including the rolling `dev-latest` release) and verify it in `install.sh` and `update.sh` before extracting.

## Why

The installer and updater fetched the release archive and extracted it after checking only that the file is not empty (`[[ -s ... ]]`), and the release workflow published no checksums. A truncated download, a corrupted or swapped asset or a bad mirror was extracted and installed as root. #5396 added the same verification for the Xray-core archive the panel downloads; the panel's own archive was the remaining unverified download.

Scope of the guarantee, stated plainly: the sidecar comes from the same host over the same channel as the archive, so this catches corruption, truncation and a bad mirror; it does not defend against a party that controls the release or the TLS channel, which needs an out-of-band trust anchor (signed releases) and is out of scope here.

## Scope

- `release.yml`: the Linux `Package` step writes `sha256sum` output to `x-ui-linux-<platform>.tar.gz.sha256`; the Windows step writes the same format via `Get-FileHash`. Both files go to the artifacts and to the release (`file_glob: true`), and the `publish-dev` job uploads `dev-artifacts/*.sha256` too.
- `install.sh`: `verify_release_checksum()` downloads `${url}.sha256`, compares it with `sha256sum` of the archive, deletes the archive and exits on mismatch. Only an HTTP 404 for the sidecar (every release before this change, including `legacy_version` installs) is tolerated with a warning; any other failure to fetch it (5xx, connection reset, DNS) aborts and removes the archive, so a party that can swap the asset cannot skip the check by dropping the sidecar request. *(Maintainer note: happy to fail closed on 404 as well for tags newer than the first release that ships a sidecar, if you prefer.)*
- `update.sh`: same logic inline, using `_fail` on mismatch or fetch failure.

## Type of change

- [x] New feature
- [x] Build / CI / tooling

## Areas affected

## How was this tested?

- `verify_release_checksum()` exercised against a local HTTP server with a random archive: matching sidecar passes ("Checksum verified"), 404 sidecar warns and continues, unreachable sidecar host aborts (`HTTP 000`) and removes the download, tampered archive fails with the expected/actual hashes and removes the download.
- `bash -n` on both scripts; `shellcheck -S warning` warning counts unchanged against `main` (38 / 39); `release.yml` parses as YAML.
- The workflow change itself runs only on a tag push, so it is not exercised here; the `sha256sum`/`Get-FileHash` lines and the `file_glob` option of `svenstaro/upload-release-action` are the documented forms.

## Screenshots / recordings

N/A

## Breaking changes

None. Releases without a `.sha256` sidecar install as before, with a warning.

## Checklist


rokokol:
Follow-up in 0d568b0 for the two notes: only a 404 on the sidecar is tolerated now, any other fetch failure aborts and removes the archive; and the description states the scope honestly (corruption, truncation, bad mirror; not a MITM or supply-chain defence).

PR #6404, opened 2026-09-03
fix(geofile): verify downloaded geo databases against published digests

rokokol:
## Summary

`ServerService.UpdateGeofile` now verifies every downloaded geo database against the SHA-256 its upstream publishes, pins each download to one immutable release, installs an upstream's databases atomically, and no longer restarts Xray when nothing changed.

## Why

`UpdateGeofile` wrote whatever the three upstreams returned straight into the Xray asset folder with `os.Create` + `io.Copy` and no integrity check of any kind. Xray parses these databases when it builds its domain and IP matchers, so a corrupted or substituted file takes the core down at its next start.

The panel already does the right thing for the other artifact it downloads. `installXray` verifies the release archive against the SHA-256 published in its `.dgst` sidecar (`fetchXrayDigestSHA256` / `parseXrayDigestSHA256`, covered by `server_xray_checksum_test.go`). The geo databases were the only download in the panel that skipped that step, even though all three upstreams publish a `<asset>.sha256sum` next to every `.dat`:

- `Loyalsoldier/v2ray-rules-dat`
- `runetfreedom/russia-v2ray-rules-dat`
- `chocolate4u/Iran-v2ray-rules`

## What changed

**1. Verify before installing.** Each `.dat` is downloaded to a staging folder, its `.sha256sum` sidecar is fetched, and the digest is compared against the bytes that actually arrived.

**2. Pin the release.** The asset and its sidecar are two requests, so fetching both from `releases/latest/download/…` lets GitHub re-resolve `latest` in between. These upstreams publish several times a day — `202609022346`, `202609030908` and `202609031849` are three tags from a single day — so a release landing mid-batch would check release N+1's digest against release N's bytes and report a healthy upstream as tampered with. The tag is now resolved once per upstream from the redirect GitHub already returns, and both the asset and its digest are fetched from it. This is why an entry is modelled as repo + asset rather than an opaque URL.

**3. Atomic per upstream, not across all six.** A `geoip`/`geosite` pair from one release installs together or not at all. Coupling the three independent repositories buys no integrity and costs a lot: one transient 5xx would discard every verified download and re-fetch tens of megabytes on the next attempt. Each upstream now succeeds or aborts on its own and errors are collected, as the function did before this change.

**4. No restart on a no-op refresh.** The conditional `If-Modified-Since` GET was already there, but `RestartXrayService()` ran unconditionally afterwards, dropping every client connection even when all six upstreams answered `304`.

### The digest line is matched by base name

The upstreams disagree on the name field, so `sha256sum --check` semantics — look up the name written in the file — fail on a perfectly good download from chocolate4u:

```
0d5d2ba0…de382  geoip.dat          # Loyalsoldier, runetfreedom
cb2d2752…82ff   release/geoip.dat  # chocolate4u — the path from its own build
```

A leading `*`, `sha256sum`'s binary-mode marker, is stripped for the same reason: nothing stops an upstream regenerating with `-b`.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot

## How was this tested?

New suite in `internal/web/service/server_geofile_test.go`, stdlib `testing` with `httptest` upstreams that mimic GitHub's `releases/latest/download` → `releases/download/<tag>` redirect, and a throwaway DB. Each test was checked to fail without the fix by reverting the relevant part and watching it go red:

| Test | Fails without |
|---|---|
| `TestParseGeofileDigest` (`build-path`, `binary-mode-marker`) | the base-name match — a literal match reports "no SHA-256 entry", the chocolate4u failure |
| `TestParseGeofileDigest_Errors` | each row pins the concrete message, so the two error branches cannot swallow each other's cases |
| `TestUpdateGeofileRejectsDigestMismatch` | the digest comparison |
| `TestUpdateGeofileInstallsNeitherFileOfAFailedUpstream` | staging within an upstream; iteration is sorted so `geoip.dat` is provably downloaded before the corrupt `geosite.dat`, rather than relying on map order |
| `TestUpdateGeofileKeepsGoodUpstreamWhenAnotherFails` | per-upstream isolation — a broken repo must not discard a healthy one's verified file |
| `TestUpdateGeofileSurvivesReleaseRotation` | release pinning — with `latest` re-resolved, a healthy upstream is reported as "corrupted or has been tampered with" |
| `TestUpdateGeofileSkipsRestartWhenNotModified` | the restart gate |

```
go build ./...
go test ./... -shuffle=on -count=1
golangci-lint run          # 0 issues
gofumpt -l                 # clean
```

The `.sha256sum` asset names and the `304` behaviour were also checked against the live upstreams: all six sidecars exist at `releases/latest/download/`, GitHub honours `If-Modified-Since` through the release redirect and answers `304` with a zero-byte body, and the first hop of that redirect is what carries the concrete tag:

```
location: https://github.com/Loyalsoldier/v2ray-rules-dat/releases/download/202609022346/geoip.dat
```

Endpoint level: `POST /panel/api/server/updateGeofile` (and `/updateGeofile/:fileName`). The first run installs and restarts; an immediate second run gets `304` for all six, installs nothing, and does not restart the core.

## Screenshots / recordings

N/A — no UI change.

## Breaking changes

None for configuration or API shape. One behaviour change worth calling out: a failed download or digest check now aborts that **upstream's** update and returns an error, where before every failure was collected into a message and whatever succeeded was installed regardless. Files from the other upstreams are still installed, so a single flaky repository no longer costs the whole refresh — but a repository whose `geoip` verifies and whose `geosite` does not will now install neither, by design.

## Checklist

- [ ] For frontend changes: `npm run lint`, `npm run typecheck`, and `npm run build` pass. — N/A, no frontend change
- [ ] I updated the Wiki / README / API docs if user-facing behavior changed. — N/A, no endpoint or payload change

### DIFF
--- a/internal/web/service/server.go
+++ b/internal/web/service/server.go
@@ -21,6 +21,7 @@ import (
 	"net/url"
 	"os"
 	"os/exec"
+	"path"
 	"path/filepath"
 	"regexp"
 	"runtime"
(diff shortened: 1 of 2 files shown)


rokokol:
Thanks — all five points addressed in `806b74a1`.

- **Release pinning.** The tag is now resolved once per upstream from the redirect GitHub already returns, and both the asset and its `.sha256sum` are fetched from it. `TestUpdateGeofileSurvivesReleaseRotation` reproduces the old failure — a healthy upstream reported as tampered with — and goes red without the fix.
- **Atomicity scoped to one upstream.** A `geoip`/`geosite` pair still installs together, but a failing repository no longer discards another's verified downloads; errors are collected as the function did before this feature.
- **The all-or-none test no longer depends on map order.** Iteration is sorted, and the test asserts the good file was actually downloaded before the corrupt one.
- **Error rows pin their concrete message** instead of `err != nil`.
- **The three over-long comment blocks are trimmed** to the two-line limit.

PR description updated to match.

PR #6442, opened 2026-09-09
fix(tgbot): handle deleted inbounds on stale get_clients_for_* taps

BlindMaster24:
## Summary

Stale `get_clients_for_sub` / `get_clients_for_individual` / `get_clients_for_qr` inline-keyboard taps no longer crash the panel when the inbound is deleted in the window between the callback's two `GetInbound` reads.

## Why

The three `get_clients_for_*` callback cases fetched the inbound a **second** time with the error discarded:

```go
clientsKB, err := t.getInboundClientsFor(inboundIdInt, "client_sub_links")
if err != nil { ... }
inbound, _ := t.inboundService.GetInbound(inboundIdInt)   // error discarded
t.SendMsgToTgbot(chatId, t.I18nBot("tgbot.answers.chooseClient", "Inbound=="+inbound.Remark), clientsKB)
```

The first lookup (inside `getInboundClientsFor`) already guarded a missing inbound; the panic needed the TOCTOU window between the two reads — the row vanishing (delete + concurrent tap, or a transient DB failure) between keyboard build and `Remark` deref. Because `answerCallback` runs on a bare goroutine with no `recover()`, that panic killed the **entire panel process**, not just the bot.

## What changed

- **Single-read design** — `getInboundClientsFor` now receives the `*model.Inbound` row instead of re-reading the DB by ID; the between-reads window is removed rather than guarded. It is called from exactly one place, so the signature change breaks no other caller.
- **`chooseInboundClient` helper** — fetches the inbound once, logs the failure (`logger.Warning` — a sick DB and a deleted inbound no longer look identical to an operator), answers a `getInboundsFailed` callback, and builds the keyboard. The three duplicated case bodies collapse to a one-liner each.
- **`recover()` barrier on all three handler paths** — commands and callbacks run through `runBotHandler` (worker slot + `recoverBotPanic`), and the `th.AnyMessage()` free-text state machine (which runs inline on telego's dispatch goroutine, with no library-side recovery in telego v1.12.1) defers `recoverBotPanic` directly. No bot handler panic can take down the process; it logs via `logger.Error` instead.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [x] Telegram bot

## How was this tested?

- `TestChooseInboundClientStaleInbound` — drives `chooseInboundClient` against a test DB where inbound 42 does not exist, and **asserts `answerCallbackQuery` was called exactly once** (the error answer) via a counting mock Telegram API. Removing the guard fails the test with the nil panic itself; dropping only the `sendCallbackAnswerTgBot` call fails the count assertion.
- `TestGetInboundClientsForUsesProvidedInbound` — pins that the keyboard builder consumes the caller-provided row (a reintroduced second DB read reopens the stale-row window). Asserts the built keyboard button text.
- `TestRunBotHandlerRecoversPanic` — asserts a panicking handler is contained by `runBotHandler`; the panic escapes and fails the test without the `recover()`.
- Mock pattern follows `tgbot_send_test.go` (`httptest` + `telego.WithAPIServer`), throwaway SQLite via `database.InitDB(t.TempDir())`, std-lib only. gofmt-clean.
- Full `go test ./...` requires GCC (CGo/SQLite) which is available on CI; the local Windows box has no C compiler.

## Screenshots / recordings

N/A

## Breaking changes

None — the happy path is unchanged; only the stale-tap path changed from a potential process crash to an answered error callback, and handler panics are now logged instead of fatal.

## Checklist

- [x] I added or updated tests for the new behavior.
- [ ] `go build ./...` and the test suite pass locally. (requires GCC/MinGW — CI will verify)
- [ ] For frontend changes: npm run lint, npm run typecheck, and npm run build pass.

### DIFF
--- a/internal/web/service/tgbot/tgbot_inbound.go
+++ b/internal/web/service/tgbot/tgbot_inbound.go
@@ -109,12 +109,7 @@ func (t *Tgbot) getInboundsFor(nextAction string) (*telego.InlineKeyboardMarkup,
 }
 
 // getInboundClientsFor lists clients of an inbound with a specific action prefix to be appended with email
-func (t *Tgbot) getInboundClientsFor(inboundID int, action string) (*telego.InlineKeyboardMarkup, error) {
-	inbound, err := t.inboundService.GetInbound(inboundID)
-	if err != nil {
-		logger.Warning("getInboundClientsFor run failed:", err)
-		return nil, errors.New(t.I18nBot("tgbot.answers.getInboundsFailed"))
-	}
+func (t *Tgbot) getInboundClientsFor(inbound *model.Inbound, action string) (*telego.InlineKeyboardMarkup, error) {
 	clients, err := t.inboundService.GetClients(inbound)
 	var buttons []telego.InlineKeyboardButton
 
--- a/internal/web/service/tgbot/tgbot_router.go
+++ b/internal/web/service/tgbot/tgbot_router.go
@@ -16,6 +16,40 @@ import (
 	tu "github.com/mymmrac/telego/telegoutil"
 )
 
+// recoverBotPanic must be deferred by every bot handler entry point: telego's
+// dispatch has no recovery of its own, so one bad update would kill the panel.
+func recoverBotPanic() {
+	if r := recover(); r != nil {
+		logger.Error("Recovered panic in Telegram bot handler:", r)
+	}
+}
+
+// runBotHandler runs a bot handler on a worker slot and recovers panics: a bad
+// callback must not take down the whole panel, the way a cron panic would not.
+func runBotHandler(fn func()) {
+	messageWorkerPool <- struct{}{}
+	defer func() { <-messageWorkerPool }()
+	defer recoverBotPanic()
+	fn()
+}
+
+// chooseInboundClient fetches the inbound once and reuses the row: the inline
+// keyboard outlives the inbound, so a stale tap must answer an error, not panic.
+func (t *Tgbot) chooseInboundClient(callbackQuery *telego.CallbackQuery, chatId int64, inboundID int, action string) {
+	inbound, err := t.inboundService.GetInbound(inboundID)
+	if err != nil {
+		logger.Warning("chooseInboundClient GetInbound failed:", err)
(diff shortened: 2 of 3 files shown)


BlindMaster24:
Thanks for the careful review — all three actionable findings are addressed in 5742e240:

- **LOW 1 (discarded `GetInbound` error):** `chooseInboundClient` now logs `logger.Warning("chooseInboundClient GetInbound failed:", err)` before answering the error callback, so a sick DB is distinguishable from a deleted inbound.
- **LOW 2 (narrative overstated the reachable panic):** commit message and PR body rewritten — the panic required the TOCTOU window between the two `GetInbound` reads, which is exactly what the single-read design removes. PR body updated to describe that precisely.
- **LOW 3 (assertion-free test):** `TestChooseInboundClientStaleInbound` now uses a counting mock and asserts `answerCallbackQuery` was called exactly once — dropping only the answer call now fails the count, emptying the helper fails both the count and the no-panic pin.

The two pre-existing findings:
- **Draft-globals race (MEDIUM):** agreed it's real and unsynchronized; left out deliberately to keep this PR scoped to the panic fix — happy to take it as a follow-up.
- **Unrecovered `th.AnyMessage()` handler (LOW):** also addressed in the same push — the free-text handler now defers `recoverBotPanic()` directly (kept synchronous to preserve message ordering for the user state machine), so all three `OnReceive` paths are covered and the recover logic is shared between the worker-pool and inline paths.

BlindMaster24:
One more from the pre-existing list while at it (c3fce6d5): the two remaining inline handlers — `th.TextEqual(closeKeyboard)` and the `th.AnyCommand` dispatcher wrapper — now also defer `recoverBotPanic()`. All four `OnReceive` handler paths are covered; the heavy command/callback bodies still run through `runBotHandler` (worker slot + recover), the inline parts just get the barrier.

PR #6448, opened 2026-09-10
fix(frontproxy): restart the reverse proxy on a panel-only restart

kuzzrus:
## Summary
- `StopPanelOnly`/`StartPanelOnly` (the SIGHUP handler behind the "Restart Panel" button) never stopped `frontproxy.GetManager()` — it lived inside `stop()`'s `stopXray`-gated sidecar block, so a panel-only restart skipped it entirely.
- `Manager.Start()` then found its own still-running server and silently no-op'd, so a changed reverse-proxy domain, cert mode, port, or listen address could never actually take effect from the UI — only a real process restart (`systemctl restart`) rebuilds the listener.
- Moved the `StopAll()` call out of the gated block so it pairs with `start()`'s already-unconditional `FrontProxyService.AutoStart()` — the door now rebuilds with current settings on every restart, panel-only included. The other sidecars (Tor/AmneziaWG/MTProto/Psiphon/AdGuard) are untouched and still correctly ride through a panel-only restart undisturbed, since they're unrelated to panel settings.

Found live while testing the reverse-proxy's `auto` TLS mode: after changing the configured domain and clicking "Restart Panel," the certificate status never updated — confirmed via the SIGHUP handler in `main.go` and `Manager.Start()`'s own "already running, no-op" guard that this restart path genuinely never touches frontproxy at all.

## Test plan
- [x] `gofmt -l` clean on the changed file
- [ ] `go build ./...` (blocked locally by an unrelated pre-existing CGO/sqlite3 driver mismatch on this Windows machine — same known limitation as every other DB-adjacent change this fork has shipped; relying on CI)
- [ ] Live-verify on a real box: change the reverse-proxy domain/cert mode, click "Restart Panel" (not a full process restart), confirm the certificate is re-issued for the new domain and the panel/subscription paths still work through the door afterward


kuzzrus:
Opened against the wrong repository by mistake (this fixes an internal/frontproxy bug that only exists in a downstream fork, kuzzrus/3x-ui-awg -- that package isn't part of this repo). Closing and re-opening in the right place, sorry for the noise.

PR #6461, opened 2026-09-11
fix(amneziawg): honor inbound listen when binding UDP socket

### LINKED ISSUE #6367: [Bug]: AmneziaWG inbound ignores the "listen" field and always binds all addresses
### Describe the bug

An inbound with `protocol: amneziawg` and a non-empty `listen` value binds to all addresses
instead of the configured one. Inbounds backed by Xray honour `listen` normally on the same
panel.

The panel stores the value correctly — `GET /panel/api/inbounds/get/:id` returns
`"listen": "SECONDARY_IP"` — but the embedded device listens on the wildcard address.

On a host with two public IPv4 addresses this is not cosmetic. Because the socket is bound
to the wildcard address, the source address of every reply is chosen by the routing table
instead of the address the request arrived on. Clients that connect to the secondary IP get
answers from the primary one, discard them, and never complete a handshake.

1. Have a host with two public IPv4 addresses on one interface (`PRIMARY_IP` as the primary
   address and the default route source, `SECONDARY_IP` as an additional `/32`).
2. Create an inbound with `protocol: amneziawg`, `port: 1235`, `listen: SECONDARY_IP`.
3. Add at least one client (the device does not open its socket until an inbound has peers).
4. `ss -lunp | grep 1235` — the socket is bound to `0.0.0.0` and `[::]`, not to `SECONDARY_IP`.
5. Connect a client to `SECONDARY_IP:1235` and capture traffic on the server.

### Expected behavior

An AmneziaWG inbound with a non-empty `listen` binds only to that address, the same way
Xray-backed inbounds do, and replies leave from the address the client connected to.

If per-address binding is not supported by the embedded device, the field should be rejected
on save or documented as ignored for this protocol, so that multi-IP setups do not fail
silently.

### Actual behavior

The socket is bound to all addresses, replies are sourced from the primary IP, and the
handshake never completes. The client retries every ~5 seconds indefinitely. Nothing is
logged by the panel.

Workaround that makes the second inbound usable — note that `SNAT` alone is not enough,
because the incoming packet has already created a conntrack entry whose reply tuple is
exactly what `SNAT` produces, so the translated packet clashes on insert and is dropped
silently. The incoming flow has to be left untracked as well:

```bash
iptables -t raw -A PREROUTING -p udp -d SECONDARY_IP --dport 1235 -j NOTRACK
iptables -t nat -A POSTROUTING -p udp --sport 1235 -o IFACE -j SNAT --to-source SECONDARY_IP
```

### Relevant logs

```shell
The panel logs nothing at all for this case. Evidence is on the socket and on the wire.

Configured value read back from the API:


"listen": "SECONDARY_IP", "port": 1235, "protocol": "amneziawg"


Actual binding:


# ss -lunp | grep 1235
UNCONN 0 0    0.0.0.0:1235    0.0.0.0:*   users:(("x-ui",pid=632053,fd=21))
UNCONN 0 0       [::]:1235       [::]:*   users:(("x-ui",pid=632053,fd=22))


Client connecting to the secondary IP, server answering from the primary one:


IP CLIENT_IP.1345 > SECONDARY_IP.1235: UDP, length 47
IP CLIENT_IP.1345 > SECONDARY_IP.1235: UDP, length 149
IP CLIENT_IP.1345 > SECONDARY_IP.1235: UDP, length 243
IP PRIMARY_IP.1235 > CLIENT_IP.1345: UDP, length 425
IP CLIENT_IP.1345 > SECONDARY_IP.1235: UDP, length 47
IP CLIENT_IP.1345 > SECONDARY_IP.1235: UDP, length 94
IP PRIMARY_IP.1235 > CLIENT_IP.1345: UDP, length 260
```

### Screenshots

_No response_

### 3x-ui version

3.7.0

### Xray-core version

26.7.28

install.sh script

### Operating system

Ubuntu 24.04 LTS, x86_64

Xray config generation

_No response_

None

### Before submitting

mrchatam:
## Summary
- AmneziaWG inbounds ignored `listen` and always bound `0.0.0.0`/`::` via `awgconn.NewDefaultBind()`, so multi-IP hosts replied from the primary address and clients connecting to a secondary IP never completed the handshake.
- Carry `Inbound.Listen` on `amneziawg.Instance`, open a `pinnedBind` for non-wildcard listen values (empty / `0.0.0.0` / `::` keep the default bind), and include listen in `addressFingerprint` so edits rebuild the Device instead of hot-applying an unchangeable Bind.

## Test plan
- [x] `go test ./internal/amneziawg/ ./internal/amneziawgnet/`
- [ ] Create AmneziaWG inbound with `listen: SECONDARY_IP`, add a client, confirm `ss -lunp` shows the socket bound to that address only
- [ ] Client handshake to `SECONDARY_IP:port` completes; replies source from `SECONDARY_IP`
- [ ] Empty / `0.0.0.0` listen still binds all addresses as before
- [ ] Editing listen rebuilds the device (fingerprint change)

Fixes #6367

### DIFF
--- a/internal/amneziawg/instance.go
+++ b/internal/amneziawg/instance.go
@@ -63,6 +63,7 @@ func InstanceFromInbound(ib *model.Inbound) (Instance, bool) {
 		Tag:                   ib.Tag,
 		InterfaceName:         interfaceNameForID(ib.Id),
 		ListenPort:            ib.Port,
+		Listen:                ib.Listen,
 		PrivateKey:            server.PrivateKey,
 		PublicKey:             server.PublicKey,
 		Address:               addresses,
--- a/internal/amneziawg/instance_test.go
+++ b/internal/amneziawg/instance_test.go
@@ -32,7 +32,7 @@ func TestInstanceFromInboundParsesEnabledPeers(t *testing.T) {
 		{Email: "c@x", Enable: true, PublicKey: "", AllowedIPs: []string{"10.8.1.4/32"}}, // no key: skipped
 		{Email: "d@x", Enable: true, PublicKey: "pubD", AllowedIPs: nil},                 // no address: skipped
 	})
-	ib := &model.Inbound{Id: 7, Tag: "awg-tag", Protocol: model.AmneziaWG, Port: 51820, Settings: settings}
+	ib := &model.Inbound{Id: 7, Tag: "awg-tag", Protocol: model.AmneziaWG, Port: 51820, Listen: "203.0.113.10", Settings: settings}
 
 	inst, ok := InstanceFromInbound(ib)
 	if !ok {
@@ -40,6 +40,9 @@ func TestInstanceFromInboundParsesEnabledPeers(t *testing.T) {
 	}
 	if inst.Id != 7 || inst.Tag != "awg-tag" || inst.ListenPort != 51820 {
 		t.Fatalf("instance identity not carried over: %+v", inst)
+	}
+	if inst.Listen != "203.0.113.10" {
+		t.Fatalf("Listen = %q, want inbound listen carried through", inst.Listen)
 	}
 	if inst.InterfaceName != "awg7" {
 		t.Fatalf("InterfaceName = %q, want awg7", inst.InterfaceName)
--- a/internal/amneziawg/types.go
+++ b/internal/amneziawg/types.go
@@ -75,8 +75,14 @@ type Instance struct {
 	Tag           string
 	InterfaceName string
 	ListenPort    int
-	PrivateKey    string
-	PublicKey     string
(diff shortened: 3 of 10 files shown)


mrchatam:
### Follow-up (1addc08)

- **HIGH:** unusable / non-IP `listen` no longer hard-fails AmneziaWG startup — falls back to dual-stack `StdNetBind` with a warning (probes bindability for concrete IPs).
- **MEDIUM:** `::0` / `[::0]` treated as wildcards (aligned with `isAnyListen`); bracketed IPv6 literals like `[::1]` parse as pinned addresses.
- Fingerprint normalizes wildcard / fallback listen so empty↔`0.0.0.0` does not rebuild the Device.
- Trimmed comment blocks to the 2-line cap.

PR #6462, opened 2026-09-11
fix(node): skip selected-inbound delete when import is pending

### LINKED ISSUE #6329: [Bug]: Selecting an existing node inbound in the central panel deletes it from the node
### Describe the bug

When multiple nodes are connected to a central 3x-ui panel, selecting an existing node inbound from the central panel unexpectedly deletes that inbound from the node.

The inbound is created directly through the node’s own panel and works correctly. However, when I open the **Nodes** section in the central panel and select that inbound so it can be managed or displayed among the central panel’s inbounds, the central panel removes the inbound from the node instead.

1. Connect one or more 3x-ui nodes to a central panel.
2. Open the panel of one of the connected nodes.
3. Create a new inbound directly on that node.
4. Verify that the inbound exists and works correctly.
5. Open the central panel.
6. Go to the **Nodes** section and select the affected node.
7. Select the newly created inbound so that it is included in the central panel’s inbound list.
8. Check the node’s own panel again.
9. The inbound has been deleted from the node.

### Expected behavior

Selecting an existing node inbound in the central panel should import, synchronize, or display it in the central panel without deleting it from the node.

### Actual behavior

The central panel deletes the selected inbound from the node instead of synchronizing or importing it.

### Relevant logs

```shell

```

### Screenshots

_No response_

### 3x-ui version

3.7.0

### Xray-core version

26.7.28

install.sh script

### Operating system

Ubuntu 24

Multi-node (sub-nodes)

_No response_

Yes — Cloudflare (proxied DNS)

### Before submitting

mrchatam:
## Summary
Fixes #6329 — selecting an existing node inbound in **Selected inbounds** mode no longer deletes it from the node.

In `ReconcileNode`, the selected-mode delete sweep treated a tag present in `Node.InboundTags` but missing a central `inbounds` row as a master-side deletion. That state is also how a *pending import* looks right after the operator selects a node-created inbound (reconcile runs before the snapshot merge that would adopt it).

## Change
- `DelInbound` records a short-lived per-node tag tombstone when a node-attached inbound is removed centrally.
- Selected-mode sweep only deletes a managed remote tag when that tombstone is present; otherwise it leaves the tag for snapshot import.
- Regression: `TestReconcileNode_SelectedModeSkipsPendingImport`; existing selected-mode delete tests now set a tombstone.

## Test plan
- [x] `go test ./internal/web/service/ -run TestReconcileNode_ `
- [ ] Manual: create inbound on node → select it in central panel Selected mode → confirm it remains on the node and appears centrally after traffic sync
- [ ] Manual: delete a selected node inbound from central while node is down → when node returns, confirm remote inbound is swept

### DIFF
--- a/internal/web/service/inbound.go
+++ b/internal/web/service/inbound.go
@@ -1394,6 +1394,11 @@ func (s *InboundService) DelInbound(id int) (bool, error) {
 	}); err != nil {
 		return needRestart, err
 	}
+	// Record deletion intent before the remote push so a selected-mode
+	// reconcile can finish the sweep if the node was unreachable (#6329).
+	if loadErr == nil && ib.NodeID != nil {
+		tombstoneNodeInboundTag(*ib.NodeID, ib.Tag)
+	}
 	if postCommitApply != nil {
 		postCommitApply()
 	}
--- a/internal/web/service/inbound.go
+++ b/internal/web/service/inbound.go
@@ -1388,14 +1388,16 @@ func (s *InboundService) DelInbound(id int) (bool, error) {
 			}
 		}
 		if loadErr == nil && ib.NodeID != nil {
+			if err := (&NodeService{}).RemoveInboundTagAllowedTx(tx, *ib.NodeID, ib.Tag); err != nil {
+				return err
+			}
 			return (&NodeService{}).MarkNodeDirtyTx(tx, *ib.NodeID)
 		}
 		return nil
 	}); err != nil {
 		return needRestart, err
 	}
-	// Record deletion intent before the remote push so a selected-mode
-	// reconcile can finish the sweep if the node was unreachable (#6329).
+	// Tombstone before remote push so selected-mode reconcile finishes offline deletes (#6329).
 	if loadErr == nil && ib.NodeID != nil {
 		tombstoneNodeInboundTag(*ib.NodeID, ib.Tag)
 	}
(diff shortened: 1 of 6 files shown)


mrchatam:
### Follow-up (bbdc01f0) — HIGH residual note

Hardened the selected-mode import/delete gate without a full durable redesign:

- **Pending import:** selected tags are still adopted while `config_dirty` (so import is not deferred past reconcile), and selected+missing without a tombstone is never swept.
- **Deletes:** `DelInbound` still tombstones the tag (so offline remote delete can finish) **and** removes it from `Node.InboundTags` so an expired/restart-lost tombstone cannot re-adopt the row into the central DB. Sweep deletes tombstoned tags even after deselection.
- **Tag rename / port edit:** old auto-tag is tombstoned + deselected so the orphan remote can be swept.

**Residual limitation:** delete tombstones remain in-process with a 90s TTL. If the panel restarts (or the node stays down past the TTL) before reconcile runs, the **remote** inbound may remain on the node as an unmanaged orphan — but it will **not** reappear as a central row (deselected from `InboundTags`). A durable pending-delete list would be needed to guarantee remote cleanup across restarts; out of scope for this small fix.

MHSanaei:
Thanks for taking on #6329. Closing this because the fix doesn't survive restarts: in selected mode the sweep can't tell a pending import from an inbound deleted while its node was offline unless that delete is recorded durably, and a 90-second in-memory tombstone isn't — after a panel restart or a node outage longer than that, the deleted inbound stays on the node unmanaged, as your follow-up notes. #6329 stays open for a durable approach: no sweep in selected mode with the tag dropped on delete, a persisted per-node list of deleted tags, or importing synchronously when the node is saved.

PR #6463, opened 2026-09-11
feat(inbound): excludeFromSub hides links without disabling

### LINKED ISSUE #6435: [Feature]: Hide an inbound from subscription output without disabling it
### Problem

Subscription output emits one link per `(inbound, client)` pair. In a multi-node
high-availability setup this produces duplicate, functionally identical endpoints that
clients do not need to see.

### Scenario

A relay sits in front of two exit nodes. Failover happens at the relay (it rewrites its
forwarding target when an exit becomes unreachable), not in the client — because in a
whitelisted network the client has exactly one reachable entry point and cannot choose.

For that failover to be transparent, both exits must carry an **identical REALITY identity**:
same private/public key, same shortIds, same serverNames/target, same client UUIDs. Otherwise
switching breaks authentication instead of preserving it.

Since an inbound belongs to a single node (`node_id`), covering two nodes requires two inbounds
per identity. With two identities that is four REALITY inbounds, and the subscription therefore
contains four REALITY links where two would be enough — the node clones are pure duplicates from
the client's point of view.

### Why the existing options don't help

- **`Enable = false`** hides the inbound from the subscription, but also stops it serving
  traffic. Clients must stay attached to the clone for authentication to work after failover, so
  this is not usable.
- **Hosts** have `excludeFromSubTypes` and an `is_hidden` column, but those govern the extra
  links a host contributes, not the inbound's own link.
- The `Inbound` model has `Enable`, `SubSortIndex`, `ShareAddrStrategy`, `ShareAddr` and
  `DisableFlow` — there is no equivalent visibility switch.

### Proposal

A per-inbound boolean, e.g. `hideFromSub` / `excludeFromSub`, that keeps the inbound fully
operational (Xray config, client auth, traffic accounting) while omitting its links from
subscription output.

This mirrors what hosts already have, and would let operators keep redundant backing inbounds
for failover without cluttering what end users see.

### Workarounds considered

- `SubSortIndex` only reorders links, it cannot remove them.
- Managing the standby node's Xray config outside the panel would hide the links but gives up
  central management, node sync and traffic accounting — which is the reason to use nodes at all.

### Version

Panel 3.7.0, Xray 26.7.28, SQLite backend, one master + one node.

mrchatam:
## Summary
- Adds per-inbound `excludeFromSub` so operators can omit an inbound from subscription output without disabling it (Xray, auth, and accounting stay active).
- Gates `getInboundsBySubId` with `exclude_from_sub = false`, persists/syncs the flag through inbound CRUD and node wiring, and exposes a switch in the inbound form.
- Includes migration, OpenAPI/generated schema updates, translations, and focused backend/frontend tests.

Fixes #6435

## Test plan
- [ ] Create two enabled inbounds sharing one client `subId`; enable Exclude from subscriptions on one; confirm raw/JSON/Clash subscription only lists the other.
- [ ] Confirm the excluded inbound still accepts traffic and appears in panel traffic accounting.
- [ ] Edit an inbound to toggle the flag on/off and verify persistence after reload.
- [ ] On a node-hosted inbound, toggle the flag and verify it syncs through remote wire/update paths.


mrchatam:
### Follow-up (4a61e7bd)

- **golangci:** gofumpt on `ExcludeFromSub` in `model.go`.
- **MEDIUM:** synced `docs/public/openapi.json` with `frontend/public/openapi.json`.
- **MEDIUM (old node):** `exclude_from_sub` is no longer mirrored from the node snapshot (master-authored); avoids older builds resetting the flag to `false` on heartbeat. Still pushed via `wireInbound`.
- **LOW:** migration test now seeds a legacy `inbounds` table without the column and asserts `AddColumn` + default `false`.

PR #6465, opened 2026-09-11
fix(inbounds): allow negative subSortIndex for subscription order

### LINKED ISSUE #5332: [Feature]: Allow to use negative sorting order (index) in inbounds

I have servers that stand out for their stability and speed. I’d like to set them as the primary inbound in the subscription. Currently, to do this, I have to assign a higher index to each slower inbounds. I’d prefer to simply assign a negative index to the primary inbound server once, and leave the rest at their default values.

I already did it in https://github.com/MHSanaei/3x-ui/pull/5214. 

Allow to users sets a negative sorting index

_No response_

### Concrete use case

_No response_

_No response_

_No response_

### Are you willing to help?

- [x] I'd like to implement this feature myself and open a PR.

### Before submitting

mrchatam:
## Summary
Fixes #5332

`subSortIndex` already drives subscription output order (`ORDER BY sub_sort_index ASC`), but every layer rejected values below 1, so primary inbounds could not sort first without bumping every other inbound.

This change keeps omitted/`0` normalizing to `1` for backward compatibility, while preserving explicitly set **negative** indices end-to-end (model validation, API binding, service normalize, startup DB lift, form schema/UI, and generated OpenAPI).

## Test plan
- [x] `go test ./internal/web/service/ -run 'SubSort|NormalizeSubSort'`
- [x] Frontend `inbound-form-adapter` vitest (including negative preservation)
- [ ] Manually set inbound `subSortIndex` to `-1` and confirm it sorts ahead of default `1` in subscription output


mrchatam:
**Note on MEDIUM (node reject of negatives):** Leaving as-is for now.

A negative `subSortIndex` stored on the master will fail validation on any not-yet-upgraded sub-node (`gte=1` / `min=1` still present there), which blocks full inbound pushes until the node is upgraded or the index is put back to ≥ 1. A safe mitigation needs an explicit master↔node capability / version gate (or clamping only on the wire for old nodes) — that is a design choice beyond this CI follow-up. Operators rolling negative indices should upgrade nodes first (or keep indices ≥ 1 until then).

Issue #6342, opened 2026-08-27
[Feature]: Expose subJsonRules as raw JSON and allow customizable subscription DNS


Thedrezik:
### Is your feature request related to a problem?

I'm always frustrated when trying to distribute advanced routing rules to my clients via the subscription link due to two hardcoded limitations:

1. **Hardcoded DNS:** The subscription JSON generation hardcodes `8.8.8.8` as the DNS server for clients. Currently, the only way to change this is to manually patch the binary using `sed -i 's/8.8.8.8/1.1.1.1/g' /usr/local/x-ui/x-ui`.
2. **Overwritten Routing Rules:** I need to pass complex routing structures to the clients (e.g., blocking `udp/443` to force HTTP/2 over TCP for XTLS-Vision stability, blocking `bittorrent`, and routing custom domain arrays). The current Web UI only supports simple string tags. If I manually inject a complex JSON array into the SQLite database (`subJsonRules` key), the Web UI completely overwrites and deletes my advanced rules the next time I click "Save" in the Panel Settings.

### Describe the solution you'd like

1. **Custom DNS Input:** Add an input field in the Web UI (Panel Settings -> Subscription) to define the default DNS servers for the generated client config, replacing the hardcoded `8.8.8.8`.
2. **Raw JSON Input for subJsonRules:** Add a raw text area for `subJsonRules` in the Subscription settings page, exactly like the existing `xrayTemplateConfig` text area. This would allow advanced users to inject complete, complex JSON routing arrays directly into the database without the UI parsing logic breaking or resetting them upon saving.

### Alternatives you've considered

1. Using the Web UI tags (comma-separated domains) — this works for basic domains but fails if I need to pass specific network parameters, ports, or protocols (like blocking `bittorrent` or `udp/443`).
2. Injecting raw JSON directly into the SQLite database (`x-ui.db`) using the CLI. This works temporarily, but as soon as any setting is changed and saved via the Web UI, the custom JSON is overwritten and deleted.
3. For the DNS issue, I'm currently using a `sed` command to patch the compiled binary to replace `8.8.8.8` with `1.1.1.1`, which is highly unideal and breaks on every panel update.

### Concrete use case

I manage a server for several non-technical users using clients like v2rayNG and NekoBox. I use XTLS-Vision, which requires UDP/443 (QUIC) to be blocked on the client side for maximum stability and to bypass ISP throttling. I also need to route a specific list of 50+ regional domains directly and block torrents locally. 

[...]

Issue #4850, opened 2026-06-03
[Feature]: Add panel settings to customize DNS, queryStrategy, and domainStrategy for generated client subscription configs


rusik-afk:
### Is your feature request related to a problem?

Yes. As discussed in issue #4848, the client subscription JSON config endpoint (`/sub/json/<subId>`) currently relies on a hardcoded embedded template (`sub/default.json`). 

Because of this, parameters like `dns.servers` (Google DNS), `dns.queryStrategy` (`UseIP`), and `routing.domainStrategy` (`AsIs`) are hardcoded for all clients. Administrators cannot enforce privacy-centric DNS (e.g., Cloudflare) or specific routing strategies globally from the panel. This requires manually instructing every end-user to configure these parameters inside their respective client applications (Hiddify, Happ, etc.), which is inefficient and error-prone for non-technical users.

### Describe the solution you'd like

I would like the panel to expose new configuration fields in **Panel Settings -> Subscription Settings -> JSON Settings** (similar to how `subJsonRules` or `subJsonFragment` currently work).

Ideally, add the following setting keys to the UI:
- `subJsonDNSServers`: To override the default `"8.8.8.8"` with custom upstream DNS arrays.
- `subJsonDNSQueryStrategy`: A dropdown or string field to override `"UseIP"` (e.g., with `"UseIPv4"`).
- `subJsonDomainStrategy`: A dropdown or string field to override `"AsIs"` (e.g., with `"IPIfNonMatch"`).

When generating the subscription via the `/sub/json/` endpoint, the panel should check these fields and inject them into the client config instead of using the hardcoded values from `sub/default.json`.

### Alternatives you've considered

1. Manually configuring every single client device: Does not scale well when managing nodes for a group of friends or clients.
2. Building the panel from source after patching `sub/default.json`: This works but creates immense maintenance overhead, as every official panel update forces a manual re-compile of the binary.

### Concrete use case

I manage a server providing Xray access to a close group of friends/clients. I need all generated configs to seamlessly use Cloudflare DNS (`1.1.1.1`), utilize the `IPIfNonMatch` domain strategy, and strictly enforce the `UseIPv4` query/domain strategy to prevent connection issues caused by unstable IPv6 routing, DNS leaks, or geofence blocks. Having this exposed in the Web UI would allow me to manage this centrally and instantly update all clients upon subscription renewal.

### Mockups, screenshots, or examples

[...]

PR #6472, opened 2026-09-11
feat(sub): make JSON subscription DNS configurable

mrchatam:
## Summary
- Adds panel settings `subJsonDNSServers` (default `8.8.8.8`) and `subJsonDNSQueryStrategy` (default `UseIP`) so generated JSON subscriptions no longer need a binary sed patch for client DNS.
- Plumbs settings through entity → SettingService → subscription server → `SubJsonService.SetDNSConfig`, with a DNS tab under Settings → Sub Formats and i18n coverage.
- Preserves current behavior when unset; invalid query strategies fall back to `UseIP`. Baked Happ/INCY routing profiles still own their DNS subtree.

## Scope / non-goals
- **Addresses #6342** for the DNS slice only (not the raw `subJsonRules` textarea).
- **Overlaps #4850** for `subJsonDNSServers` + `subJsonDNSQueryStrategy`.
- Follow-up: `subJsonDomainStrategy`, and the raw `subJsonRules` editor requested in #6342.

## Test plan
- [x] `go test ./internal/sub/ ./internal/web/service/`
- [ ] Panel UI: Settings → Sub Formats → DNS — set `1.1.1.1`, `UseIPv4`, save, restart/reload sub server
- [ ] Fetch `/json/<subId>` and confirm `dns.servers` / `queryStrategy` match
- [ ] Clear fields / leave defaults and confirm `8.8.8.8` + `UseIP`


DIMFLIX:
Heads up, and a note for the maintainer: #6485 covers the same feature and was opened after this one. Both add SubJsonService.SetDNSConfig on the same receiver and both create internal/sub/json_dns_test.go, so one of them will need a rebase or a close. Comparison is in #6485.

While testing your patch I hit this: parseSubJsonDNSServers splits on spaces, quotes and brackets, so pasting the object form {"servers": ["1.1.1.1"]} gives literal entries {"servers": and } in the servers list, and those go into every subscription as nameserver addresses. Only the array form works. Worth either rejecting braces or validating the addresses before they land in the dns block.

PR #6485, opened 2026-09-12
feat(sub): let the panel set the JSON subscription DNS servers

DIMFLIX:
## Summary

Adds the `subJsonDns` panel setting: an operator-defined list of DNS servers — or a full xray `dns` block — baked into every JSON subscription document.

## Why

A baked routing profile (#6402) carries only the DNS its preset defines, so an operator who wants their own resolvers (a self-hosted DoH, several upstreams, per-domain scoping) has to override the whole profile or patch the subscription behind a proxy. This makes it a first-class panel setting. It has the final word over the profile's own DNS while leaving the profile's routing rules intact, and it reaches per-inbound, balancer and info-node documents alike.

Refs #6402.

## Type of change

- [x] New feature

## Areas affected

## How was this tested?

Backend, on a live panel with xray-core 26.3.27:

- Set the value through `POST /panel/api/setting/update`, restarted the panel, fetched `/json/<sub-id>` over HTTP and validated the emitted client config with `xray run -test` → `Configuration OK`.
- Covered the array form (2–3 servers, including an object entry with `domains`/`expectIPs`), the full object form (`tag`, `queryStrategy`, `hosts`, `serveStale`), the override over a routing profile (rules from the profile survive, DNS comes from the panel), balancer documents, and the info-node document.
- Degradation: `[]`, a server without `address`, and a block with a broken field type (`"hosts": 5`) are rejected when the settings are saved with xray's own error text, and if such a value reaches the database some other way the sub server logs a warning and keeps the template resolver.

Tests:

- `make test-go` green, including the new `internal/xray/dnsconf`, `internal/sub` and `internal/web/service` cases (each fails without the fix).
- Frontend: `npm run typecheck`, `npm run lint`, `npm run format:check`, unit tests and `npm run build` pass.

## Screenshots / recordings

image

## Breaking changes

None. Empty value keeps the previous behavior (template resolver, or the routing profile's DNS when one is set).

## Checklist

      (OpenAPI + generated frontend artifacts regenerated; the docs site does not document the sibling routing-profile setting, so no docs page was touched.)


DIMFLIX:
Just noticed #6472 while going through open PRs, same feature and it was opened before mine. Should have checked first.

They can't both be merged as they are: both add SubJsonService.SetDNSConfig (different signatures) and both create internal/sub/json_dns_test.go, so whoever goes second has to rebase anyway. If the smaller one fits the project better, I'll close this or trim it down to the full dns block on top of those settings.

Difference, in case it matters for picking:

#6472 uses two settings (subJsonDNSServers, subJsonDNSQueryStrategy) and a plain list of addresses. Simpler to type, and when a routing profile is set it keeps the profile's DNS.
Here there's one setting (subJsonDns) that takes either the whole dns block or just an array of servers, so hosts, queryStrategy and per-server domains/expectIPs work too. The per-server domains are the main reason for the object form: they let you split domestic and remote resolvers inside the subscription itself, without having to set up a routing profile for it. Everything is validated with xray's own conf.DNSConfig, and the same check runs when settings are saved and when the subscription is served. When a routing profile is set, the setting here wins over the profile's DNS while the profile's routing rules stay.

UI is a textarea in the JSON card, no separate tab. If this direction is fine, adding the DNS tab and a queryStrategy dropdown afterwards is easy, so users don't lose that.

PR #6489, opened 2026-09-12
fix(tgbot): require client ownership for non-admin link callbacks

BlindMaster24:
## Summary

The gate also has to be able to read which client the button names: payloads
past 64 characters arrive as a hash, so the raw callback data is now decoded
before it is judged.

A non-admin could pull **any** client's subscription links through the bot's link
buttons. The three callbacks that carry an email (`client_sub_links <email>`,
`client_individual_links <email>`, `client_qr_links <email>`) did not re-check
that the client belongs to the Telegram account pressing the button.

## Why

The bot answers callbacks for two audiences: admins (who own every client) and
clients (who must only ever see their own). For the second audience the email in
the callback data was trusted as-is, so the keyboard — not the caller — decided
which client was served. The sibling callback `client_traffic` already does it
correctly: it resolves the caller's clients from `callbackQuery.From.ID`
(`getClientUsage` -> `GetClientTrafficTgBot(tgUserID)`, `tgbot_client.go:557`),
and the guard's own doc comment claimed the same for all of them
(`isClientSelfCallback`, `tgbot_router.go:1331`).

The intent was there; only the check was missing.

### Why the data cannot authorise itself

An inline keyboard outlives the chat it was sent to, so "the button exists"
is not proof of access:

1. **A group chat.** The admin adds the bot to a group that also has clients, and
   taps *All clients* -> inbound -> client. The keyboard is delivered to the
   group, so every member can press *Subscription links*. The callback arrives
   with the member's `From.ID`, `isAdmin` is false, and before this change the
   member was served the client's sub URL and then, from that message, the
   *Individual links* / *QR code* buttons — which contain the `vless://`
   / `trojan://` URI with the client's UUID or password.
2. **A forwarded card.** Forwarding a client card keeps its inline keyboard.
   The recipient taps it and receives a working configuration for a client they
   were never granted.
3. **A revoked client.** Clearing a client's Telegram ID (`tg-user` ->
   *Remove*) is how an operator ends that client's bot access, but the card
   already sitting in the client's chat keeps working: nothing on the callback
   path re-read the client's `tgId`, so the ex-client could keep minting fresh
   links and QR codes for the config after being cut off.

Scenario 3 is the one that also matters for an operator who never uses groups:
revocation through the bot did not actually revoke anything that had already
been delivered.

### The payload has to be readable before it can be judged

`encodeQuery` replaces any callback payload longer than 64 characters with its
MD5 hash, and this gate tested the *raw* data. For a client whose email is long
enough the gate therefore saw a bare hash:

- `client_individual_links ` is 24 characters, so it hashes from a 41-character
  email
- `client_sub_links ` and `client_qr_links ` are 17, so from a 48-character one

Those clients' own link buttons — the ones the bot builds for them — arrived as a
hash, matched nothing, and were dropped in silence: the ownership guard never
ran, and the client pressed a button that did nothing. The raw data is now
decoded before the gate, using the same decode the admin path has always
performed.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Xray config generation
- [x] Subscription (share links / Clash / JSON) — the bot's copy of the link flow
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [x] Telegram bot

## How was this tested?

`internal/web/service/tgbot/tgbot_client_links_authz_test.go`, three cases against
a mocked Telegram API and a throwaway SQLite DB seeded with one inbound whose
settings bind `owner@x` to Telegram id 4242:

- `TestClientLinkCallbackRefusesForeignClient` — the owner's id taps
  `client_sub_links someone-else@x`. Asserts **no** `sendMessage` was sent and
  that the tap was answered exactly once. Without the ownership check the
  keyboard is served, so this test fails on the old code.
- `TestClientLinkCallbackServesOwnClient` — the same id taps its own email.
  Asserts one `sendMessage` and no refusal, so the guard cannot pass by simply
  blocking everyone.
- `TestHashedLinkCallbackServesOwnClient` — the same seeded client with a
  54-character address, whose button data `encodeQuery` turns into a hash. The
  owner taps that hash and must still be served. On the old code the gate never
  runs and nothing is sent, so this fails before the fix as well.

I could not build or run the tests on my machine: it has no C compiler at all,
and the CGo SQLite driver (`mattn/go-sqlite3`) makes every package that imports
`internal/database` — including this one — fail to compile without one
(`CONTRIBUTING.md` covers the MinGW-w64 install for Windows). `gofmt` is clean;
CI is the build here.

## Screenshots / recordings

N/A — bot callbacks, no UI.

## Breaking changes

None for admins. A non-admin who legitimately owns the client is unaffected
(`TestClientLinkCallbackServesOwnClient`). A non-admin whose client has no
`client_traffics` row yet is refused by the same lookup the *Client usage*
button already relies on.

## Checklist

- [ ] I tested the change locally and confirmed the described behavior. — no C compiler on this machine, see above; the two tests cover it and CI runs them.


BlindMaster24:
Review addressed in `7a058f9b`.

The MEDIUM (pre-existing) is taken, because it bounds this guard rather than sitting beside it.

**`encodeQuery` hides the email from the gate.** `encodeQuery` replaces any payload past 64 characters with its MD5 hash, and the gate tested the raw data — so for an email long enough (`client_individual_links ` is 24 chars, i.e. a 41-character address; `client_sub_links ` / `client_qr_links ` are 17, i.e. 48) the non-admin path saw a bare hash. Those clients' own link buttons matched nothing and were dropped in silence: the new ownership check never executed for exactly the clients whose buttons the bot builds for them. The raw data is now decoded before the gate, with the same decode the admin path has always performed.

- New test `TestHashedLinkCallbackServesOwnClient`: a 54-character address, taps the hashed payload `encodeQuery` produces. One `sendMessage` after the change, zero before it.
- The refusal path is unchanged (a foreign email is still answered with `errorOperation`), so `TestClientLinkCallbackRefusesForeignClient` keeps its meaning.

The review covered head `7540c1fa`, which no longer exists — the new code needs a fresh look (`@claude review` if that is the way to re-run it).

PR #6340, opened 2026-08-27
fix(tgbot): suppress 'message not modified' log spam from refresh buttons

BlindMaster24:
## Summary

Suppress noisy "message is not modified" warnings in Telegram bot edit calls when users click Refresh buttons.

## Why

When users click 🔄 Refresh in the Telegram bot (usage, client info, IP logs, onlines), `editMessageText` and `editMessageReplyMarkup` are always called even when the data hasn't changed. Telegram returns a 400 error (`message is not modified`), which was logged as `Warning` — cluttering the logs on every single refresh click.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring (no behavior change)
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Frontend (UI / panel pages)
- [ ] Backend (API endpoints, login, settings)
- [ ] Xray config generation
- [ ] Subscription (share links / Clash / JSON)
- [ ] Statistics / traffic counters
- [ ] Database / migrations
- [ ] Install / upgrade script
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [x] Telegram bot

## How was this tested?

- Unit test `TestIsTelegramNotModifiedError` verifies the helper correctly identifies "not modified" and "No fields to modify" errors while rejecting unrelated errors.
- `gofmt` passes. Full `go test` requires GCC (CGo/SQLite) which is available on CI.

## Screenshots / recordings

N/A

## Breaking changes

None

## Checklist

- [x] I tested the change locally and confirmed the described behavior.
- [x] I added or updated tests for the new behavior.
- [ ] `go build ./...` and the test suite pass locally. *(requires GCC/MinGW — CI will verify)*
- [ ] For frontend changes: `npm run lint`, `npm run typecheck`, and `npm run build` pass.
- [ ] I updated the Wiki / README / API docs if user-facing behavior changed.
- [x] My commits follow the project's existing message style.
- [x] I have no unrelated changes mixed into this PR.

PR #6490, opened 2026-09-12
fix(tgbot): send the admin traffic reports as one message

BlindMaster24:
## Summary

Two details the batching must not lose are part of this change: the reset report
still answers when the panel has no clients, and both reports are HTML-escaped
as a whole, because a single stray `<` in a remark or an email now costs the
~15-client page it lands on instead of one client's message.

Both admin reports looped over every client and called `SendMsgToTgbot` inside
the loop, so one tap produced one Telegram message per client. They now
accumulate a single report, which `SendMsgToTgbot` already pages at 2000
characters per message.

## Why

Two callbacks built their output by sending inside the loop:

- `reset_all_traffics_c` — `/clearall` → *Confirm*, plus the *Reset all
  traffics* button (`tgbot_router.go:1259`)
- `get_sorted_traffic_usage_report` — the *Sorted traffic usage report* button
  on the bot's main keyboard (`tgbot_router.go:1287`)

### What the operator sees today

An operator with a few hundred clients taps *Sorted traffic usage report* (the
button exists precisely for that size of panel). The bot then issues one
`sendMessage` per client, back to back, with 100 ms between them.

- Telegram throttles a chat far below that burst rate. The excess comes back
  as `429 Too Many Requests`, and `SendMsgToTgbot` classifies a rate-limit
  answer as neither `connection` nor `timeout` nor `closed`, so it stops
  retrying and logs `Error sending telegram message:` once per message
  (`tgbot_send.go:154`). The report arrives in fragments, silently missing the
  clients whose messages were refused, and the log fills with one warning line
  per refused message.
- The same tap on a 500-client panel is ~500 API calls, each one holding a
  worker slot in the 10-goroutine pool that also serves every other callback.

The single-message path avoids all of it: `SendMsgToTgbot` already splits a
long report with `pageMessage`, so the same content goes out as a handful of
paged messages, and the page ceiling is the documented 2000-character message
limit rather than the client count.

This is the same burst pattern as the merged `#6340` (per-refresh log spam);
only the loop was left.

### The two things batching changes

**A panel with no clients.** *Reset all traffics* → *Confirm* always ended with
the "process finished" reply and its `ReplyKeyboardRemove`. Guarding the whole
send with `report.Len() > 0` made an empty panel answer nothing: the prompt had
already been deleted, so the operator tapped *Confirm* and watched the message
vanish. `FinishProcess` is appended unconditionally again and that guard is gone
from the path.

**An unescaped `<`.** Both reports go out through `SendMsgToTgbot`, which sets
`ParseMode: "HTML"`, and the text carries operator input — the inbound remark
(`clientInfoMsg` → `describeAttachedInbounds`) and the email, both raw. Before,
a remark like `DE <fast>` broke that one client's message and the rest of the
report still arrived; with ~15 clients per page it now costs the whole page, and
the 400 matches none of the `connection`/`timeout`/`closed` substrings the retry
logic looks for, so the page is dropped with one warning.

Both reports are therefore escaped as a whole before they are sent. That is safe
because none of the i18n values on these two paths carry markup in any of the 13
locales (checked across `internal/web/translation/*.json`); the bot's HTML markup
lives in the command help and welcome strings, not in the report strings.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Xray config generation
- [x] Statistics / traffic counters — report delivery only, reset semantics untouched
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [x] Telegram bot

## How was this tested?

`internal/web/service/tgbot/tgbot_report_batching_test.go` seeds one inbound with
three clients — each with its traffic row, its `clients`/`client_inbounds` link
and its settings entry, so all three reach the report — and drives the admin
callbacks through `answerCallback` against a mocked Telegram API that records
what Telegram would have received:

- `TestTrafficUsageReportIsOneMessage` — exactly **one** `sendMessage` for the
  whole report. On the old code that tap sends three, so it fails before the fix.
- `TestResetAllTrafficsAnswersWithNoClients` — an empty panel. Asserts one
  message, non-empty text and `{"remove_keyboard":true}`. Fails on the version
  that guarded the whole send.
- `TestTrafficUsageReportEscapesHtml` — an inbound remarked `DE <fast>`. Asserts
  the text carries `&lt;fast&gt;` and no raw `<fast>`. Fails without the escape.

The message text is otherwise unchanged: the per-client blocks are the same
`clientInfoMsg` output, now separated by a blank line and joined into one
report, and the "no result" entries for emails without a traffic row are
appended to the same report.

Both loops share the mechanism, but the tests drive the read-only report path and
the empty-panel answer; `reset_all_traffics_c` performs real resets through the
traffic writer once clients exist, which the throwaway-DB test environment does
not stand up.

## Screenshots / recordings

N/A — bot messages, no UI.

## Breaking changes

None for the values the bot itself formats. Values carrying `<`, `&`, `>`, `"`
or `'` are now escaped, which is what Telegram's HTML parse requires — such a
report previously lost the page it landed on. The reply keyboard is still
removed, and `SendMsgToTgbot` applies it to the last page.

## Checklist

- [ ] I tested the change locally and confirmed the described behavior. — no C compiler on this machine, see above; CI runs the new test.


BlindMaster24:
Review addressed in `f3d661e5`. All three LOWs were right; two of them were regressions this PR introduced, so they are fixed here rather than left as follow-ups.

**1. An empty panel answered nothing.** The unconditional `FinishProcess` + `ReplyKeyboardRemove` is back — the whole send is no longer guarded by `report.Len() > 0` on that path, so a panel with no clients behaves exactly as it did before. Covered by `TestResetAllTrafficsAnswersWithNoClients`, which asserts one message, non-empty text and `{"remove_keyboard":true}`; it fails on the previous head of this branch.

**2. The widened blast radius is closed, not just documented.** Both reports are escaped as a whole before they are sent. I checked that this is safe instead of assuming it: none of the i18n values on these two paths carry markup in any of the 13 locales (`tgbot.messages.{email,upload,download,total,expire,expireIn,refreshedOn,enabled,online,lastOnline,active}`, `SuccessResetTraffic`, `FailedResetTraffic`, `FinishProcess`, `noResult`, `wentWrong`) — the bot's HTML markup lives in the command help and welcome strings, which these paths never reach. `TestTrafficUsageReportEscapesHtml` fails without the escape.

**3. The three-line comment is two lines.** The comment check is clean in every file this PR touches; the three-line block at `tgbot_router.go:1335` (`isClientSelfCallback`) is pre-existing and left alone.

The report text itself is otherwise unchanged: same `clientInfoMsg` blocks, joined and paged.

BlindMaster24:
Head moved to `a50fd209`: the empty-panel test now asserts the reply markup contains `"remove_keyboard":true` rather than matching the whole JSON object, so an extra marshalled field in telego cannot make it fail for the wrong reason. No production code changed from `f3d661e5`.

BlindMaster24:
Heads-up on the red checks: `build (386)` and `build (armv7)` failed **after** the Go binary was built and passed the static-binary check, at the step that unpacks the Xray release archive — `##[error]Process completed with exit code 8` right after `inflating: xray`. That is a download step, not the diff; the other 18 checks are green, including `go-test` and `race`. I cannot re-run the workflow from a fork (`Must have admin rights`), so a re-run of those two jobs should clear it.

BlindMaster24:
Thanks for merging main in — but I think that merge resolved onto a `main` that predates your own follow-up fix, so the two red checks are that follow-up, not this PR.

`e74be3f2` merged `main` at `aaa5e61c`, which is *before* `b98f947e` ("answer only the link callbacks that match nothing"). So the branch carried the exact broken combination of #6489 + #6493 that `b98f947e` repaired, and `go-test`/`race` fail on precisely the two tests your commit message named:

- `TestClientLinkCallbackServesOwnClient`: `answerCallbackQuery calls = 1, want 0` — a served link tap got the spurious error toast.
- `TestUnroutableCallbackIsAnswered` (`no_such_callback`, `get_backup_legacy`): `answerCallbackQuery calls = 0, want 1` — an unknown payload was left unanswered.

No data race in the `race` job, by the way — it reports the same two assertion failures.

Fixed by merging current `main` (`c3b08b6d`) in: head is now `a960ac36`. After the merge, `git diff upstream/main..HEAD` is exactly the two files this PR owns — `internal/web/service/tgbot/tgbot_router.go` and `tgbot_report_batching_test.go` — so the branch carries nothing else back, and both of your follow-ups (including the `staleButtonServer` count guard in `tgbot_stale_button_test.go`) are in the tree.

No production change here beyond the merge; this PR's own content is unchanged from `a50fd209`. CI is re-running on `a960ac36` now.

BlindMaster24:
@MHSanaei ?

PR #6492, opened 2026-09-12
fix(tgbot): render the add-client draft as HTML and escape its values

BlindMaster24:
## Summary

The wizard's other two text steps — the email and comment prompts, which echo the
same draft values and delete the card before they are sent — are escaped too, so
the flow cannot end on an empty screen.

`BuildClientDraftMessage` — the card the bot shows while an admin fills in a new
client — was written in Markdown but is sent with `ParseMode: "HTML"`, and its
admin-supplied values went into the message unescaped. The card now uses the
same markup as every other message and escapes the values it interpolates.

## Why

`addClient` renders the draft through `SendMsgToTgbot` / `editMessageTgBot`, and
both set `ParseMode: "HTML"` (`tgbot_send.go:151`, `tgbot_send.go:238`). The
draft's title and email were written for Markdown instead:

```go
b.WriteString("📝 *New client draft*\r\n")
fmt.Fprintf(&b, "📧 Email: `%s`\r\n", client_Email)
```

### What the operator sees

**On every card, always.** The title renders as `*New client draft*` — literal
asterisks — and the email as `` `a@b.c` `` with literal backticks, while the rest
of the bot (the client info card, the subscription links, every paged report)
renders its emphasis and its monospace values correctly through `<b>`/`<code>`.
The add-client flow is the one screen written for the other parser.

**On some cards, nothing at all.** The values interpolated into the draft come
straight from what the admin typed: the comment, the TG id, the email, and — on
the `Attached:` line — the remarks of the inbounds the client is being attached
to. Telegram's HTML parser refuses the whole message when it meets an unescaped
reserved character, and an ampersand is enough:

- Admin picks *Add client*, opens *Change comment*, and types `Ali & Co` (or
  `paid <10 GB>`, or attaches an inbound remarked `A&B`).
- The draft is re-rendered with that text raw, Telegram answers
  `400 Bad Request: can't parse entities`, and `SendMsgToTgbot` logs one
  warning and gives up (`tgbot_send.go:154`).
- The admin sees no new card and no error — the wizard looks frozen, and every
  later step that re-renders the same text fails the same way. The only way out
  is to change the comment to something without `&`, `<` or `>`, which nothing
  tells them.

The same mistake sat in the TG-id prompt of the same wizard
(`tgbot_router.go:1081`), which is plain text with Markdown backticks around
`-`, `none` and the current value; it is rendered through the same HTML parse
mode.

### The sibling prompts

Two steps of the same wizard echo a draft value and delete the card before the
prompt goes out: `add_client_ch_default_email` and
`add_client_ch_default_comment` (`tgbot_router.go:1057`, `:1067`). Both
interpolated that value raw into an HTML-parsed prompt, so the symptom was the
same as the card's: an admin who kept a comment of `Ali & Co`, or typed an email
with a `<`, lost the prompt after the card was gone, with no error shown. Both
now escape the value they echo, exactly like the TG-id prompt in the same
wizard.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [x] Telegram bot

## How was this tested?

`internal/web/service/tgbot/tgbot_draft_render_test.go` sets the draft's own
fields to values that expose both halves — a comment of
`<b>promo</b> & <10 GB>` — and asserts on the emitted string: the title is
`<b>New client draft</b>`, no `*` or backtick marker survives anywhere in the
message, the raw comment markup does **not** reach it, and the escaped form
does. Both assertions fail on the old code.

`TestAddClientPromptsEscapeDraftValues` covers the two sibling prompts the same
way: it drives both wizard callbacks through `answerCallback` against a mocked
API with the draft set to `long<name>@example.com` / `promo <b>tag</b>`, and
asserts the prompt that reached Telegram carries the escaped form and not the raw
one. That assertion fails on the old code.

The remaining draft fields (traffic, expiry, IP limit) are generated by the bot
itself and are left unescaped on purpose; only the values an admin or an inbound
remark can carry are escaped.

I could not build or run the test locally: this machine has no C compiler, and
the CGo SQLite driver makes every package importing `internal/database` — this
one included — fail to compile without one (`CONTRIBUTING.md` documents the
MinGW-w64 install). `gofmt` is clean; CI is the build here.

## Screenshots / recordings

N/A — bot messages, no UI. The change is textual: `*New client draft*` becomes
**New client draft**, `` `email` `` becomes `email` in monospace, and a comment
of `Ali & Co` now delivers the card instead of being refused.

## Breaking changes

None. Same cards, same fields, same buttons; only the markup they are written in
changes, and a comment that used to break the message now renders.

## Checklist

- [ ] I tested the change locally and confirmed the described behavior. — no C compiler on this machine, see above; CI runs the new test.


BlindMaster24:
Review addressed in `5db23a89`.

The MEDIUM (pre-existing) is the same defect this PR fixes, and fixing the card while the two prompts beside it keep dropping the screen would have left the PR's own claim incomplete — so the prompts are escaped here too, not deferred.

- `add_client_ch_default_email` (`tgbot_router.go:1057`) and `add_client_ch_default_comment` (`:1067`) echo the draft value and delete the card before the prompt is sent; both now escape that value, exactly like the TG-id prompt this PR already handled.
- New test `TestAddClientPromptsEscapeDraftValues` drives both callbacks with `long<name>@example.com` / `promo <b>tag</b>` and asserts the prompt that reached Telegram carries the escaped form and not the raw one. Fails on the previous head.

Left unescaped on purpose, same reasoning as the card: the values the bot formats itself (traffic, expiry, IP limit, timestamps).

BlindMaster24:
Head moved to `5668ab42`: the prompt test failed on the first push because the tgbot package's tests never initialize the localizer, so `I18nBot` returned the bare key and there was no rendered text to assert on. It now installs a go-i18n bundle carrying the same two prompt templates, so the assertion is on the message Telegram would actually receive. Nothing else changed.

PR #6493, opened 2026-09-12
fix(tgbot): answer the callbacks the bot cannot route

BlindMaster24:
## Summary

Three paths through `answerCallback` returned without calling
`AnswerCallbackQuery`: a payload that matches no case in the email-scoped
switch, one that matches nothing in the switch after it, and a button whose hash
had aged out of the 20-minute storage, which replied with a chat message only.
All three answer now, and the expired-hash path keeps its notice.

## Why

Telegram keeps the tapped button in a loading state until the bot calls
`answerCallbackQuery`; the Bot API documents answering as mandatory for every
callback query. A callback that is never answered leaves that indicator running
until the client gives up, and the user is told nothing about what happened.

### The reachable half: a button older than its payload

Payloads longer than 64 characters are not sent as button data — `encodeQuery`
replaces them with an MD5 hash kept in `hashStorage` (`tgbot.go:493`). That
storage is in-memory and TTL-limited: `Start` creates it with a 20-minute
expiration (`tgbot.go:245`) and a cron sweeps it every two minutes.

Long payloads are ordinary, because the payload carries the client email:
`client_individual_links <email>` crosses 64 characters for an email of about 40
(`tgbot_client.go:241`), `limit_traffic_in <email> ...` at about 43, and the
same for every other per-client button.

So an admin opens a client card, comes back to it later — or the panel is
restarted, which empties the storage outright — and taps *Refresh*:

- `decodeQuery` fails, the admin gets a "no query" chat message, and the button
  keeps spinning: the tap is never resolved.
- That message was the *only* feedback, so the reason for the dead tap is easy
  to miss among the bot's other messages.

It now answers with the same text, shown right where the tap happened, which
clears the indicator, and keeps the chat message: `sendCallbackAnswerTgBot` is a
single unretried call while `SendMsgToTgbot` retries connection errors, and the
only explanation for a dead button should not depend on one attempt succeeding.

### The unroutable payloads

Three places dropped a tap on data the bot could not route:

- the email-scoped switch (`tgbot_router.go:329`), reached by an admin's payload
  that carries arguments. An unknown action there falls out of the switch into a
  `return` that says nothing, and the two-token shape is exactly what an older
  panel version's buttons look like;
- the `default:` fall-through of the switch that follows it
  (`tgbot_router.go:1324`): a payload that matches no case at all, including the
  single-token button names an older version used;
- the admin decode path (`tgbot_router.go:316`): a payload longer than 64
  characters travels as a hash, and once that hash ages out of `hashStorage`
  (20 minutes, or a panel restart, which empties it) `decodeQuery` fails and the
  tap is left unresolved.

All three now answer with the bot's generic error text. The expired-hash path
also keeps the chat message it already sent, because `sendCallbackAnswerTgBot`
is one unretried call while `SendMsgToTgbot` retries connection errors — after a
restart that notice is the only explanation the admin gets.

The numeric-pad answers this PR originally carried are gone, as the review
thread says: `add_client_limit_traffic_in`, `add_client_reset_exp_in` and
`add_client_ip_limit_in` only fall out of their parse on a payload the bot itself
never generates, so the branch was unreachable and its test certified nothing.
The admin-only gate is left alone on purpose — refusing a non-admin silently is
the deliberate default-deny side of that check, and whether a client should be
told that a button exists but is not theirs is a product decision, not a routing
bug.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [x] Telegram bot

## How was this tested?

`internal/web/service/tgbot/tgbot_declined_callback_test.go` drives
`answerCallback` as an admin against a mocked Telegram API, with `isRunning` set
so a posted message would be counted:

- `TestUnroutableCallbackIsAnswered` — three payloads that match nothing, each
  answered with nothing at all before the change: `no_such_callback` and
  `get_backup_legacy` (single token, the shape an older panel's button names
  had), and `add_client_legacy_step 5` (the two-token shape, which reaches the
  email-scoped switch). Asserts exactly one answer and zero posted messages.
- `TestExpiredCallbackHashIsAnsweredAndReported` — a 32-hex payload that was
  never stored, with an empty `hashStorage`. Zero answers before the change; one
  answer and one notice after it.

The two-token case is what found the third path: it was the one case that stayed
red on the first push of this test, because that payload returns out of the
email-scoped switch before reaching the answer added for the other two.

## Screenshots / recordings

N/A — bot callbacks, no UI.

## Breaking changes

One feedback detail changes on the expired-hash path: it gains a callback answer
in addition to the chat notice it already sent, so the admin sees a toast as well
as the message. Nothing is removed, and no other callback changes.

## Checklist

- [ ] I tested the change locally and confirmed the described behavior. — no C compiler on this machine (the CGo SQLite driver needs one), so `go test` cannot run the packages importing `internal/database`, this one included; `gofmt` is clean and CI runs the new tests.


BlindMaster24:
Review addressed in `2c6c2458`. Both LOWs landed.

**1. The numeric-pad answers are gone.** You were right that `strconv.Atoi` on those payloads cannot fail for data the bot itself generates, so the branch was unreachable and the test that drove it certified nothing. Rather than keep a green test over dead code, I removed the three added answers and that test.

**2. The one-shot channel is no longer the only feedback.** The expired-hash path now answers *and* keeps the chat message it already sent, so the notice still benefits from the retry/backoff in `SendMsgToTgbot` while the answer clears the indicator. That is the path the description leads with, and it is the one where the durable half matters after a panel restart.

With the pads gone, the PR now covers the two declines that are reachable and silent:
- the admin-only gate — a client tapping a button carried out of an admin's chat was refused with no answer at all, so the button spun;
- the `default:` fall-through — a payload the bot no longer recognises matched nothing and was dropped silently.

`TestDeclinedCallbacksAreAnswered` covers both (exactly one answer, zero chat messages with `isRunning` set), and `TestExpiredCallbackHashIsAnsweredAndReported` asserts both halves of the expired-hash reply.

One note for the merge order: #6489 rewrites the same gate block, so whichever lands second may need a rebase. Otherwise the two are independent.

BlindMaster24:
Correction to my previous comment, head is now `5cef315c`.

The admin-only gate answer I announced there is out. Adding it made the existing `TestAnswerCallbackDeniesPrivilegedActionToNonAdmin` panic (that test drives the gate with no bot stub, so the new answer reached a nil `bot`), and stubbing the API inside it would have left the test checking less than it does today. Rather than weaken a test that is not part of this change, the gate keeps its silent refusal — default-deny is deliberate there anyway, and whether a client should be told that a button exists but is not theirs is a product decision.

So the PR is back to the two paths it can fix cleanly: the `default:` fall-through for a payload that matches no case, and the expired-hash path, which now answers *and* keeps its chat notice. The test file was renamed to match (`TestUnroutableCallbackIsAnswered`) and no existing test is touched by this PR.

BlindMaster24:
One more correction: the merge-order warning in my first comment is obsolete. With the gate hunk gone this PR no longer touches the block #6489 rewrites, so the two are independent and can land in either order.

BlindMaster24:
Head moved to `29a14e93` — the third path is now covered too.

While rewriting the test I hit a case that stayed red: an admin payload with arguments whose action matches no case (`add_client_legacy_step 5`) returns out of the email-scoped switch at `tgbot_router.go:329`'s tail `return` before ever reaching the answer added for the two single-token shapes. That is the two-token form an older panel version's buttons have, so it is reachable in exactly the same way as the others. It now has a `default:` clause on that switch that answers with the same generic error text, and the payload is a case in `TestUnroutableCallbackIsAnswered`.

Diff is 5 lines of production code in this push (the `default:` clause and its two-line comment). The earlier correction still stands: no gate, no `answerCallbackQuery` on the admin-only refusal, no existing test touched.

PR #6384, opened 2026-09-01
feat: Add opt-in Happ crypt5 generation to the client QR modal

NgaiYeanCoi:
## Summary

- Adds an optional Happ view to the QR modal for an individual client.
- Selecting Happ requests a fresh, validated `happ://crypt5/` link from the authenticated backend and displays it through the existing QR, copy, and image actions.
- It generates a fresh, validated `happ://crypt5/` link through the authenticated backend without persisting or caching the result, while keeping the Standard subscription view available as the default and fallback.

## Why

Administrators currently need to copy a client's complete subscription URL into an external Happ service manually. This change provides a client-scoped workflow in which the browser sends only the client's internal record ID, while the backend reconstructs the current subscription URL and applies bounded provider requests, strict response validation, and redacted diagnostics.

Privacy note: selecting Happ and generating the link sends the client's complete current subscription URL to the third-party service `crypto.happ.su`.

Fixes #6381 
Refs #3902

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Refactoring (no behavior change)
- [X] Documentation
- [ ] Tests only
- [X] Build / CI / tooling
- [ ] Other

## Areas affected

- [X] Frontend (UI / panel pages)
- [X] Backend (API endpoints, login, settings)
- [ ] Xray config generation
- [X] Subscription (share links / Clash / JSON)
- [ ] Statistics / traffic counters
- [ ] Database / migrations
- [ ] Install / upgrade script
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot

### Frontend Tests
- npm run lint
- npm run typecheck
- npm run build
- npm run build-storybook
- npm test -- --project components src/test/client-qr-modal.test.tsx
- npm test -- --project unit src/test/i18n-dead-keys.test.ts

### Backend Tests
- go build ./...
- go test ./internal/web/service -run 'Test(Happ|ParseHapp|SanitizeHapp)' -count=1
- go test ./internal/web/controller -run TestGenerateHappLink -count=1
- go test ./internal/web -run TestRouteRegistryContract -count=1
- go test ./tools/openapigen -run TestWalkPackagesExportsAllowedStructsWithoutDIInterfaces -count=1
- go test ./...
- make verify 
but shown 
```bash
[openapi] paths: 184, operations: 191, tags: 13
git diff --exit-code -- frontend/src/generated frontend/public/openapi.json
golangci-lint run
[...]

Issue #6381, opened 2026-09-01
[Feature]: Add opt-in Happ crypt5 generation to the client QR modal


NgaiYeanCoi:
### Is your feature request related to a problem?

Administrators who need a Happ `happ://crypt5/` link do not currently have a safe and explicit workflow for generating it from a client's current subscription in the panel.

Doing this manually requires copying a complete subscription URL to an external service. That URL is sensitive because possession of it may grant access to the subscription. 

### Describe the solution you'd like

Add an optional **Happ** view to the QR modal for an individual client on the **Clients** page.

Expected behavior:

- Open the modal on **Standard** by default. Opening the modal must not contact any external service.
- Let the administrator explicitly select **Happ**.
- Display the validated `happ://crypt5/...` link using the existing QR, copy, and image-download flow.
- If generation fails, show a generic error and Retry action while leaving the Standard link fully usable.

Backend and privacy constraints:

- The browser sends only the client's internal record ID, not a caller-supplied subscription URL.
- The authenticated backend loads the current client and constructs the current canonical subscription URL.
- Use the existing authorization and CSRF protections.
- Send the URL only to the fixed HTTPS Happ endpoint documented at:
  https://www.happ.su/main/zh/dev-docs/crypto-link
- Do not follow redirects or automatically retry the provider request.
- Apply a bounded timeout and response-size limit, and accept only one valid, non-empty `happ://crypt5/` result.
- Mark responses as `no-store`.
- Keep generation stateless: no database fields, cache, history, background jobs, or persisted generated links.
- Logs and user-facing errors must not contain subscription URLs, subscription tokens, provider bodies, or generated Happ links.

### Alternatives you've considered

- Continue using only the Standard subscription link, which does not provide the requested `happ://crypt5/` representation.
- Ask administrators to paste sensitive subscription URLs into the Happ website manually. This makes the disclosure easy to overlook and may result in using a stale or incorrect client URL.
- Call the provider directly from frontend code. This would expose the integration to browser and CORS behavior and make consistent validation and redacted diagnostics harder.
- Store generated Happ links for reuse. This introduces unnecessary sensitive state and may return stale links after client or panel settings change.

### Concrete use case

An administrator opens the QR modal for one client. The existing Standard QR code is displayed and no external request is made.

The administrator selects Happ, The backend receives only the client record ID, reconstructs the client's current subscription URL, performs one bounded provider request, validates the response, and returns a fresh `happ://crypt5/...` link.

[...]

PR #6494, opened 2026-09-13
feat(happ): generate Crypt5 subscription links locally

NgaiYeanCoi:
## Summary

Add on-demand Happ Crypt5 link generation to the client QR code dialog. Generation runs entirely inside the 3x-ui backend process, without sending the client's full subscription URL to `crypto.happ.su` or another generation service. The feature retains the administrator-controlled switch, disabled by default, the Standard view as the default, regeneration, and full-link copying.

The implementation uses the salted Crypt5 format with a pinned public key and marker validated in client tests, and supports subscription URLs beyond the capacity of a single RSA block. It addresses disclosure to third parties during generation; it does not promise that links cannot be decoded or shared.

## Why

Refs [#6384](https://github.com/MHSanaei/3x-ui/pull/6384).
Close[#6381](https://github.com/MHSanaei/3x-ui/issues/6381).

The original design sent the full URL, including the subscription access credential, to Happ's remote generation service.  @MHSanaei explained that even with a default-off switch and disclosure, the panel should not become a channel for handing client subscription URLs to a third party. @MHSanaei indicated willingness to revisit local generation if Happ publishes the public key and format. [Closing comment](https://github.com/MHSanaei/3x-ui/pull/6384#issuecomment-5633854984)

This revision removes remote requests from the generation process. Happ's [official encrypted-link documentation](https://www.happ.su/main/zh/dev-docs/crypto-link) describes its web/API generation service and a deprecated RSA/Crypt4 method, but does not provide a complete specification for generating Crypt5 links locally. Links generated with the RSA public key published on that page could not be imported in my Android and Windows tests. I also evaluated a Crypt4 fallback, including a [third-party key kastov/cryptohapp's src/constants/crypt4.constant.ts](https://github.com/kastov/cryptohapp/blob/04e82904c3f71cf8c3b2fe02e65f854e3550f81a/src/constants/crypt4.constant.ts) that passed mobile and desktop import tests.

I therefore developed the local Crypt5 implementation by analyzing the reverse-engineered format and cross-referencing the public implementations in [hpwnr's src/keys.rs](https://github.com/Omegaplexx/hpwnr/blob/main/src/keys.rs) and [happ-decryptor's src/crypt5.js](https://github.com/LeeeeT/happ-decryptor/blob/main/src/crypt5.js). The matching public key was extracted from the published RSA private-key entry, and the format was implemented in Go with pinned reference versions and a fixed public-key fingerprint.

Local Crypt5 links using this format have passed offline interoperability checks and real-client import and update tests on Android 4.3.0 and Windows 4.1.2, including correct Title retrieval and a 502-byte source URL. The real-client checks used an isolated subscription server; their scope and the remaining validation gaps are detailed below.

## User interaction and request flow
he browser submits only the client's internal record ID. After the existing authentication and authorization checks, the backend reads the current subscription settings and the client's SubID, constructs the actual source URL, and passes it to the local generator.

After generation, the backend checks again for request cancellation, changes to the source URL, or the administrator disabling the feature. Stale results are discarded. Generated links are neither written to the database nor cached by the backend, and responses carry `Cache-Control: no-store`. The frontend also discards results from invalidated requests when the dialog closes or the selected client changes.

```mermaid
sequenceDiagram
    participant B as Panel browser
    participant P as 3x-ui backend
    participant D as Panel database
    participant G as In-process Crypt5 generator
    participant H as Happ client
    participant S as Subscription server

    B->>P: POST /panel/api/clients/happLink/:id
    P->>P: Check authentication, authorization, CSRF, and feature switch
    P->>D: Read current subscription settings and client record
    D-->>P: Subscription base URL and SubID
    P->>G: Original subscription URL
    Note over P,G: Generation does not contact a third-party generation service
    G-->>P: Newly generated happ://crypt5/ link
    P->>P: Recheck cancellation, source URL, and feature switch
    P-->>B: Return link with Cache-Control: no-store
    B-->>H: User copies or scans the link to import it
    H->>H: Recover the original subscription URL
    H->>S: Fetch or update the subscription
    S-->>H: Return subscription content and name information
```
"Local generation" means that the subscription URL is not sent to an external service while generating the link. Happ still needs to access the original subscription server when importing or updating it. Crypt5 encapsulates the subscription URL. This feature does not add or rewrite the subscription name, which remains determined by the existing subscription response and Happ's parsing logic. The existing subscription response and node configuration generation logic remain unchanged.

### 1. Two encryption paths
ChaCha20-Poly1305 encrypts the payload. A fresh symmetric key is generated for each link, processed according to the format, and encrypted with the pinned RSA public key. RSA does not directly encrypt the entire long URL.

```mermaid
flowchart TD
    URL["Original URL as UTF-8 bytes P"] --> URLPACK["Standard Base64, then swap adjacent bytes"]
    RNG["crypto/rand"] --> KEY["Random key K: 32 bytes"]
    RNG --> NONCE["Nonce N: 12 alphanumeric characters"]
    RNG --> SALT["Salt S: 8 alphanumeric characters"]
    RNG --> TAG["Format field T: 2 letters"]

    URLPACK --> AEAD["ChaCha20-Poly1305 with empty AAD"]
    KEY --> AEAD
    NONCE --> AEAD
    AEAD --> C["C: Base64 of payload ciphertext and 16-byte authentication tag"]

    KEY --> XOR["XOR K with repeated S"]
    SALT --> XOR
    XOR --> KEYPACK["Standard Base64, then swap adjacent bytes: exactly 44 bytes"]
    KEYPACK --> RSA["RSA-4096 with PKCS#1 v1.5"]
    PUB["Pinned matching RSA public key"] --> RSA
    RSA --> R["R: Base64 of the 512-byte RSA ciphertext"]

    NONCE --> BODY["Concatenate N, T, S, length of C, V, C, and R"]
    TAG --> BODY
    SALT --> BODY
    C --> BODY
    R --> BODY
    BODY --> MARK["Prepend vdfz and append foff"]
    MARK --> SWAP["Swap the two halves of each complete four-byte block"]
    SWAP --> LINK["Prepend happ://crypt5/"]
```
### 2. Exact encoding definitions

The following describes the Go implementation in this PR. [Generator source](https://github.com/NgaiYeanCoi/3x-ui/blob/2646579330ca0bec392cd46236659f697991d0c6/internal/web/service/happ_crypto.go)

| Symbol | Meaning |
| --- | --- |
| `P` | The original subscription URL as UTF-8 bytes. Percent encoding, parameter order, repeated parameters, plus signs, and the fragment are preserved. |
| `B64` | Standard Base64 using the alphabet containing `+` and `/`, with `=` padding retained. |
| `S2` | Swap each pair of adjacent bytes: `ABCD → BADC`. A final unpaired byte is unchanged. |
| `S4` | Swap the two halves of each complete four-byte block: `ABCD → CDAB`. Any trailing group of fewer than four bytes is unchanged. |
| `K` | The 32-byte random symmetric key generated for this link. |
| `N` | Twelve ASCII letters or digits, with their raw bytes used directly as the nonce. |
| `T` | Two ASCII letters forming a format field in the salted layout. |
| `S` | Eight ASCII letters or digits used as the salt when wrapping the key. |
| `M` | The fixed key marker `vdfzfoff`. |

```text
Q = S2(B64(P))
C = B64(ChaCha20-Poly1305.Seal(K, N, Q, AAD=empty))

W[i] = K[i] XOR S[i mod 8], i = 0..31
R = B64(RSA-PKCS1-v1_5-Encrypt(PublicKey, S2(B64(W))))

body  = N || T || S || decimal(len(C)) || "V" || C || R
frame = "vdfz" || body || "foff"
link  = "happ://crypt5/" || S4(frame)
```

`||` denotes direct concatenation. The length field records the length of the Base64 string `C`, not the URL byte length or the total link length. The RSA input is always 44 bytes. The RSA ciphertext is always 512 bytes, represented by 684 Base64 characters.

ChaCha20-Poly1305 uses the IETF variant with a 12-byte nonce. Its output contains the payload ciphertext followed by a 16-byte authentication tag. This authentication tag is distinct from the two-letter format field `T`.

The reference decoder reverses these steps to recover the URL: remove the scheme prefix, apply `S4` again, and read the marker and fields; decrypt `R` with the RSA private key selected by the marker, then apply `S2`, Base64 decoding, and salt XOR to recover `K`; decrypt and authenticate `C` using `K` and `N`, then apply `S2` and Base64 decoding to recover the original URL. Both byte-swap operations and XOR are reversed by repeating the same operation. This describes the public reference implementation's decoding path, not knowledge of Happ's internal client code.

### 3. Public key, marker, and separator

The implementation pins the following values:

- Marker: `vdfzfoff`
- Separator: the letter `V`
- Public key: the RSA-4096 public key matching this marker, with public exponent `65537`
- SPKI DER SHA-256: `22319c7b13647897bf5fd4f827ba92bf3946d738007a0054ccd931c31f221768`

The client needs the marker to identify the matching decryption key. Generating an arbitrary new RSA key pair does not automatically make it compatible with Happ. The PEM, RSA parameters, and fingerprint are checked when loading the public key to prevent accidental mismatches between the marker and key. These checks do not constitute certificate authentication or official endorsement.

hpwnr's generator explicitly writes a colon separator with `body.push(b':')` in [src/lib.rs, lines 572–575](https://github.com/Omegaplexx/hpwnr/blob/3745cb96e2551e003cb217ab7705b4d67f8ac006/src/lib.rs#L572-L575). Its decoder skips one separator byte without requiring it to be a colon, as shown in [lines 472–478](https://github.com/Omegaplexx/hpwnr/blob/3745cb96e2551e003cb217ab7705b4d67f8ac006/src/lib.rs#L472-L478). Separately, in three non-sensitive synthetic samples obtained during this investigation by submitting form POST requests to the official website endpoint (`https://crypto.happ.su/`) from a local Python script, the observed separators were `V`, `R`, and `o`. The isolated tests therefore included both the letter `V` and a colon control. Both forms passed on Android and Windows, and the implementation pins the tested `V`. These three observations do not establish an official commitment to a fixed separator or marker.

The implementation does not rotate through unvalidated markers, download keys dynamically, or fall back to remote generation on failure. If future clients change the format or keys, the mapping will need to be updated and the format checks, interoperability tests, and real-client tests repeated.

### 4. How the official-service samples were collected and checked

On **13 September 2026, between 02:23:46 and 02:23:52 UTC** (10:23:46–10:23:52 UTC+8), I used a Python `urllib` script in a local Windows terminal to submit three form requests to [Happ's official generation website](https://crypto.happ.su/). This was a scripted HTTP request to the website's form endpoint, not a browser UI test.

| Request detail | Value |
| --- | --- |
| Method and endpoint | `POST https://crypto.happ.su/` |
| Content-Type | `application/x-www-form-urlencoded` |
| Form field | `url=https://example.com/sub` |
| User-Agent | `Mozilla/5.0` |
| Requests | 3, all using the same synthetic source URL |

An equivalent request for collecting one sample is:

```bash
curl --request POST 'https://crypto.happ.su/' \
  --header 'Content-Type: application/x-www-form-urlencoded' \
  --user-agent 'Mozilla/5.0' \
  --data-urlencode 'url=https://example.com/sub'
```

All three responses were recorded as HTTP 200. The script decoded the returned HTML, unescaped HTML entities, and extracted `happ://crypt5/` links. After deduplicating links within each response, each response yielded one complete **795-character** link; the three collected links were different. The complete extracted links and terminal request/extraction records were retained. Complete response HTML and response headers were not retained. These are outputs collected from the official service for this investigation, not test vectors published or certified by Happ. A later request may produce a different link, marker, or separator.

I then checked the three links offline using the original direct decoder in [happ-decryptor's src/crypt5.js at commit `5a05cfe5`](https://github.com/LeeeeT/happ-decryptor/blob/5a05cfe59f45db8684be26077ce28979a3ee3208/src/crypt5.js), with native fallback disabled. The recorded environment was Node `v24.14.1`, `node-forge 1.4.0`, and `@noble/ciphers 1.3.0`. All three decoded exactly to `https://example.com/sub`. Their framing used the salted layout and marker `vdfzfoff`:

| Sample | Link characters | Header after the nonce, ending with the separator | Separator | Exact source URL recovered |
| --- | --- | --- | --- | --- |
| 1 | 795 | `CgncB9U2Rg64V` | `V` | Yes |
| 2 | 795 | `kcjqb4fnSG64R` | `R` | Yes |
| 3 | 795 | `rDWM9rhLq564o` | `o` | Yes |

In these headers, the first two characters are `T`, the next eight are the salt, `64` is the Base64 payload segment length, and the final character is the separator. These observations establish what these three samples contain; they do not establish the complete set of separators accepted by Happ.

The evidence sets serve different purposes:

- **These three official-service samples:** observations of the wire format and offline decoder checks, using only the synthetic `example.com` URL.
- **The separate 21/21 and 42/42 checks:** interoperability of the local Go generator with the reference decoder, plus rejection of modified payload ciphertext or authentication tags.
- **The separate A/B/D device tests:** locally generated links pointing to an isolated test subscription server, used to check import, update, and Title retrieval on the reported Happ versions. The three `example.com` samples above were not used as evidence of working subscription updates.

Only the research step described here contacted the official generation service, using a synthetic URL without subscription credentials. Link generation in the application runs locally in the 3x-ui backend and does not make these requests.

<details>
<summary>Complete collected samples and an offline separator inspection script</summary>

The following is the retained sample data. Save it as `official-samples.json` to inspect the exact links used in this investigation. The timestamp records when the collected sample file was written; individual response timestamps were not retained.

```json
{
  "retrieved_utc": "2026-09-13T02:23:52.509625+00:00",
  "samples": [
    {
      "sample": 1,
      "source": "https://example.com/sub",
      "link": "happ://crypt5/fzvdA0w0kL2gXAXcncCgU2B964RgJzV1gsJtUWsTICDU80jRLNg9s6HDUG4OFJBLn/taPXj18iUchr7opgQ6LG1+dgtu+tGj2MBgi0Thihpm8LKqfcJVq6a+6ltFamrCau2YsJBpeWKT27TJHpyLj9i91q2H/WOcLCULKsgkbW/6kJvD2KS5FEmEujAp0FpEbPREPjRIiF8z/O1lQUKC092330HmaZoDAIzJR2XQ5t+MavleBrobLiV6YQ+2ncac5cK0qCx0ch9vScBrUi+DLv0xUGLVXfcBqjVgUaFJylVsqtx8gDdBh5MBB8PNotX3T0C++aHhEMYZUSyu9kfJwS0CKCsMfj5wLN13hPUd3dijUrTbDzBnPCAMLMeNhKIvDDQgVon5cQzoLaXuFrj424ixzX8yy8+ebi9p1v5Ir9YGyCfkXIynexAW35kGX1jmyDXTLIEwn/V5uwOrFs3h5fbco1WvM0WtQF7Yeo4t2axFkrk3RlLZgt1xAzEpRsvpLKvVhCOJnwBCaHqJx09lE3uMEXFWehqqCqnwktaIpvCWExuOY5BJvjJYsCD8RlSHZ2NDLMDVYkI8QdJ+lM/kuoJw+WFBk+fxqHTbhgxeyeUQ+URpqmov4XQJnp70jAFzjPYeZWUDmP7rJ0uwZA8w/9hNV1p1jC4qSnY+uXOFAy/TOLs3MSxQ6MWX8AOe/V/t1tz9wc+n+NYjzAYd2AU+vRaXd/dqhbVzNVpGmOcL8EhY1Fc+lDSgdVgAupWeP1iakXwtR8Emof=ff",
      "endpoint": "https://crypto.happ.su/"
    },
    {
      "sample": 2,
      "source": "https://example.com/sub",
      "link": "happ://crypt5/fzvdtxjWzvGBY6Ijjqkcfnb464SG+iRnnUwonDgGXh+y+zD4SOrBGn9fE8TWrha6Qa2aUrtXLhHLzSVSyCmubAJ8EWD3cZ7qNs2o35sRbfjxQ7A4TbN5HwkSf090QZhzVRN/pP0C8rnUGfehd7g+JR1IVEXFHENY1i91kaIbOTqHiaz71qCURiSpWJiosyisEirJv9elxkxR35yX/rDw42AqvcPswuBedLoY+Xsi8C47nquACyCNjP5s2V0tfRYvR+35KYcVtWlz/om0RhTTNt2KJdq9QagqdO7hrRGsnOLc6prxDBd3ZWx/sjoyBCA5jhyRhOrTcRkiKhkwkHsgDtaA+vextYWcJHg7y7cvAgKMSbAWCygvwMqFnTu9AOoQesv8BVCN+ZLRppKSONASbaIfnxgQKpP8omH7Vl8WC0MbQE9ll9ESzQghWzfXwFRYdU+gd6QC44gcItdE8Pp3QYJQbS4NKHtfEAe/FZjyZuG3Ns+wxirkq+2jADQwtUvr8viDmhVszD2tZzOhFxsBWCV5CaQbtl7q6RjmsAlo3C7EL1DM/kQptdAkuj+MEBJ3GntNyhGDuG4+3wYMQQvewH5c1Ts0i3ySqz4rBib7MSKHrYd3MTbfwIWTcR9AvQTA9xWm/fQ8seOUNxriLzoSRkFVco3IkXfsNPgltMSRrpJt30JDTwO3O9QsyNkL4hO/+z8EYg1wt/fws6hQoukdtfKADD//RGWyIQOK3wjhkXjQPJD1tWjoP286y8lP4jFJ5wpGvA2yof=ff",
      "endpoint": "https://crypto.happ.su/"
    },
    {
      "sample": 3,
      "source": "https://example.com/sub",
      "link": "happ://crypt5/fzvdbWBGid55mlRPWMrDhL9r64q5xGod8Ose+LVA9dzhmUebYDtfCHvBApvGI1IOKe1ItJ7hR0MOkUBEJULPmOBArKrdA7q1/M8lXepOcRpSBE9y2JOBMf1t21cT80fjLCuKvb3CROcByqhHRNm39GIYY5SeGe9OIBer16Dkhh6dYShvZtqU9JXwyNMChSz+1ywQE/CwhyifV5hlWJwJDf/3lTThmalRPrUiej2Mfw4p+YjD0vwrpQH0RoeFi547pSuFAEmlPLEHWABf2KH9/XE3EmaBqGZRRsFkkK4/gJUjw5LdJGttGQ9Spm/1qiV9XhL/cHbN/aa559GlUgFcsA8wghHBqr0s5X7CnQxHsLyGyhX1yRamCGjf1DzUGeIY5hg7g8SuCXXLaWGINLOf3gAIJ7XDWlPC9CH2m+vuqHy2pIphhIphaqsLpLnb/sAZniVMVU9t2/mbgOrVOAHwxJ3m1EP2edLgd/TYTkW6Da2wW7T4I5CRkxfkCzy9326tJmJokf1kfcRDDT1QJnjX0+6iEuhBk4+rJs3I0/fx60xdUnBokG6xaO5hCJGHWB+iaefglWrMo7Q/uNxQPe3KrGvpOlN9TsjdGRqvq3lOuO6ajHpQshnb+zkdGh2jeWncNONH2i/hlPaM5250HmpSftcT5PKLWkTHOD6dSm6Nbmn3S7vzlFoE1LKN5EK76OgIKYpq1gH3DjYu0zKvaF2tSoKQuc/t/hTmZlQRfJE0VDZPqUaCkbWYwCs3Psc68OXy/qiDUA0Iof=ff",
      "endpoint": "https://crypto.happ.su/"
    }
  ]
}
```

Save the following as `inspect-official-samples.py` and run `python inspect-official-samples.py official-samples.json`. It uses only the Python standard library and performs no network requests. It reverses the outer byte swap and reads the fields of these recorded salted samples. It does not perform RSA/ChaCha decryption or establish client compatibility; exact URL recovery was checked separately with the pinned JavaScript decoder above.

```python
import base64
import json
from pathlib import Path
import sys

# Offline format inspection only; this neither contacts Happ nor decrypts the URL.
samples_path = Path(sys.argv[1]) if len(sys.argv) > 1 else Path("official-samples.json")
samples = json.loads(samples_path.read_text(encoding="utf-8"))
prefix = "happ://crypt5/"
for sample in samples["samples"]:
    link = sample["link"]
    assert link.startswith(prefix)
    frame = bytearray(link[len(prefix):].encode("ascii"))
    for offset in range(0, len(frame) - 3, 4):
        frame[offset:offset + 4] = frame[offset + 2:offset + 4] + frame[offset:offset + 2]
    marker = (frame[:4] + frame[-4:]).decode("ascii")
    body = frame[4:-4]

    # These recorded samples use the salted layout: nonce(12), T(2), salt(8).
    cursor = 22
    while cursor < len(body) and 48 <= body[cursor] <= 57:
        cursor += 1
    segment_length = int(body[22:cursor].decode("ascii"))
    separator = chr(body[cursor])
    ciphertext = body[cursor + 1:cursor + 1 + segment_length]
    rsa_ciphertext = body[cursor + 1 + segment_length:]
    base64.b64decode(ciphertext, validate=True)
    rsa_bytes = base64.b64decode(rsa_ciphertext, validate=True)
    print(json.dumps({
        "sample": sample["sample"],
        "link_bytes": len(link.encode("ascii")),
        "marker": marker,
        "header_after_nonce": body[12:cursor + 1].decode("ascii"),
        "separator": separator,
        "segment_base64_chars": segment_length,
        "rsa_cipher_bytes": len(rsa_bytes),
    }))
```

The expected separators are `V`, `R`, and `o` in sample order; all three report marker `vdfzfoff`, 795 link bytes, 64 Base64 payload characters, and 512 RSA ciphertext bytes.

</details>

## How the two reference projects were used

| Project and pinned version | Role in this work | Included in the application runtime? |
| --- | --- | --- |
| [Omegaplexx/hpwnr](https://github.com/Omegaplexx/hpwnr/tree/3745cb96e2551e003cb217ab7705b4d67f8ac006) | Used to cross-check the salted Crypt5 encoding flow, RSA/ChaCha parameters, and byte swaps. The public component was extracted from its published `vdfzfoff` PKCS#8 RSA private-key entry, and its SPKI fingerprint was pinned. The main references were [src/lib.rs](https://github.com/Omegaplexx/hpwnr/blob/3745cb96e2551e003cb217ab7705b4d67f8ac006/src/lib.rs) and [src/keys.rs](https://github.com/Omegaplexx/hpwnr/blob/3745cb96e2551e003cb217ab7705b4d67f8ac006/src/keys.rs). | No. The application uses a Go implementation and retains only the public key needed for generation. It does not include that private key. |
| [LeeeeT/happ-decryptor](https://github.com/LeeeeT/happ-decryptor/tree/5a05cfe59f45db8684be26077ce28979a3ee3208) | Its original [direct decoder in src/crypt5.js](https://github.com/LeeeeT/happ-decryptor/blob/5a05cfe59f45db8684be26077ce28979a3ee3208/src/crypt5.js) was used offline to check whether Go-generated output recovers the exact source URL and to check rejection of payload authentication failures. | No. The decoder and private-key table exist only in isolated research tools, outside the application code and application tests. |

The two projects' key tables were identical in this comparison, so they cannot be treated as two independent official sources. Here, "independent interoperability validation" means checking this Go generator with a separate decoder implementation. It does not imply independent key provenance or official confirmation.

The Go application uses the standard library's `crypto/rand` and `crypto/rsa`, plus the project's existing `golang.org/x/crypto/chacha20poly1305` dependency. It adds no third-party generation service, Rust CLI, native emulator, or new application cryptography dependency. Random ASCII fields are sampled uniformly using a cryptographically secure random source.

## Security and behavior boundaries

- The generation service does not fetch subscription content or send the full subscription URL to a third party. Authentication, authorization, CSRF protection, and the default-off switch remain in effect.
- Source URLs must be valid HTTP/HTTPS URLs. Userinfo, control characters, and invalid UTF-8 are rejected. Failure logs and API error responses for this feature do not include the original URL, SubID, or ciphertext details.
- Base64, XOR with a public salt, and byte swaps are compatibility format conventions. They do not provide additional confidentiality guarantees.
- AEAD authenticates the payload, but AAD is empty. This does not guarantee authentication of every outer format field, and AEAD is not a signature identifying the panel.
- Public decryption projects include the corresponding private keys, so someone holding a link may still recover and share the subscription URL. This feature does not replace subscription credential rotation, revocation, expiration, or server-side access control.
- If backend generation fails or the returned value does not match the Crypt5 format accepted by the panel, the panel displays a safe error. Actual Happ import errors are reported by the client. The Standard subscription option remains available.

## Long URLs and QR code handling

Encrypting a raw URL directly with a single RSA-4096 block and PKCS#1 v1.5 padding allows `512 - 11 = 501` bytes of plaintext. This implementation uses RSA for a fixed 44-byte key package and symmetric encryption for the payload, removing that single-block constraint.

The panel imposes a resource limit of **8192 UTF-8 bytes** on the source URL. Exceeding it returns `success: false`, `msg: happ_source_too_long`, and `obj: null`. This is the panel's processing boundary, not a guarantee of Happ client support or QR code capacity.

QR codes have a separate limit of 2953 link bytes under QrPanel's current error correction level L configuration. When the link exceeds that limit, the UI shows a notice and preserves full-link copying rather than providing a truncated link. For example, offline validation produced a 15322-byte Crypt5 link from an 8192-byte source URL; that link cannot be displayed as a QR code.

## Type of change

- [ ] Bug fix
- [X] New feature
- [X] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [X] Frontend (UI / panel pages)
- [X] Backend (API endpoints, login, settings)
- [ ] Xray config generation
- [X] Subscription (share links / Clash / JSON)
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot

## How was this tested?

```bash
go test -count=1 ./internal/web/service ./internal/web/controller -run Happ
go test -race -count=1 ./internal/web/service ./internal/web/controller -run Happ
go vet ./internal/web/service ./internal/web/controller

cd frontend
npm test -- --project components src/test/client-qr-modal.test.tsx src/test/client-qr-modal-qr-capacity.test.tsx
npm run typecheck
npm run lint
cd ..
make build build-storybook
make verify
```

### Real Happ clients

The following A/B/D results were confirmed by me using locally generated links and an isolated subscription server. These were separate from the three official-service samples for https://example.com/sub documented above. Android and Windows cases A/B/D each checked import, update, and the returned Title. On iOS, only successful import of A and the HTTP update block were recorded. These results do not mean that the integrated panel has undergone the same acceptance checks on every device, and offline decryption success is not being substituted for real-client validation.

| Client | A: salted + V | B: salted + colon | D: 502-byte source URL + V |
| --- | --- | --- | --- |
| Android 4.3.0 | Import and update succeeded; correct Title | Import and update succeeded; correct Title | Import and update succeeded; correct Title |
| Windows 4.1.2 | Import and update succeeded; correct Title | Import and update succeeded; correct Title | Import and update succeeded; correct Title |
| iPadOS/iOS 5.6.0 | Import succeeded; update reported that HTTP was prohibited | Testing not continued | Testing not continued |

The test server used HTTP.  The IOS only import has been confirmed , but HTTPS subscription update validation remains incomplete.

## Breaking changes
None

## Checklist

- [X] I tested the change locally and confirmed the described behavior.
- [X] I added or updated tests for the new behavior (when applicable).
- [X] `go build ./...` and the test suite pass locally.
- [X] For frontend changes: `npm run lint`, `npm run typecheck`, and `npm run build` pass.
- [X] I updated the Wiki / README / API docs if user-facing behavior changed.
- [X] My commits follow the project's existing message style.
- [X] I have no unrelated changes mixed into this PR.

## references
https://github.com/kastov/cryptohapp
https://github.com/LeeeeT/happ-decryptor
https://github.com/Omegaplexx/hpwnr
https://leeeet.dev/happ-decryptor/
https://docs.happ.info/main/dev-docs/crypto-link

my brains already overloaded...
Thx them  and maintainer again


spacexm406-web:
Not successful

NgaiYeanCoi:
> Not successful

?

PR #6486, opened 2026-09-12
feat(discord): add Discord notification bot service

poise52:
## Summary

Integrates native Discord notification support via a lightweight Discord REST API v10 client and the existing `internal/eventbus/` infrastructure, accompanied by a dedicated Discord settings tab and event notification triggers in the UI.

## Why

Many server administrators prefer Discord over Telegram for monitoring server health and receiving incident alerts. This adds first-class support for Discord embeds covering system health (`cpu.high`, `memory.high`), node status (`node.down`, `node.up`), outbound connectivity (`outbound.down`, `outbound.up`), core crashes (`xray.crash`), and security events (`login.attempt`), complete with rate limiting and secret protection.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring (no behavior change)
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [x] Frontend (UI / panel pages)
- [x] Backend (API endpoints, login, settings)
- [ ] Xray config generation
- [ ] Subscription (share links / Clash / JSON)
- [ ] Statistics / traffic counters
- [ ] Database / migrations
- [ ] Install / upgrade script
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot

## How was this tested?

1. **Automated Unit & Contract Tests:**
   - `go test -v ./internal/web/service/discord/...` — all 18 tests passed (covering HTTP status codes, rate limiter semantics, rune-safe embed truncation, value/pointer event data).
   - `go test -v -run TestRouteRegistryContract ./internal/web` — verified bidirectional route parity for `POST /panel/api/setting/testDiscord`.
   - `npx vitest run src/test/factory-defaults-contract.test.ts` — verified setting default synchronisation across Go and TypeScript.
   - `npx vitest run src/test/i18n-dead-keys.test.ts` — verified 0 dead keys and exact key parity across all 13 locales.
   - `npm run typecheck && npm run lint && npm run format:check` — clean.
   - `go build ./...` and `npm run build` pass without warnings.

2. **Manual & Live Discord API Testing:**
   - Configured a real Discord Bot Token and Channel ID in `/settings#discord`.
   - Clicked "Send Test Notification" in the UI and verified instant receipt of the test embed in the Discord channel.
[...]

PR #6496, opened 2026-09-13
fix(discord): keep the !inbounds reply within Discord's embed caps

BlindMaster24:
## Summary

Pages the Discord bot's `!inbounds` reply so it stays inside the limits Discord enforces on a message, and truncates the inbound remark the way the rest of the package already truncates field names and values. It counts those limits in the units Discord counts them, and waits out a 429 rather than dropping the pages behind it.

## Why

`sendInbounds` (`internal/web/service/discord/gateway.go`) built one embed holding a field per inbound and sent it unchanged. Discord rejects the **whole message** when it exceeds 25 fields per embed, ten embeds per message, or 6000 characters counted across those embeds, and rejects a field name over 256 characters. Three consequences:

- **An operator past 25 inbounds gets no reply at all.** One inbound per protocol/port is normal on this panel, and the reply is rejected, not truncated.
- **A long remark breaks the command on its own.** The field name is `"📍 " + in.Remark`, and the pin prefix spends four runes of that same 256-rune cap, so a remark over ~252 characters rejects the message.
- **Neither failure leaves a trace.** The send error was discarded (`_ =`), so the channel stays empty and the log stays quiet: the operator cannot tell a broken panel from a broken bot.

Scenario, start to finish: an operator runs 30 inbounds (VLESS-Reality, VMess, Trojan, per-node inbounds) and one of them carries a long descriptive remark. They type `!inbounds` in the configured channel, and the bot answers with nothing — no embed, no error, no log line. Every other `!` command in that channel still works, so the bot looks alive while this one command is dead.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)

## How was this tested?

New `TestInboundsCommandStaysWithinDiscordLimits` drives the real command path — `handleMessage` → `sendInbounds` → the REST client — against an `httptest` server, and asserts the caps Discord documents: at most 25 fields per embed, at most ten embeds, at most 6000 characters per message, field names at most 256 runes, values at most 1024, and **every inbound still listed**.

Two sub-cases:

- 300 inbounds must arrive in more than one message, each within the caps.
- A 400-rune remark must come back truncated to 256 runes and still identify the inbound, not dropped.

The test reddens without the change, with the numbers it was written from:

```
message 0 has an embed with 300 fields, Discord accepts 25
message 0 carries 32195 characters, Discord accepts 6000
300 inbounds reached Discord in 1 message(s), want the reply paged
field name is 402 characters, Discord accepts 256
```

Two more cases in the same file:

- 40 inbounds whose remark is 300 🚀 runes each. The field name cap is applied in UTF-16 units, the way Discord counts; with rune counting the names come back 508 units long and the assertion reads `field name is 508 units, Discord accepts 256`.
- 300 inbounds where the channel answers the first POST with `429 {"retry_after": 0.05}`. The page is waited out and re-sent, and every inbound still arrives (`posted 7 times for 6 messages`); with the retry removed the assertion reads `reply listed 245 inbounds after the retry, want 300`.

The whole `internal/web/service/discord` package suite passes with the change, and `go vet` is clean for the package. The two three-line comments the review flagged are two lines again.

One honest caveat: this machine has no C compiler, so the CGo `mattn/go-sqlite3` driver cannot build here. I ran the suite above through a local pure-Go SQLite build of the driver; CI is the authoritative run.

## Screenshots / recordings

N/A — no UI change.

## Breaking changes

None.

## Checklist


BlindMaster24:
All three findings are addressed on the new head (`2d3f2cd4`).

**MEDIUM — a refused page took the rest of the reply with it, and a 429 was terminal.** The loop no longer returns on a failed page: it logs and keeps sending, so a partial list reaches the operator instead of nothing. A 429 now carries its wait as a typed error (`RateLimitedError`, built from the response's `retry_after` in `parseDiscordResponse`), and a page answered with one is waited out and sent once more — capped at 5s, past which the wait would be worse than the missing page.

New `TestInboundsCommandRetriesARateLimitedPage`: 300 inbounds, the channel answers the first POST with `429 {"retry_after": 0.05}`, and every inbound must still arrive in 6 messages from 7 POSTs. Without the retry it reads `reply listed 245 inbounds after the retry, want 300`.

**LOW (unverified) — rune counting against a cap that may count UTF-16 units.** Settled in the conservative direction rather than left to chance: the budget and the field caps are now counted in UTF-16 units (`discordCharLen`, and `truncateUnits` replacing `truncateRunes` in `subscriber.go`, whose other 22 call sites get the same treatment). Units are never fewer than runes, so the reply fits the cap under either reading of Discord's docs.

New sub-case: 40 inbounds with a 300-🚀 remark. With rune counting the names come back 508 units long and the assertion is `field name is 508 units, Discord accepts 256`.

**LOW — the two three-line comment blocks** are down to two lines each in the same head.

The whole `internal/web/service/discord` package suite passes and `go vet` is clean for it.

PR #6497, opened 2026-09-13
fix(discord): drop the gateway connection when heartbeats go unanswered

BlindMaster24:
## Summary

The Discord Gateway client now closes the connection when a heartbeat goes unacknowledged, so the existing reconnect loop can replace a connection that has silently stopped working.

## Why

`connectAndListen` (`internal/web/service/discord/gateway.go`) sent op 1 every `heartbeat_interval` and did nothing with op 11 — the ACK was a no-op case in the dispatch switch. Discord's Gateway documentation is explicit that a client which does not receive a heartbeat ACK between its attempts is on a failed or "zombied" connection and should close the connection and reconnect.

The case this costs an operator is a half-open socket, and it is the one no other check catches:

- The kernel keeps accepting the heartbeat writes, so `WriteJSON` returns nil — the only error the old code could see.
- The dispatch loop is blocked in `ReadJSON`, which never returns for a peer that is gone but never sent FIN or RST (a NAT/router reboot, a dropped VPN, a provider black-holing the flow).
- Discord's own side eventually times the session out, but nothing tells the panel: `IsRunning()` still reports true, the settings page shows the bot enabled, and every `!` command in the channel goes unanswered until the panel is restarted by hand.

User-visible story: the operator's VPS sits behind a NAT whose mapping expires overnight. The next morning the bot's commands are dead, the panel says the bot is running, and the log is empty — the only fix on offer is restarting the panel. With this change the connection is dropped on the first unanswered heartbeat and the reconnect the code already implements brings the bot back within seconds.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)

## How was this tested?

New `TestGatewayDropsConnectionWhenHeartbeatsGoUnanswered` stands up a real WebSocket server that sends Hello with a 50 ms interval and then never answers a heartbeat — the socket stays open and reads fine, which is exactly what a zombied connection looks like — and asserts the client closes it. Connection close is the observable the fix produces and the only one available: the server sees the client's FIN through `ReadJSON` failing.

- With the change: the connection is closed ~100 ms in, the test passes in 0.4 s, and the log shows `Discord heartbeats went unanswered; dropping the zombied gateway connection` followed by the pre-existing `reconnecting in 5s` line.
- Without it: `gateway held on to a connection whose heartbeats were never acknowledged`, after the full 3 s window.

The whole `internal/web/service/discord` package suite passes with the change, including `TestGatewayRequestedHeartbeatDoesNotRaceTicker`, which drives the same ticker and the server-requested op 1 path that still has to keep working. `go vet` is clean for the package.

The pre-existing `TestGatewayRequestedHeartbeatDoesNotRaceTicker` needed the same treatment: its fake gateway answered no heartbeat at all, which the new check reads as a dead socket, so it now acknowledges heartbeats the way Discord does and paces its op 1 flood. That keeps the socket up for the full second and leaves its concurrent-write window intact; the test passes five runs in a row on the new head.

Caveats, stated rather than glossed: `-race` cannot run on this machine (it needs cgo and there is no C compiler), so CI's `race` job is the check for concurrency here; the ACK flag is a `sync/atomic.Bool` shared by the ticker and the dispatch loop, and `Conn.Close` from the ticker is the method gorilla documents as safe to call concurrently with reads.

## Screenshots / recordings

N/A — no UI change.

## Breaking changes

None.

## Checklist


BlindMaster24:
Both findings are fixed on the new head (`0819f9c3`).

**HIGH — the fake gateway never answered a heartbeat.** `TestGatewayRequestedHeartbeatDoesNotRaceTicker` now behaves like Discord: the server's read goroutine answers each op 1 it receives with op 11, and its two writers share one mutex (gorilla panics on concurrent writes; the race this test pins is the client's). Its `heartbeat_interval` is 10 ms rather than 1 ms, and the op 1 flood paces itself at one frame per millisecond, so the socket survives the whole second and the concurrent-write window is the same second it always was.

Verified on the new head, locally: that test five runs in a row, every other `TestGateway*` test three runs, and the whole `internal/web/service/discord` package once. No `Discord heartbeats went unanswered` line appears in those runs.

The 1 ms interval was the real problem: an ACK has to land before the next tick, and under a flood that saturates the socket a 1 ms window is a coin flip, which is why the head you reviewed went red while a narrower local run passed.

**LOW — the three-line comment block** at `gateway_heartbeat_ack_test.go:31` is down to two lines in the same head.

PR #6498, opened 2026-09-13
fix(discord): report a start-after-first-use client as days, not unlimited

BlindMaster24:
## Summary

`!usage` now renders a client whose expiry is the "Start After First Use" duration as `N Days`, the same way the panel and the Telegram bot already render it, instead of labelling it `Unlimited`.

## Why

`sendUsage` (`internal/web/service/discord/gateway.go`) started from `tr("unlimited")` and only replaced it for a positive timestamp:

```go
expireStr := tr("unlimited")
if client.ExpiryTime > 0 {
    expireStr = time.Unix(client.ExpiryTime/1000, 0).Format("2006-01-02 15:04:05")
}
```

A negative `expiryTime` is not "no expiry" in this panel — it is the *Start After First Use* mode, stored as the duration negated and turned into an absolute deadline by `adjustTraffics` on the client's first traffic tick (`internal/web/service/inbound_traffic.go`, the `UPDATE ... WHERE expiry_time < 0` block). The Telegram bot renders exactly this value as days (`internal/web/service/tgbot/tgbot_client.go`, the `client_ExpiryTime < 0` case), and the frontend edits it as `delayedStart` (`Start After First Use`, `internal/web/translation/en-US.json`). So the Discord embed was the one surface telling an operator that a client which does expire never will.

Operator's story: they hand out 30-day trial keys with "Start After First Use" so the clock starts when the customer first connects. Checking one with `!usage trial@example.com`, the embed reads `Expiry: Unlimited` while the panel's client list shows the same client counting down — the two disagree, and the safest reading of the Discord answer ("this key never expires") is the wrong one.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Xray config generation
- [x] Statistics / traffic counters (client expiry display)
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)

## How was this tested?

New `TestUsageExpiryForDelayedStart` drives `!usage <email>` through `handleMessage` and the REST client against an `httptest` server, and reads the expiry field of the embed that would reach Discord.

- A client stored as `-2592000000` ms — what the panel writes for *Start After First Use: 30 days* — must come back as `30 Days`. Without the change it comes back as `Unlimited`:
  `delayed start shows "Unlimited", want "30 Days"`.
- A client with an absolute deadline (`4102444800000` ms) must still come back as a formatted date, so the new branch does not swallow the case that already worked.

The whole `internal/web/service/discord` package suite passes with the change, and `go vet` is clean for the package.

Caveat, stated rather than glossed: this machine has no C compiler, so the CGo `mattn/go-sqlite3` driver cannot build here; the suite above ran through a local pure-Go SQLite build of the driver, and CI is the authoritative run.

## Screenshots / recordings

N/A — no UI change.

## Breaking changes

None. The value shown for a delayed-start client changes from a wrong label to the one the rest of the panel already uses; no key, setting or endpoint is touched, and `tgbot.days` is reused rather than a new i18n key added.

## Checklist


BlindMaster24:
The pre-existing MEDIUM is confirmed and is fixed in a separate PR: #6500, opened just now.

Your reading of `clientInfoMsg` is exactly right, including the `flag` half I had missed on my own pass: a disabled client takes `diff > 172800 || !traffic.Enable` before `traffic.ExpiryTime < 0` is ever considered, so `time.Unix(-2592000, 0)` becomes `1969-12-02 04:00:00` and the card labels it **Expire Date** rather than **Expire In**.

It went into its own PR rather than this one because it is a Telegram-bot change and this PR is a Discord one — #6500 moves the sign check above the disabled-client branch in `clientInfoMsg` (a three-line reorder that only affects `ExpiryTime < 0`, so every other combination takes the branch it took before) and adds `TestClientInfoShowsStartAfterFirstUseWhenDisabled`, which asserts the card reads `Expire In: 30 Days` for `Enable: false` and `ExpiryTime: -2592000000` (30 days in ms). On the pre-fix code the same assertion renders the 1969 date.

Thanks for naming it — it is the same value read four different ways across the panel, and that card was the one place still getting it wrong.

PR #6499, opened 2026-09-13
fix(tgbot): keep the add-client draft with the chat that owns it

BlindMaster24:
## Summary

The add-client wizard now keeps one draft per chat instead of one package-level draft for the whole bot, so two admins adding a client at the same time cannot fill in each other's client.

## Why

The wizard's state lived in package-level variables — `receiver_inbound_ID`, `receiver_inbound_IDs`, `client_Email`, `client_LimitIP`, `client_TotalGB`, `client_ExpiryTime`, `client_Enable`, `client_TgID`, `client_SubID`, `client_Comment`, `client_Reset` (`internal/web/service/tgbot/tgbot.go`) — while its steps run on the ten-goroutine pool (`messageWorkerPool = make(chan struct{}, 10)`, wired through `runBotHandler` in `internal/web/service/tgbot/tgbot_router.go`). Two admins sharing the bot is the documented case: the bot serves every id in `adminIds`, and the wizard is the one flow that carries state between messages.

What that produces:

- **A client is created from another chat's answers.** Admin A opens the wizard on inbound 3 and sets a 50 GB limit; admin B starts a client on inbound 7 before A submits. `SubmitAddClient` then builds the client from the shared values, so A's "Create" writes B's inbound, traffic limit and expiry — and the links the bot then sends back are for the wrong client.
- **The rendered card lies about the draft.** A's card is re-rendered from the shared state, so the email, limits and attached inbounds shown are the ones the other chat typed.
- **The attach picker mutates one shared slice from several goroutines.** `add_client_toggle_attach` splices `receiver_inbound_IDs` in place, which is a data race, not just a last-writer-wins mistake.

User story: two operators run one panel and both use the Telegram bot (that is what `adminIds` is for). During a bulk client migration they each open Add Client and work through the card at the same time. Each ends up with a client built from a mixture of the two forms — and the share links they forward belong to a client whose limits neither of them chose.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [x] Telegram bot

## How was this tested?

New `TestAddClientDraftIsPerChat` drives the real callback path (`answerCallback` → `addClient` → the bot API) for two chats against an `httptest` bot server, and reads the draft card that reaches each chat:

1. Chat A starts a client (`add_client_to 1`), chat B starts one (`add_client_to 2`). Each card must carry the email the wizard just generated for that chat.
2. Chat A re-renders its card (`add_client_default_traffic_exp`) with chat B's wizard already past its start. A's card must still show A's own email.

With the per-chat keying reverted — one shared draft, the old behaviour — the test reddens exactly on the bug:

```
chat A's card shows email "w91oofmd", want its own "npjv7vys" from chat B's draft
```

The whole `internal/web/service/tgbot` package suite passes with the change, including the two existing tests that drive this wizard (`TestClientDraftMessageRendersHTML`, `TestAddClientPromptsEscapeDraftValues`, updated to seed the draft of the chat they use). `go vet` is clean for the package.

Two things stated rather than glossed:

- `-race` cannot run on this machine (it needs cgo and there is no C compiler), so CI's `race` job is what runs the detector over this. The behavioural test above needs no race detector to fail.
- A single chat's own steps are now serialized, because whatever the chat's draft touches is done under that chat's draft lock. That is deliberate: it also settles a double-tap on the same card. Other chats are unaffected.

## Screenshots / recordings

N/A — no UI change.

## Breaking changes

None for users. `BuildClientDraftMessage`, `SubmitAddClient`, `getInboundsAttachPicker` and `getCommonClientButtons` take the chat's draft as an argument; all four are package-private and used only inside `internal/web/service/tgbot` (checked with a repository-wide grep), so no other package is affected.

## Checklist


BlindMaster24:
All three findings on the locking are addressed on the new head (`da9a4f24`).

**MEDIUM — the lock was held for the whole of `answerCallback`.** It is taken only for an admin wizard step now: `if isAdmin && isAddClientStep(callbackQuery.Data)`. The wizard is the only flow that reads a draft and the non-admin gate rejects everything else, so a report tap never waits on this lock — same chat or not — and never occupies a worker slot while doing it. Matching the raw data is the same as matching the decoded one here: every `add_client*` payload is far under the 64-character `encodeQuery` threshold, and every draft-reading case in the admin branch's `switch dataArray[0]` is an `add_client_*` one.

**MEDIUM — the map grew for every chat and `StopBot` did not clear it.** Entry creation now sits inside the same admin-and-wizard check, so a non-admin tap of a forwarded button — which the gate rejects two lines later — stores nothing. Both submits drop their entry through `addClientDrafts.reset(chatId)`, `add_client_cancel` already did, and `StopBot` calls `addClientDrafts.resetAll()` beside `userStateMgr.reset()`. I did not add a `maybePrune` counterpart, and I want to say why rather than leave it implied: the entry is one small struct per chat that actually started a wizard, `add_client_to` overwrites the fields of the existing struct instead of allocating another, and an idle sweep needs a timestamp plus a retention policy this map has never had. If you want the prune as the user-state store has it, I'll add it — it just looked like more machinery than the leak justifies.

**LOW — `addClient` re-looked-up the draft by chat.** It takes the caller's draft now (`addClient(chatId, draft, msg, messageID...)`) and the eight call sites pass the one they hold. After `add_client_cancel` deletes the entry, the card text and its keyboard come from the same draft, and the keyboard is built under a lock this goroutine actually holds.

New `TestNonWizardCallbackTakesNoDraftLock` pins both halves: with one chat's draft locked by the test, an admin report tap and a non-admin wizard tap must each return within 2s — both block on the head you reviewed — and a chat whose only tap was rejected must have no entry stored.

**LOW pre-existing — `draft.expiryTime = 0` at `tgbot_router.go:614` making the accumulate branch dead.** I left it, and I think it should not be fixed inside this PR: the branch is dead, but `add_client_reset_exp_c 0` is the *Unlimited* button (`tgbot_client.go:1146`), so deleting the reset makes that button preserve the expiry it exists to clear. The labels read "Add 7 days" / "Add 1 month" while the code sets, so reviving the branch is a real behaviour change for everyone who uses those presets — it needs its own PR with its own reasoning rather than a drive-by here. Say the word and I'll write that PR; I would not want to guess which semantics you want.

CI is queued on `da9a4f24`. I'll report `race` explicitly once it lands rather than assuming it, since that is the one result this change is about.

BlindMaster24:
The result I promised to report rather than assume: `race` concluded **success** on `da9a4f24`. 20 of 21 check runs are green (`Publish rolling dev release` is the skipped one), including `go-test`, `golangci`, `codegen`, `frontend`, `govulncheck`, `fuzz-smoke`, `postgres-durable-first` and all nine build legs.

So the detector passes with the new test that locks one chat's draft and drives an admin report tap plus a non-admin wizard tap through `answerCallback` concurrently.

PR #6503, opened 2026-09-13
refactor(tgbot): make the add-client expiry presets say what they do

BlindMaster24:
## Summary

Removes the dead accumulate branch in the add-client wizard's expiry presets, makes the preset labels say what they do, and pins the intent with a test.

## Why

The automated review of #6499 raised this twice as a **LOW pre-existing**: `draft.expiryTime = 0` runs unconditionally at the top of `add_client_reset_exp_c`, so the `if draft.expiryTime > 0` below it tests a value set to zero one line earlier. The branch is unreachable and only `date = 0 - days` ever runs.

I said in that PR I would not fix it there, because reviving the branch is a behaviour change rather than a drive-by, and I did not want to guess which semantics you want. Here is the decision, with the reasoning, so you can disagree with something concrete.

**The wizard sets the term; it does not add to it.**

- A wizard draft is a *new* client, so there is no expiry to add to — `add_client_to` starts every field at its default, expiry at 0.
- The presets and the **custom keypad share this one case**: `add_client_reset_exp_in <n>` types a number and confirms through `add_client_reset_exp_c <n>`. With add semantics, correcting 30 days to 90 would silently produce 120.
- Every other wizard step sets: traffic, IP limit, email, comment, TG user. A preset that accumulated would be the only exception.

**`0` stays the Unlimited button** — it clears the term, which is what the button says.

So the whole reachable behaviour collapses to one assignment, and the code now reads as the decision instead of as an accident. The **renewal** handler (`reset_exp_c`, the same preset row on an existing client) is the opposite case and is untouched: there the client *does* have remaining time, so it adds to the current expiry, and its `if days > 0` guard is what keeps `0` meaning Unlimited.

**The labels follow.** The wizard's preset buttons read **Add 7 Days**, **Add 1 Month**, … while assigning rather than adding, so the operator-facing half contradicted the code. They now read **7 Days**, **1 Month**, **3 Months**, with the term the presets actually set. The renewal keyboard keeps its **Add** — there the word is true. No translation key was added or removed: the wizard rows compose `"7 " + tgbot.days` instead of `tgbot.add + " 7 " + tgbot.days`, and `tgbot.add` stays in use by the renewal row.

### Honest framing

This is a **refactor with no behaviour change**, so the test passes on the base commit too — I am not going to pretend otherwise. What it pins is the decision, and after the review pass it pins it per row: each row starts from a term a preset could have left (`client_ExpiryTime = -7 * 86400000`), so `add_client_reset_exp_c 30`, `90` and `0` each fail on their own if the accumulate branch is revived — the card would read `37 Days`, `97 Days` and `7 Days` instead of `30 Days`, `90 Days` and `Unlimited`. Before that, only the `90` row discriminated, and only because the `30` row had run first.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [x] Telegram bot

## How was this tested?

`TestAddClientExpiryPresetReplacesTheTerm` drives the real `answerCallback` path with a mocked Telegram API, so the assertion is on the card text that would actually be sent: `30 Days` / `90 Days` / `Unlimited`. It installs a go-i18n bundle for the two labels the card prints, because `I18n` returns the bare key with no localizer, and it clears `receiver_inbound_IDs` so the card's attached-inbounds line cannot reach the database from whatever test ran before it — both the expiry global and that slice are saved and restored around the test.

The mocked-Telegram helpers are the package's shared ones (`draftLocalizer`, `draftTexts`), which now register only the messages a test passes in and record `editMessageText` alongside `sendMessage`. They replaced a second, near-identical copy this PR had added, as the review asked.

Behaviour is otherwise untouched: the renewal preset and the custom keypad keep their existing paths.

### Verification

This machine has no C compiler, so the local run substitutes a pure-Go SQLite driver (`modernc.org/sqlite`) behind a `go build -overlay` whose only entries are `internal/database`'s three driver call sites. No source file this PR touches is overlaid, so the handler and the test below are the ones in the diff.

Under that rig:

- `go test ./internal/web/service/tgbot/` — the whole package passes, including the prompt-escaping test that shares the helpers the review asked me to consolidate.
- The test fails on the semantics this PR rejects. With the accumulate branch revived and the zeroing line gone, all three rows go red on their own: the card reads `37 Days`, `97 Days` and `7 Days` instead of `30 Days`, `90 Days` and `Unlimited`.
- With upstream `main`'s handler verbatim, the same test passes — the "no behaviour change" claim above, checked rather than asserted.

`gofmt` and `go vet` are clean on all three files. CI still runs the full gate against the real CGo driver.

## Screenshots / recordings

N/A — Telegram message text, covered by the test above.

## Breaking changes

None. Existing drafts and clients keep the expiry the wizard already gave them.

## Checklist

- [x] I tested the change locally and confirmed the described behavior. (pure-Go driver behind an overlay — see Verification)
- [x] I added or updated tests for the new behavior (when applicable). (pins existing behaviour by design — see above)
- [x] I updated the Wiki / README / API docs if user-facing behavior changed. (no docs describe this wizard)


BlindMaster24:
All three taken — pushed as `9bd4999a` on top of the refactor commit, no force-push.

**Test state (LOW, `tgbot_add_client_expiry_test.go:86`).** You were right: only the `90` row discriminated, and only because `30` had run before it and left the global negative. Every row now starts from a term a preset could have left (`client_ExpiryTime = -7 * 86400000`) inside its own subtest, so each one fails on its own under the accumulate semantics this change rejected — the card would read `37 Days`, `97 Days` and `7 Days` instead of `30 Days`, `90 Days` and `Unlimited`. `client_ExpiryTime` is saved and restored in `t.Cleanup` now too; previously it was left at whatever the last row set, which happened to be the zero value.

**Duplicated helpers (LOW, `:61`).** The second copy is gone. `botPromptLocalizer` / `promptTexts` became `draftLocalizer(t, msgs ...*i18n.Message)` and `draftTexts(t)`, taking only the messages a test needs and recording `editMessageText` alongside `sendMessage`. The prompt test and the expiry test both use them, so the mocked envelope now has one home in the package.

**Labels (LOW pre-existing, `tgbot_router.go:605`).** Folded in, since it is the half of the PR goal the operator actually reads. The wizard presets are now `7 Days`, `10 Days`, `14 Days`, `20 Days`, `1 Month`, `3 Months`, `6 Months`, `12 Months`. No key was added or removed — the rows compose `"7 " + tgbot.days` instead of `tgbot.add + " 7 " + tgbot.days`, and `tgbot.add` stays in use by the renewal keyboard, where `reset_exp_c` genuinely adds to the client's remaining time. A two-line comment at that spot records why the verb is correct there and not here.

CI is re-running on the new head.

BlindMaster24:
Correction to my previous comment, and the evidence I said I could not produce before.

I had been claiming I cannot build or test here because this machine has no C compiler. That is true of the CGo SQLite driver, but not of the code in this PR: with a pure-Go driver (`modernc.org/sqlite`) substituted behind a `go build -overlay` whose only entries are `internal/database`'s three driver call sites, the package builds and runs. Nothing this PR touches is overlaid — the handler and the test executed are the ones in the diff.

- `go test ./internal/web/service/tgbot/` — **whole package passes**, including `TestAddClientPromptsEscapeDraftValues`, which shares the helpers the first review note asked me to consolidate.
- **The test fails on the rejected semantics.** Putting the accumulate branch back (and dropping the zeroing line) makes all three rows fail on their own:

```
--- FAIL: TestAddClientExpiryPresetReplacesTheTerm/30  📅 Expire: 37 Days
--- FAIL: TestAddClientExpiryPresetReplacesTheTerm/90  📅 Expire: 97 Days
--- FAIL: TestAddClientExpiryPresetReplacesTheTerm/0   📅 Expire: 7 Days
```

  That is the point of the start value you asked for: 37 and 97 are the accumulate results from the `-7 * 86400000` a previous preset left, and the `0` row now discriminates too — accumulating would keep the 7 days instead of clearing them, so the Unlimited button silently becomes a no-op.

- **Upstream `main`'s handler verbatim passes the same test**, which is the "refactor, no behaviour change" claim in the body, checked rather than asserted.

`gofmt` and `go vet` are clean on all three files. The PR body now says this instead of the disclaimer it carried before. CI on `9bd4999a` is still the real build; it runs with the CGo driver.

PR #6505, opened 2026-09-13
fix(link): rebuild shadowsocks tcp/http obfuscation on import

BlindMaster24:
## Summary

`ParseLink` now rebuilds the tcp/http response header a shadowsocks link describes through the SIP002 `plugin=obfs-local` parameter. Importing a link the panel had just exported no longer throws the obfuscation away.

## Why

`genShadowsocksLink` exports tcp/http obfuscation **only** as the SIP002 plugin parameter. That is deliberate — `internal/sub/service.go:1303-1310` says why:

```go
// SIP002 clients (v2rayN) ignore the xray-native type/headerType/host/path
// params and only read `plugin`. Re-encode a TCP http header as obfs-local so
// they build a matching tcp/http outbound (v2rayN forces request path "/").
if streamNetwork == "tcp" && params["headerType"] == "http" {
    delete(params, "type")
    delete(params, "headerType")
    delete(params, "host")
    delete(params, "path")
    params["plugin"] = "obfs-local;obfs=http;obfs-host=" + host
}
```

The importer read none of it. `parseShadowsocks` passed only `type`/`security`/`sni`/`alpn`/`fp` into `applyTransport`, and the emitter had just deleted four of those, so a link round-tripped through the panel lost the one thing that made it work.

Reproduced against the current `main` with the export path itself, not a hand-written string:

```
exported: ss://YWVzLTI1Ni1nY206Y2xpZW50cGFzcw@203.0.113.1:8388?plugin=obfs-local%3Bobfs%3Dhttp%3Bobfs-host%3Dobfs.example.com#user
imported: {"network":"tcp","security":"none","tcpSettings":{"header":{"type":"none"}}}
```

The header is `none`. The inbound behind that link speaks `tcp`/`header.type: http`, so the imported outbound cannot connect to it.

**Operator scenario.** Someone runs a 3x-ui panel with a shadowsocks inbound using TCP + HTTP-header obfuscation (a common way to make an SS inbound look like plain HTTP). They hand out the share link, or point a second 3x-ui node at it with an external subscription. Both import paths end in `link.ParseLink` — `internal/sub/external_config.go:216` for a pasted link, `internal/web/service/outbound_subscription.go:394` for a fetched subscription. What lands in the second panel is a plain-TCP SS outbound: it saves fine, shows no error, and simply never connects. The same link in v2rayN works, because v2rayN reads `plugin`, so the first panel looks correct and the second one looks broken. The other direction — the panel's own JSON/Clash output — never goes through `ParseLink` and is unaffected.

This is the same class as #6467, which preserved the shadowsocks TLS query params on import in this function; the plugin parameter was the remaining unread one.

## What changed

`applyObfsLocalPlugin` maps `plugin=obfs-local;obfs=http;obfs-host=<host>` onto `type=tcp`, `headerType=http`, `host=<host>` — exactly the four keys the emitter deleted, so the existing `applyTransport` branch rebuilds `tcpSettings.header` (with `path: ["/"]`, matching the emitter's note that SIP002 clients force `/`).

- Only `obfs=http` is mapped. `obfs=tls` and third-party plugins (`v2ray-plugin`, …) are left alone, because Xray has no header that corresponds to them and guessing would be worse than the current silence. Both cases are asserted.
- An explicit `headerType`/`type=http` in the link wins; the helper returns early rather than overwriting it.
- Credentials, `mport` handling, the `identity` string (and therefore import dedup) and every other protocol are untouched.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot

## How was this tested?

Two tests, and both were watched going red without the fix (by running them against `main`'s `outbound.go`):

- `internal/sub/shadowsocks_plugin_import_test.go` — the round trip, driven through the real exporter: `genShadowsocksLink` on a tcp/http inbound, then `link.ParseLink` on the result. Asserts `header.type == "http"` and the `obfs-host` survives. Without the fix it fails with `{"header":{"type":"none"}}`.
- `internal/util/link/outbound_test.go` — `TestParseShadowsocksObfsLocalPlugin`, a table over the three plugin shapes: `obfs=http` becomes the header, `obfs=tls` and `v2ray-plugin` stay `none`. Without the fix the first row fails with `header = map[type:none], want type "http"`.

Local run: this machine has no C compiler, so the Go tests ran with a pure-Go SQLite driver (`modernc.org/sqlite`) substituted behind `go build -overlay`, whose only entries are `internal/database`'s driver call sites. `go test ./internal/util/link/ ./internal/sub/` — both green; `gofmt` and `go vet` clean on both packages. The existing shadowsocks cases in `TestParseShadowsocks`, including the two that already carried a `plugin=` query, still pass; they only asserted credentials, so the new mapping is the behaviour change they never pinned.

## Screenshots / recordings

N/A — wire format, covered above.

## Breaking changes

None.

## Checklist

- [x] I tested the change locally and confirmed the described behavior. (pure-Go driver behind an overlay, see above)
- [x] I updated the Wiki / README / API docs if user-facing behavior changed. (no docs describe this import detail)


BlindMaster24:
Both the HIGH and the LOW are fixed in `0573de86`; the MEDIUM pre-existing I am leaving alone, see the end.

**HIGH — the two importers disagreed.** Correct, and my fault: I grepped the frontend for the emitter and missed `parseShadowsocksLink` in `frontend/src/lib/xray/outbound-link-parser.ts`, which is what the Add Outbound button calls. It now maps the plugin through `applyObfsLocalPluginParams` before the transport is applied, so it lands on the same `tcpSettings.header` branch the file already had (`outbound-link-parser.ts:227-238`). `pcs`-style shape differences don't apply here — the mapping only sets `type`, `headerType` and `host`, which the TS side reads the same way.

Two tests in `frontend/src/test/outbound-link-parser.test.ts`: the positive one fails on `main`'s parser with `AssertionError: expected { type: 'none' } to match object { type: 'http', request: { …(1) } }` and passes with the mapping; the negative one keeps `obfs-local;obfs=tls` at `type: 'none'`, so a future "map every plugin" change goes red. `npx vitest run src/test/outbound-link-parser.test.ts` — 56 passed; `tsc --noEmit` and `oxlint` clean on both files.

**LOW — confirmed, and now handled.** I reproduced it exactly as you described rather than taking it on faith: with the first commit's parser, `ss://…?plugin=obfs-local;obfs=http;obfs-host=obfs.example.com` imports as `header = map[type:none]`, because stdlib discards the pair. `applyObfsLocalPlugin` now falls back to reading `plugin` straight out of the raw query when the parsed value is missing, and the table has a row for the unencoded shape: red against the first commit, green now. The encoded form, which the panel's own export produces, is unchanged and still covered.

**MEDIUM pre-existing — the Clash external-link branch.** You are right that `clashProxyFromExternal` returns before the transport/security calls for `ss` and emits a bare proxy. I am not touching it here: it is pre-existing, and unlike this PR it is not a case of "read back what we wrote" — Clash has no `tcp`/`header.type` key, so the entry needs `plugin: obfs` with `plugin-opts`, which is a mapping decision rather than the same fix. If the maintainer wants that, it is a separate PR: the alternative (`clash_service.go:1000-1002`'s native-inbound behaviour of omitting the entry) is a different answer to the same question and someone should pick one.

Local evidence for the Go half is unchanged from the PR body; the full frontend gate (`npm test` with its Storybook project) is CI's, since the Chromium that project needs is not installed on this machine.

PR #6507, opened 2026-09-13
fix(link): read the vmess certificate checks on import

BlindMaster24:
## Summary

`parseVmess` now reads `ech`, `vcn` and `pcs` from the vmess share object. Importing a vmess link the panel had just exported no longer discards the certificate checks it carries.

## Why

`applyVmessTLSParams` writes six TLS fields into the share object — `sni`, `fp`, `alpn`, `ech`, `vcn`, `pcs` (`internal/sub/service.go:1797-1821`) — while `parseVmess` read back only the first three. The same three the url-param protocols read in `applySecurity` (`internal/util/link/outbound.go:711-714`) were the ones missing here, so vmess was the only protocol that lost them.

Reproduced through the exporter itself on current `main`:

```
exported: {"sni":"vmess.example.com","fp":"chrome","alpn":"h2","ech":"AEX+DQBB","vcn":"vcn.example.com","pcs":"AAAA…=",…}
imported: {"alpn":["h2"],"echConfigList":"","fingerprint":"chrome","pinnedPeerCertSha256":"","serverName":"vmess.example.com","verifyPeerCertByName":""}
```

All three arrive empty.

**Operator scenario.** A vmess inbound whose certificate is trusted only through a pin — the case the panel calls out as the replacement for the removed `allowInsecure` (`verifyPeerCertByNameValue`'s own comment) — exports that pin as `pcs`. The subscriber pastes the link into another panel, or a second node pulls it as an external subscription (`internal/sub/external_config.go:216`, `internal/web/service/outbound_subscription.go:394`). The pin is gone, so the outbound verifies against the system roots instead: it saves without a warning and cannot reach the server. The vless/trojan/ss equivalents keep their pins today, so the same certificate setup works on every protocol except vmess. xray-core has also dropped `allowInsecure` in favour of exactly these fields (`infra/conf/transport_security.go:362`), which is what makes losing them harder to work around in the client.

Same class as #6467, which preserved the shadowsocks TLS query params on import, and as #6505 for the shadowsocks `plugin` parameter; this is the last of the three the panel exports and the importer never read.

## What changed

Three assignments in the vmess TLS block, matching `applySecurity` exactly:

- `ech` → `echConfigList`
- `vcn` → `verifyPeerCertByName`
- `pcs` → `pinnedPeerCertSha256`

`pinnedPeerCertSha256` is filled as the joined string the core reads (`infra/conf/transport_security.go:314` declares it as a `string`), which is also the shape the url-param path and `SubJsonService` already write. A link that carries none of the three gets empty values, the same as before. Nothing else about vmess parsing changes, and the other protocols are untouched.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot

## How was this tested?

`internal/sub/vmess_tls_import_test.go`: `genVmessLink` on a vmess inbound whose `tlsSettings.settings` carries a fingerprint, an ECH config, a verify-by-name and one pin, then `link.ParseLink` on the result, asserting all five fields plus `alpn`. Run against `main`'s `outbound.go` it fails on exactly the three fields this PR adds:

```
tlsSettings["echConfigList"] = , want "AEX+DQBB"
tlsSettings["verifyPeerCertByName"] = , want "vcn.example.com"
tlsSettings["pinnedPeerCertSha256"] = , want "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA="
```

Local run: this machine has no C compiler, so the Go tests ran with a pure-Go SQLite driver (`modernc.org/sqlite`) substituted behind `go build -overlay`, whose only entries are `internal/database`'s driver call sites — no file this PR touches is overlaid. `go test ./internal/util/link/ ./internal/sub/` — both green; `gofmt` and `go vet` clean on both packages.

## Screenshots / recordings

N/A — wire format, covered above.

## Breaking changes

None.

## Checklist

- [x] I tested the change locally and confirmed the described behavior. (pure-Go driver behind an overlay, see above)
- [x] I updated the Wiki / README / API docs if user-facing behavior changed. (no docs describe this import detail)


BlindMaster24:
You are right, and this is the part I got wrong: I grepped the frontend for the emitter only and concluded the TS side had no importer. `parseVmessLink` in `frontend/src/lib/xray/outbound-link-parser.ts` is the second one, and it is the one the Add Outbound button calls.

Fixed there too, in `66219f9a`, so both importers agree again:

```ts
if (typeof json.ech === 'string') tls.echConfigList = json.ech;
if (typeof json.vcn === 'string') tls.verifyPeerCertByName = json.vcn;
if (typeof json.pcs === 'string') tls.pinnedPeerCertSha256 = json.pcs;
```

The `typeof` guards mirror Go's `getString`, so a non-string in the JSON leaves the seeded `''` alone. The string form for `pcs` matches the same file's url-param path (`applySecurityParams`), which is what `buildStream` seeds and what the core reads.

`Test keeps the ech, vcn and pcs certificate checks` in `frontend/src/test/outbound-link-parser.test.ts` covers it. Watched red first: with `main`'s `outbound-link-parser.ts` restored and the test kept, it fails on all three fields; with the change it passes. `npx vitest run src/test/outbound-link-parser.test.ts` — 55 passed; `tsc --noEmit` and `oxlint` clean on both files. I could not run the repo-wide `npm test` here — it includes the Storybook project, whose Chromium is not installed on this machine — so the full gate is CI's.

The Go half and the local evidence for it are unchanged from the PR body.

PR #6508, opened 2026-09-13
fix(sub): drop external Clash shadowsocks nodes the panel cannot express

BlindMaster24:
## Summary

`clashProxyFromExternal` returned as soon as it had built the shadowsocks proxy, so an `ss://` link skipped the transport/security mapping that every other protocol — and the inbound path — goes through.

## Why

Pasting an `ss://` link into an external Clash subscription produced a proxy built from `type/server/port/cipher/password` alone. Two user-visible consequences:

**1. TCP/HTTP obfuscation was ignored, and the node was not dropped.** mihomo cannot express an xray `tcpSettings.header.type=http` on a shadowsocks proxy, which is why `applyTransport` returns false and the inbound path (`buildProxy` → `applyTransport` → `getProxies`) *omits* such a node: the YAML ships with a working set of proxies. The external-link path emitted it anyway, so mihomo opens a plain shadowsocks stream at a server that requires the HTTP header. The node appears in the subscription, the client shows it, and it silently never connects — the same failure mode as #6505, one layer further along. Pasting the identical node as an inbound and as a link produced two mutually inconsistent configs.

**2. TLS was silently stripped.** For `security=tls` the link path never called `applySecurity`, so the proxy carried no `tls: true` / `servername` — a plaintext stream at a TLS listener.

**3. And once the node is dropped, it took the client's quota with it.** The client email that feeds `AggregateTrafficByEmails` was recorded only when a link produced a proxy (`clash_service.go:91-93`), so a dropped node also disappeared from the `Subscription-Userinfo` header — a two-client subscription served A's numbers as the whole subscription's numbers. The `!ext.Active` branch already counted an email without a proxy, so the rule is now unconditional; that fixes the same under-report for the vmess/vless/trojan links that were already droppable (`?type=kcp`, unknown `security=`).

The panel's own `genShadowsocksLink` writes obfs as `plugin=obfs-local;obfs=http;obfs-host=…` (SIP002, pinned by `TestShadowsocksTcpHttpHeaderUsesObfsLocalPlugin`). #6505 maps that plugin onto the tcp/http header on import, so both spellings of the same link now reach this gate. The xray-native spelling (`?type=tcp&headerType=http`) reaches it on `main` today, which is what the test below uses, so this PR stands on its own.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot

## How was this tested?

New `TestClashExternalShadowsocksMatchesInboundPath` builds the same shadowsocks node twice — once as an inbound through `buildProxy`, once as a share link through `clashProxyFromExternal` — and requires the two results to agree, for three links:

| link | expected |
| --- | --- |
| `?type=tcp` (plain) | both build the same proxy |
| `?type=tcp&headerType=http&host=test` | both dropped |
| `?type=tcp&security=tls&sni=ss.sni` | both build the same proxy, with TLS |

Before the fix, all three rows fail — the interesting one is the obfs row, where the inbound path returns `nil` and the link path returns a proxy:

```
tcp_http_header: not representable in clash, want both dropped:
  inbound map[string]interface {}(nil)
  link    map[string]interface {}{"cipher":"aes-256-gcm", "name":"external",
          "password":"clientpw", "port":8443, "server":"203.0.113.1",
          "type":"ss", "udp":true}
plain tcp: inbound {… "network":"tcp", "tls":false …} != link {… no network/tls …}
tls:       inbound {… "servername":"ss.sni", "tls":true …} != link {… no tls …}
```

The fix is the `ss` case no longer returning early, so it falls into the shared transport/security tail; all three rows then pass, with the link and inbound proxies deep-equal apart from `name`.

`TestClashQuotaHeaderCountsDroppedExternalLink` drives the whole HTTP path: a subscription with client A on an inbound (11/22/1024) and client B on an obfs `ss://` link (100/200/2048). Without the quota change the header is `upload=11; download=22; total=1024` — A alone — and with it `upload=111; download=222; total=3072`. `TestClashAllUnrepresentableNodesAnswerAlike` pins that a subscription with nothing representable answers the same way whether the node came from an inbound or an external link (it compares the two responses instead of asserting a particular status).

- `go test ./internal/sub/` — ok (39s)
- `go test ./internal/util/link/` — ok
- `go vet ./internal/sub/` — clean
- `gofmt -l internal/sub/ internal/util/link/` — clean

There is no C toolchain on this machine, so the CGo SQLite driver cannot build here: I ran the packages with `CGO_ENABLED=0` plus an overlay replacing the three CGo SQLite call sites in `internal/database` with a pure-Go driver. Every file in this PR is the real one, unmodified by the overlay. CI stays the authority on `go build ./...` with the real driver.

## Screenshots / recordings

N/A (no UI change).

## Breaking changes

None for existing working subscriptions. Two behaviour changes worth stating plainly:

- An external `ss://` node whose transport Clash cannot express is now dropped from the profile, exactly as the same node as an inbound always has been. A subscription whose *only* node was such a link answers 404 with an empty body instead of serving a profile with one non-connecting proxy — the same answer the inbound path gives for the same node, and what an external vmess/vless/trojan link with `?type=kcp` already gets on `main`. Clients keep the profile they last fetched.
- `security=tls` shadowsocks links gain `tls: true` / `servername`, and a dropped external node still contributes to `Subscription-Userinfo`. The quota header of a mixed subscription changes (up) as a result.

## Checklist

- [ ] `go build ./...` and the test suite pass locally. — no C compiler here, ran with the cgo-free overlay described above; CI verifies the real build.
- [ ] I updated the Wiki / README / API docs if user-facing behavior changed. — N/A, no endpoints or settings changed.


BlindMaster24:
Good catch, and the mechanism is exactly as you describe — I verified it instead of arguing. With the drop in place and the bookkeeping unchanged, a two-client subscription (A on an inbound, B on an obfs `ss://` link) served:

```
Subscription-Userinfo = "upload=11; download=22; total=1024; ..."   <- A's numbers alone
want                    "upload=111; download=222; total=3072; ..."  <- A's plus B's 100/200/2048
```

Fixed in `41921df3`: the email is now recorded for every external link instead of only for links that produced a proxy. The inactive-link branch already did exactly that (`if !ext.Active { seenEmails[ext.Email] = … }`), so this makes the rule unconditional rather than special-casing one kind of dropped node — and it fixes the same under-report for a vmess/vless/trojan link with `?type=kcp` or an unrecognised `security=`, which is reachable today.

Two tests, both in `internal/sub/clash_external_quota_test.go`:

- `TestClashQuotaHeaderCountsDroppedExternalLink` — red without the bookkeeping fix (`upload=11; download=22; total=1024`), green with it (`upload=111; download=222; total=3072`).
- `TestClashAllUnrepresentableNodesAnswerAlike` — the point I want to make about the empty case. An inbound-only subscription whose only node is not representable **already** answers 404 with an empty body; the external-link-only subscription now answers identically. The test compares the two statuses rather than asserting `404`, so it keeps holding if the empty case is ever changed.

On the rest of your note — "if the dropped link is the subscription's only entry … answers 404" — that outcome is not new in kind: it is what the inbound path does for the same node, and it is reachable on `main` today for an external vmess/vless/trojan link with `?type=kcp` or an unknown `security=`. So this change makes `ss` join an existing, deliberate behaviour rather than inventing one: a subscription with zero representable nodes has never had anything to serve, and mihomo keeps the last profile it fetched. What was *not* deliberate, and is now fixed, is that the drop also corrupted the quota of the nodes that remain — that is the part of your finding I've acted on, and it applies to the pre-existing drops too.

I've updated the PR description's "Breaking changes" and "How was this tested?" sections to cover both effects instead of only the drop.

BlindMaster24:
Both remaining points, on `8dce470f`:

**LOW — stale doc comment.** Taken, and shortened while fixing it — the enumeration was what went stale, so the comment now says what the function does instead of listing which protocols go down which path:

```go
// clashProxyFromExternal converts a pasted share link into a mihomo/Clash proxy
// entry, or nil when Clash can't represent it — the same gate getProxies runs.
```

That is also the two-line cap `CLAUDE.md` sets; the previous block was five lines and had to be edited the moment the `shadowsocks` branch changed, which is the usual cost of restating a `switch` in prose.

**MEDIUM pre-existing — the panel's own `ss` obfs spelling.** Agreed on every step, including the merge-order sentence: `parseShadowsocks` never reads `plugin`, so until #6505 lands, `ss://…?plugin=obfs-local;obfs=http;obfs-host=…` still parses to `header.type=none`, walks past this gate, and ships the plain proxy this PR describes. #6505 maps that plugin onto the same tcp/http header, at which point the two spellings produce the same parsed stream and this gate covers both. The PR description already names #6505 as the dependency; I've left the two PRs independent in code so either can merge first, and the test here uses the xray-native spelling precisely so it does not depend on #6505 being in the tree.

The MEDIUM in your main review is fixed in `41921df3` — the email is now recorded for every external link rather than only for links that produced a proxy, with `TestClashQuotaHeaderCountsDroppedExternalLink` red on `upload=11; download=22; total=1024` and green on `upload=111; download=222; total=3072`.

PR #6510, opened 2026-09-13
feat(tgbot): add /broadcast to relay an admin message to all clients

DIMFLIX:
## What

An admin-only `/broadcast` command for the Telegram bot. The admin sends any message — text, rich text, photos, a video, a file, a sticker or a whole album — confirms it with an inline button, and the bot delivers an exact copy of it to every client that has a Telegram ID linked, on behalf of the bot.

## Why

There is currently no way to reach all clients at once: notifications only cover exhausted quotas, so an operator who wants to announce maintenance or a new feature has to copy the message to each client chat by hand.

## How it works

1. `/broadcast` asks for a message (one broadcast at a time; a second command is refused while one runs).
2. The admin's message becomes the broadcast **by reference** (admin chat + message ids) — the bot never parses the content. A preview self-copy shows the admin exactly what recipients will get and rejects content Telegram cannot copy before the run starts.
3. Delivery runs in a background goroutine: sequential `copyMessage` sends with a small pause, `429 retry_after` honored per recipient, failures counted without stopping the run, progress edited into one card at most every 25 sends or 3 seconds, and a cancel button checked between sends. The final delivered/failed/skipped summary is edited into the card so it is never duplicated.
4. Recipients repeat the `notifyExhausted` walk: clients with a linked `tg_id`, deduplicated, admins excluded — they already receive the reports. The message content is never logged.

### Design note: copy instead of content parsing

The first draft of this feature parsed the admin's message into a draft (text/media kind/file id) and re-sent it with `sendMessage`/`sendPhoto`/... That broke on the new `rich_message` format, would need a new parser for every Telegram message type, and ran into caption length limits. `copyMessage`/`copyMessages` avoid all of that: content arrives 1:1 with formatting preserved, future message types work for free, and — unlike `forwardMessage` — there is no "Forwarded from" header, so the admin's identity stays private.

Albums arrive as separate updates sharing a `media_group_id`; their ids are buffered with a short debounce, sorted (`copyMessages` requires strictly increasing ids), and delivered as one `copyMessages` call so recipients see the original album.

## Notes / limitations

- Telegram only delivers to chats where the client has started a dialog with the bot; everything else is counted as "failed" in the summary.
- Interactive rich-message blocks (buttons, maps) that a client may include are delivered as-is by Telegram's copy; the bot adds no controls of its own to the relayed content.

## Testing

- New `tgbot_broadcast_test.go` (stdlib `testing`, table-driven): recipient collection against a real DB (dedup, admin exclusion, `tg_id == 0`), run counters (delivered/failed), mid-run cancel, 429 retry pauses via a swappable sender, single-slot runner, admin-only command and callbacks, album buffering with out-of-order updates, wire assertions for `copyMessage`/`copyMessages` bodies, and an end-to-end confirm through `answerCallback` against a mock API.
- `go build ./...`, `go test ./internal/web/service/tgbot/ -count=1`, `make test-go`, `golangci-lint run` all pass.
- The new i18n keys are present in all 13 locales and referenced from the Go sources.


DIMFLIX:
Thanks for the review — all six findings are addressed in 088a32e2:

1. **MEDIUM (stale preview delivers the wrong draft)** — each preview card now carries a random token (`broadcast_confirm <token>`), and `broadcastTakePending` only releases the draft whose token matches. A stale Send tap is answered with an error instead of delivering a newer, unapproved draft.

2. **MEDIUM (process-wide draft and buffer)** — the pending draft and the album buffer are now a per-chat `broadcastComposes` map, the same ownership model the add-client wizard uses. Two admins compose independently; one chat's `/broadcast` or album no longer touches another chat's state.

3. **MEDIUM (sender never checked)** — `handleBroadcastInput` now verifies `checkAdmin(message.From.ID)` and silently ignores anyone else, so in a group only the admin's messages can become the broadcast content.

4. **LOW (cancel not observed during flood waits)** — `broadcastDeliverOne` now takes the cancel check and abandons the recipient (counted failed) instead of sleeping through up to 5 `retry_after` waits after cancellation.

5. **LOW (per-recipient pacing vs album size)** — the pause is scaled by the copied batch size (`broadcastSendDelay * len(ids)`), so a 10-item album keeps the same per-second ceiling as a single message.

6. **LOW (comment blocks over the 2-line cap)** — all trimmed to two lines.

Tests added for each of 1–5: stale-token confirm rejected, per-chat isolation (two chats compose and confirm independently), non-admin input ignored, cancel during a flood wait abandons immediately, and album pacing (one pause of `3 × broadcastSendDelay` for a three-message album). Existing tests were updated for the per-chat/token API. Full `make test-go` and `golangci-lint run` pass.

BlindMaster24:
Thanks — the authorization side here is right, and I want to say that first because I checked it specifically: `/broadcast` and both callbacks (`broadcast_confirm`, `broadcast_cancel`) sit inside the `isAdmin` branch of `answerCommand`/`answerCallback`, so a group member who happens to see the preview card cannot fire the broadcast. The pacing is right too: 60 ms × messages ≈ 16 msg/s against Telegram's ~30 msg/s bulk ceiling, `retry_after` read from `telegoapi.Error`, progress edits throttled to ≥3 s (0.33 msg/s into one chat, well under the 1 msg/s per-chat limit). Three things I would fix before this merges.

**1. `StopBot` leaves the broadcast state behind**

`StopBot()` clears `isRunning`, then `userStateMgr.reset()` and `addClientDrafts.resetAll()` — the latter landed in #6499 for exactly this class of leak. Nothing clears `broadcastComposes` or `broadcastActive`, and the only change this PR makes outside its own new files is the command registration in `tgbot.go`.

Consequences:

- `broadcastComposes[chatID].timer` (the 900 ms media-group debounce) is still armed after the bot stops, so it can fire `finalizeBroadcastAlbum` → `acceptBroadcastDraft` and attempt a self-copy through a stopped bot.
- The pending token survives the restart, and `broadcastTakePending` compares only the token — so a Send button from the previous session is still confirmable. No injection (the content is still the admin's own message and recipients are re-collected at confirm time), but it is state outliving the thing that created it.

Suggestion: a `broadcastResetAll()` that stops each compose timer, clears the map and cancels/releases the runner, called next to `addClientDrafts.resetAll()` in `StopBot`.

**2. A long flood wait parks the single-runner slot**

`broadcastDeliverOne` honours `retry_after`, but `broadcastPause` is a plain `time.Sleep` and `canceled()` is checked only *before* the wait; `runBroadcast`'s loop checks `runner.cancel` and `IsRunning()` before each recipient, never during a pause. So once a long wait starts, nothing ends it: Cancel sets a flag nobody reads again until the sleep finishes, and after a bot restart `/broadcast` answers `broadcastAlreadyRunning` because `broadcastActive` is still held by the parked goroutine. Telegram does not bound `retry_after` — a bot that has tripped the global flood limit gets waits measured in minutes.

Suggestion: sleep in slices (≤5 s), re-checking `canceled()` and `IsRunning()` between them, or clamp the wait and re-enter the retry loop.

**3. `403` is counted as a delivery failure**

`collectBroadcastRecipients` selects on `client.TgID != 0`. In 3x-ui that field can be typed into the panel by the admin, and no column records whether the chat ever talked to the bot — so a linked id is not evidence the user pressed Start. Telegram answers `403: bot can't initiate conversation with a user` for those, and `403: bot was blocked by the user` for the rest. Both currently increment `failed` and emit one `logger.Warning` per recipient: on a few thousand clients that is the log-spam class #6340 removed, and the summary shows `Failed: N` with no reason.

Suggestion: split the counters — the two 403s as `Skipped` (logged at Debug, maybe a sampled few), everything else (5xx, network, flood retries exhausted) as `Failed` — plus one line in the summary saying those recipients have to start the bot first.

Optional, not a defect: nothing persists which recipients were already reached, so a re-run after a crash or restart re-sends to everyone who got the message the first time. A per-run offset would make it resumable, if that is worth the extra state to you.

BlindMaster24:
One housekeeping note, so nobody re-litigates stale findings: the three earlier **MEDIUM**s and the "paced per recipient, not per message" **LOW** from the previous review are already addressed on this head (`088a32e2`):

- composition state is per chat now (`broadcastComposes map[int64]*broadcastCompose`), not the single `broadcastPending`/`broadcastBuf` globals, so two admins cannot destroy each other's draft;
- the confirm callback carries a per-card token and `broadcastTakePending(chatId, token)` matches it, so a stale card's Send no longer takes a newer draft;
- `handleBroadcastInput` checks `message.From` through `checkAdmin` before accepting a message, which closes the "a non-admin can supply the content in a group" path;
- the pause scales with the draft (`broadcastSendDelay * max(1, len(draft.MessageIDs))`), so an album is paced per *delivered message*, which is what the earlier LOW asked for.

Still open from that review: the flood-wait cancellation (my point 2 above) and the `CLAUDE.md` 2-line comment cap. Plus the two in my comment above that I have not seen raised anywhere: the `StopBot` state leak and the `403` classification.

DIMFLIX:
Thanks — all four are in `e4b2a13b`.

1. **`StopBot` state leak** — added `broadcastResetAll()` (stops every compose timer, clears the map, cancels and releases the active runner) and wired it into `StopBot` right next to `addClientDrafts.resetAll()`. `TestStopBotResetsBroadcastState` arms an album timer, registers a runner, calls `StopBot`, then asserts the slot is free, the run is cancelled, no pending draft survives, and the timer never fired (no `copyMessage`/`sendMessage` after the debounce window).

2. **Flood wait parking the slot** — `broadcastFloodWait` now sleeps `retry_after` in `broadcastFloodWaitSlice` (5 s) slices, re-checking an `aborted()` closure (`runner.cancel || !IsRunning`) between slices; an abort returns a sentinel that `runBroadcast` maps to a clean "canceled" break instead of counting the recipient failed. `TestBroadcastFloodWaitSlicesLongWaits` covers both directions: a 30 s wait becomes six 5 s slices when uninterrupted, and one slice when aborted after the first check.

3. **403 as failure** — 403 now increments `Unreachable`/`Skipped` (logged at `Debugf`, one per recipient instead of warning-spam), while 5xx/network/exhausted-flood-retries stay `Failed`. The summary appends one new `tgbot.messages.broadcastUnreachable` line naming the count and the reason (key added to all 13 locales). `TestRunBroadcastCounters` asserts the split and that the note appears only when there are unreachable recipients.

4. **Comment cap** — re-audited both files mechanically (no block over two lines) rather than by eye this time; the remaining offenders were in `runBroadcast`, `broadcastDraft`, `broadcastCompose`, `collectBroadcastRecipients`, `deliverBroadcastCopy`, `finalizeBroadcastCard`, `confirmBroadcast` and the test helper.

Each of 1–3 verified by mutation: reverting the classification, the slicing, or the `StopBot` call each turns its test red. `make test-go` (44 packages), `golangci-lint run`, locale parity and dead-key checks all pass.

On the optional resumability note: agreed and deliberately skipped — a durable per-recipient cursor means new persisted state and a second source of truth for a broadcast, which is more than this feature's scope.

PR #6514, opened 2026-09-13
feat(settings): randomize subscription port on fresh installations

### LINKED ISSUE #6447: port 2096

is port 2096 default for subs? does every install have it open?
2096 was the only low port on my vps and it got ip blocked in russia.

make it random

_No response_

### Concrete use case

_No response_

_No response_

_No response_

### Are you willing to help?

### Before submitting

poise52:
## Summary
Randomize the default subscription port (`subPort`) within the safe non-privileged range (`1024-62000`) on fresh installations, while ensuring 100% backward compatibility for existing installations.
## Why
Using a static default port (`2096`) makes subscription endpoints easy targets for automated ISP port scanning and censorship blocking (e.g., TSPU/RKN in Russia) and can cause port collisions on single-VPS setups. Generating a distinct random port on fresh installs mitigates fingerprinting and port conflicts without breaking existing setups.
Fixes #6447
## Type of change
- [ ] Bug fix
- [x] New feature
- [x] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other
## Areas affected
- [ ] Xray config generation
- [x] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot
## How was this tested?
1. **Backend & DB tests**:
   - `go test -v -run "TestInitDB" ./internal/database/...` — verified that fresh databases receive a random `subPort` in `[1024, 62000]` and that existing databases preserve their existing port unchanged.
   - `go test -v ./internal/util/random/...` — verified port generator range and distribution.
   - `go test -v -run TestRouteRegistryContract ./internal/web` — verified route registry contract.
   - `go build ./...` — compiled cleanly without errors.
2. **Frontend & Docs contracts**:
   - `cd frontend && npx vitest run src/test/factory-defaults-contract.test.ts` — verified factory default contracts pass (138 tests).
   - `cd docs && npm run format:check && npm run lint && npm run typecheck && npm test && npm run build` — all 269 documentation pages across EN, RU, ZH, FA compiled cleanly.
3. **Collision prevention**:
   - Verified that both `install.sh` and DB seeder guarantee `subPort != webPort`.
   - Verified CLI `x-ui setting -subport` and `updateSetting` reject matching web and sub ports.
## Screenshots / recordings
N/A (Backend, database seeding, CLI, and installation script changes).
## Breaking changes
None. Existing databases are never modified (`FirstOrCreate` is gated by `isUsersEmpty`), and fallback factory defaults remain `2096`. Existing subscriptions continue to work seamlessly.
## Checklist


BlindMaster24:
One behaviour worth settling before this lands, plus a small robustness note.

**`XUI_SUB_PORT` silently outranks the panel setting.** `GetSubPort()` returns the env value whenever it is set and valid, so a port changed in the panel UI or through the API no longer affects the port the subscription server actually binds — while `x-ui setting -show true` (and the summary the installer prints) keeps showing the database value. An operator who exports the variable once and later edits the field in the panel gets a silent disagreement between what the panel displays and what it listens on. The invalid-value case is handled well (warning + fall back to the configured port); it is the valid case that wins forever.

Suggestion: either surface the override (a hint or read-only state in the panel while it is active, or a single startup warning when the two values differ), or document the precedence right next to the setting so the field is not mistaken for the source of truth.

**Reading the existing port back for the upgrade summary.** `existing_subport=$(x-ui setting -show true | grep -Eo 'subPort: .+' | awk '{print $2}')` scrapes human-formatted output. It works today, but it reads the wrong thing the moment that line's format changes; if there is already a machine-readable way to read one setting, prefer it.

Parts I checked and liked, for the record: the explicit port is validated (1–65535, and rejected when it collides with the panel port), `-subport` is only passed when the installed binary advertises it, so upgrading a build that predates the flag does not break, and both `openapi.json` copies are byte-identical at this head (sha256 `ea48b68a84ffb8d53721b2681e70ee8a5c26c2bea9d6dca21f4d5d60f72470f2`), so the docs-copy rule is satisfied.

BlindMaster24:
One more thing on the current head (`05433281`), and I want to be clear it is a *consequence* of the fix for the earlier `-subport` finding, not a reopening of it.

The probe now guards the flag — `if ${xui_folder}/x-ui setting -h 2>&1 | grep -q -- '-subport'; then setting_args+=(-subport ...)` — so an unknown flag can no longer abort the credential call on a released binary. But the *generated* port is still printed and persisted regardless of that probe:

- `echo -e "${green}Subscription Port: ${config_subport}${plain}"` in the summary block;
- `write_install_result ... "${config_subport}"`, which writes `XUI_SUBSCRIPTION_PORT=<random>` into `/etc/x-ui/install-result.env`.

`-subport` is introduced by this PR, so `releases/latest` does not carry it yet. On that binary the probe skips the flag, the panel keeps the factory `subPort: 2096`, and the installer nevertheless reports a random port and records it in a credentials file that `deploy/cloud-init/cloud-init.yaml:63` sources. Between this merge and the next release, a fresh install therefore tells the operator — and anything that reads that env file — that the subscription server is on a port it is not on.

Suggestion: gate the summary line and the recorded key on the same probe, or read the value back from the binary after `x-ui setting` (`setting -show` now prints `subPort:`, which is what the upgrade branch already parses) and report that instead. The same goes for any wording in the mdx/docs that promises a randomized subscription port on installations that cannot set one.

For the record, the two earlier findings I can see are fixed on this head as well: the recorded key no longer collides with the runtime override (`XUI_SUBSCRIPTION_PORT`), and `updateSetting` in `main.go` now range-checks `subPort` to 1–65535 and rejects a collision with the panel port, so the CLI path can no longer hand `subServer.Start()` a port it cannot bind.

poise52:
@BlindMaster24  Huge thanks for detailed review!

PR #6428, opened 2026-09-06
feat(tgbot): turn the bot into a customer self-service channel

pcxzs:
## Summary

A large Telegram-bot update that turns it from an admin-only console into a
customer-facing self-service channel. A customer links their Telegram account
to a subscription via a `/start` deep link, then fetches their own configs, QR
codes, usage and expiry, requests renewal, and messages the admin — without
ever seeing the panel. Admin capabilities grow alongside it (roster search and
filters, bulk actions with counts, broadcast, on-demand reminders, a Bot
Settings menu), and every surface is gated by a three-level access ladder.

**This is confined to the Telegram bot.** No HTTP routes, no frontend, no
database migration, no changes to Xray config generation, subscriptions or
traffic accounting — the full accounting is in **Scope** below. An operator who
leaves the bot switched off sees no behaviour change whatsoever.

A per-feature before/after table is in **Before and after**.

## Why

The bot previously assumed a single operator: every command answered whoever
typed it, so it could not be handed to paying customers. Running a subscription
service off the panel means customers need a way to self-serve config delivery,
renewal reminders and support contact, while a stranger who finds the bot must
not be able to infer the panel exists at all.

The access ladder (`tgbot_level.go`) resolves each update to **stranger /
client / admin** and allowlists commands per level. Every per-config action
routes through `clientSelfAction` → `ownsClient`, so a linked customer can only
ever act on subscriptions bound to their own Telegram ID.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring (no behavior change)
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Frontend (UI / panel pages)
- [ ] Backend (API endpoints, login, settings)
- [ ] Xray config generation
- [ ] Subscription (share links / Clash / JSON)
- [ ] Statistics / traffic counters
- [ ] Database / migrations
- [ ] Install / upgrade script
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [x] Telegram bot

Telegram bot only. Also adds 230 keys to `internal/web/translation/*.json`
(all 13 locales), updates `docs/content/docs/en/operations/telegram-bot.mdx`,
and registers one bot-only cron job in `internal/web/web.go`. No controller,
no frontend, no migration — see **Scope** below.

## Before and after

Everything below is Telegram-side. The left column is the bot on `main` today.

| Area | Before | After |
| --- | --- | --- |
[...]

PR #6518, opened 2026-09-14
feat(tgbot): access levels and /start account binding

pcxzs:
## Summary

Gives the Telegram bot access levels and a way for a customer to link their own Telegram account to their client, via a `/start` deep link built from the client's subId. This is the first part of #6428, split as suggested there: access levels and account binding first, standing on their own against current `main`.

## Why

Any Telegram account that finds the bot is treated the same as a customer an admin has bound. Binding is also manual: the customer sends `/id`, and the admin copies the ChatID into the client. Both have to change before the bot can face customers, and everything else in #6428 builds on them.

## What changed

**Access levels.** Each update resolves to stranger, client or admin (`tgbot_level.go`):
- Commands are **allowlisted** per level, so a command added later stays admin-only until it is listed.
- A stranger may run `/start` and `/id`, nothing else. `/start` answers with the existing `askToAddUserId` text (their ChatID), so manual binding still works. A stranger's callback taps are answered and dropped.
- A client keeps exactly the commands a non-admin could meaningfully use before: `/start /help /status /id /usage`.
- The level comes from the indexed `tg_id` column of the clients table, so no update, a stranger's included, triggers a scan of the inbound JSON. Ownership of a specific client is still #6489's `clientOwnedByTgUser`, unchanged.

**Private chats only.** The bot ignores updates from groups and channels. Authorization keys on the sender, but wizard state keys on the chat, and those are the same identity only in a private chat. A button tapped in a group is still answered so it stops spinning, and each ignored chat is logged once so an operator can see why the bot is quiet there.

**Binding.** The admin client card gets an **Invite Link** button next to *Set Telegram User*. It produces `https://t.me/<bot>?start=<payload>`, and the first account to open it is bound through the existing `SetClientTelegramUserID`:
- **Payload:** the subId, base64url-encoded, because Telegram accepts only `A-Za-z0-9_-` (up to 64 characters) and a subId may contain `#`, `&` or non-ASCII. A subId too long to fit is refused with a log line rather than truncated.
- **Concurrent claims:** serialized, so of several accounts opening the same link at once exactly one is bound and told so.
- **Several clients:** a subscription spanning several clients binds all of them, and is refused if any part belongs to another account. If one bind fails part-way, the ones already made are undone.
- **Reuse and probing:** re-opening your own link is idempotent. Unknown and already-claimed tokens get the same reply.
- **Guessing:** non-admin claim attempts are limited to 5 per Telegram account per hour. The first refused attempt notifies the admins with the account's name and ID. Plain `/start` without a payload is not counted.
- **Client form:** the *Subscription ID* field gets a tooltip explaining that it doubles as the bot invite code, so a short or guessable one can be claimed by someone else.
- **Unbinding:** stays where it is today, on *Set Telegram User*.

Whether the subId should double as the invite token at all is an open question for review; see the comment below.

## Scope

- **Outside `tgbot/`:**
  - One read-only lookup, `ClientService.GetRecordsBySubID` in `client_lookup.go`.
  - One `tooltip` prop on the Subscription ID field in `ClientFormModal.tsx`.
- **Not touched:** routes, settings, migrations, Xray config.
- **Locales:** 7 new keys, in all 13 files.
- **Size:** 21 files, ~830 lines (378 code, 353 tests, 91 locale, 4 frontend).

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [x] Telegram bot

## How was this tested?

New tests:
- `tgbot_level_test.go`:
  - The allowlist, in both directions.
  - Ignored chat types.
  - `levelOf` against a seeded binding.
  - `gateCommand`: a stranger gets no reply, a client gets "unknown command".
  - `gateCallback`: a stranger's forged tap is answered and dropped.
- `tgbot_invite_test.go`:
  - Token resolution: unclaimed, trimmed, idempotent owner, taken, shared subscription, partly owned, partly held by another, unknown, empty, missing sender.
  - An end-to-end claim that promotes a stranger to client.
  - Concurrent claims: 6 accounts, exactly one told it bound and holding the client. This fails without the claim lock.
  - A payload round trip for subIds containing `#`, `&`, `?` and non-ASCII, plus the 64-character limit.
  - The attempt limit: the 6th attempt is refused, admins are notified once per window, other accounts are unaffected, and the window resets. It fails if the limit is removed.
  - HTML escaping of the account name in the admin notice.

These reuse the existing `staleButtonServer` / `newLinksCallbackTgbot` fixtures.

Checks run:
- Go: `gofmt`, `go vet ./...`, `golangci-lint`, `go test -race -shuffle=on ./internal/web/service/...`, `go test -shuffle=on ./...`.
- Frontend: the i18n dead-keys test, `npm run typecheck`, `npm run lint` (oxlint) and `oxfmt --check`.

## Breaking changes

None for bound customers or admins in private chats. Two visible changes:
- **Unbound accounts** now get only `/start` (showing their ChatID) and `/id`. Previously they could open the customer menu, whose every button answered "configuration not found".
- **Commands and taps sent in a group chat** are ignored, with a one-time log line per chat. Notifications the bot *sends* to a group chat ID are unaffected.

## Checklist

- [ ] I tested the change locally and confirmed the described behavior.


pcxzs:
@MHSanaei A design question before this goes further. It's about the token behind the invite link.

The automated review rightly pointed out that `/start <subId>` makes the subId a binding credential. A subId has no minimum strength (operators set values like `alice`), and reaching the bot needs only its public username. Someone who guesses an unbound client's subId gets bound to it and can pull its links.

I've pushed the light-weight answer, which keeps the subId as the token:
- **Rate limit:** claim attempts are capped at 5 per Telegram account per hour, and the first refused attempt notifies the admins.
- **Tooltip:** the Subscription ID field in the client form says it doubles as the bot invite code, so a short or guessable one can be claimed by someone else.

The alternative is a **dedicated invite code**: a new random field on the client, generated like the subId and shown under it in the client form, independent of the subscription URL. That's safer, since a leaked or guessable subId stops mattering. It also means a model field, a migration with backfill, the client form, the node-sync merge and the API schema. That's exactly the reach beyond the bot you asked the split to avoid, so I didn't want to add it unasked.

Which would you rather see: the subId with a rate limit as it is now, or a separate invite code field?

Issue #6037, opened 2026-07-18
[Bug]: SSRF vulnerable default configuration of outbound direct (freedom)


PA4EJIJIA:
### Describe the bug

Default configuration of outbound direct (freedom) has freedom.finalRules: [{"action":"allow"}], which makes it possible for an authenticated client (vless, vmess, trojan) to bypass routing rule "deny geoip:private". With default domainStrategy: AsIs router does not resolve domain targets,  then freedom.finalRules: [{"action":"allow"}] makes it skip defaultBlockPrivateRule. Said client can expoit this by requesting a domain with private A-record (example 127-0-0-1.nip.io). As a result, the client gains access to loopback services, such as gRPC API Xray (HandlerService/RoutingService) which does not require authentication, and it provides control over xray-core/metrics. Said client essentially gains privilege on xray-core if he exploits this.

### How to reproduce the problem

1. Default fresh panel installation
2. Legit client requests domain with private A-record (example 127-0-0-1.nip.io)
3. Bypass defaultBlockPrivateRule, gains access to xray's gRPC API
4. Said client now can manipulate xray-core settings or read metrics

### Expected behavior

fix: freedom.finalRules: [{"action":"block", "ip":["geoip:private"]},{"action":"allow"}]

### Relevant logs

```shell

```

### 3x-ui version

3.5.0

### Xray-core version

26.7.11

### How did you install 3x-ui?

install.sh script

### Operating system

Ubuntu 24.04

### Which parts of the panel are affected?

Backend (API endpoints, login, settings)

### Is the panel behind a reverse proxy or CDN?

Yes — Nginx

### Before submitting

- [x] I searched [existing issues](../issues?q=is%3Aissue) and this bug has not been reported.
- [x] I am running the latest released version of 3x-ui (or have verified the bug still exists on it).
- [x] This bug report is written in English.
- [x] I have redacted any sensitive data (tokens, real domains, client UUIDs).

MHSanaei:
Fixed in #6083. Blocks private-range egress in the default freedom finalRules (+ a migration for existing installs), closing the SSRF path to loopback services via AsIs domain resolution.

PR #6521, opened 2026-09-14
fix(xray): read an outbound protocol id the way the core does

BlindMaster24:
## Summary

The three seeders that rewrite `xrayTemplateConfig` compared `protocol` case-sensitively (`proto != "freedom"`), while xray-core lowercases that id before it looks up the handler. A template that spells the direct outbound `Freedom` therefore runs as freedom and escapes all three rewrites — and each seeder still records itself as applied, so it never runs again.

## Why

The core's loader, `infra/conf/loader.go`:

```go
func (v *JSONConfigLoader) LoadWithID(raw []byte, id string) (interface{}, error) {
	id = strings.ToLower(id)
```

`infra/conf/xray.go:329` calls it with `c.Protocol` for every outbound, so `"protocol": "Freedom"` is the freedom handler, not an unknown protocol. That id is also the only thing the panel's rewriters look at, which is where the two sides disagree.

User story — a panel whose Xray template was hand-edited (the template is an editable JSON setting in the panel, and it is also what an imported config lands in) with `"Freedom"` for the direct outbound:

1. **xray refuses to start.** `rewriteRemovedOutboundKeys` exists precisely because `sockopt.addressPortStrategy` on freedom is a hard error in the core (`infra/conf/xray.go:340-344`): a panel carrying that key cannot start xray at all. Spelled `Freedom`, the key is never dropped, the seed is recorded as applied, and the install stays dead on a key the seeder was written to remove.
2. **A documented loopback exposure stays open.** `rewriteFreedomFinalRulesPrivateEgress` is the fix for #6037: with `domainStrategy: AsIs` the router never resolves domains, so a domain with a private A record (e.g. `127-0-0-1.nip.io`) sails past the `geoip:private` routing block, and freedom's allow-all `finalRules` let it reach loopback services such as the xray gRPC API and metrics listener. Skipped, the outbound keeps the allow-only rules, the seed is recorded, and no later start ever retries the hardening.
3. **The reverse migration is skipped too.** `rewriteFreedomFinalRules` replaces the legacy private-only-allow `finalRules`, a state an intermediate release shipped, with allow-all; skipped, that outbound keeps its broken egress.

All three are one root cause in one file, so the change is one predicate each.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [x] Xray config generation
- [x] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot

## How was this tested?

`internal/database/rewriter_protocol_case_test.go` adds two tests, both red before the change and green after (verified by reverting the fix and watching each subtest fail):

- `TestRewriteRemovedOutboundKeysSeesAnUppercaseFreedom` drives the vendored core: it asserts `xray.ValidateOutboundConfig` **refuses** the `"Freedom"` outbound carrying `sockopt.addressPortStrategy` (which is the proof the core loaded it as freedom rather than rejecting the id), then requires the rewrite to drop the key and the core to accept the result.
- `TestRewritersTreatProtocolCaseAlike` runs all three rewriters on the same outbound spelled `freedom` and `Freedom` and requires the same verdict and the same migrated keys.

Reproducing locally: `go test ./internal/database/ -run 'RewritersTreatProtocolCaseAlike|SeesAnUppercaseFreedom'`.

Honest note about my machine: it has no C compiler, so `internal/database` builds only under `CGO_ENABLED=0` with the cgo-only `sqlite3.SQLiteConn.Backup` call stubbed by a `-overlay` file (no file of this PR is substituted). The pure-function tests above run that way; every DB-backed test in the package needs cgo and is left to CI, as is `go build ./...`.

## Screenshots / recordings

N/A

## Breaking changes

None. The seeders only match strictly more configs than before, and every config they newly match is one the core already treats as freedom.

## Checklist

- [ ] `go build ./...` and the test suite pass locally. (no C compiler here — CI verifies)


BlindMaster24:
Update after the branch was rebased onto `main`: #6515 merged while this was open, which added a fourth rewriter to `internal/database/db.go` with the same case-sensitive check — `rewriteFreedomDomainStrategy`. It is fixed here too, so all four now read the id the way the core does:

- `rewriteRemovedOutboundKeys` (`proto == "freedom"`)
- `rewriteFreedomDomainStrategy` (`proto != "freedom"`)
- `rewriteFreedomFinalRules` (`proto != "freedom"`)
- `rewriteFreedomFinalRulesPrivateEgress` (`proto != "freedom"`)

The rebase is why the PR now carries two commits; the diff against the current `main` is four lines in `internal/database/db.go` plus the tests.

Each of the four guards is pinned: reverting them leaves five subtests red (`TestRewriteRemovedOutboundKeysSeesAnUppercaseFreedom`, all three of `TestRewritersTreatProtocolCaseAlike`, and `TestRewriteFreedomDomainStrategySatisfiesCore/uppercase_protocol_spelling`), and green with them.

BlindMaster24:
Thanks — finding 1 is right and it is the important one. Addressing each:

**1. MEDIUM — the corrected finalRules predicates never run on the installs this describes.** Agreed, and the corrected predicate alone was not enough: `FreedomFinalRulesReverseFix` and `FreedomFinalRulesPrivateEgressBlock` are in v3.7.0, so their rows exist and `runSeeders` never calls either rewriter again. Fixed in f7e6a741 with a new one-shot seeder, `UppercaseFreedomFinalRulesFix`, which replays both rewrites once for a panel that already carries those rows. It runs only when the template actually holds a freedom outbound whose id is not the canonical lowercase spelling, so a stock lowercase config is left byte-identical and a user who deliberately re-set `finalRules` after the original seeder ran is not touched by a second pass. Registered in the fresh-install pre-seed list and in the run order next to `hardenFreedomFinalRules`.

Tests: `TestUppercaseFreedomFinalRulesFixReachesHistoryGatedPanels` reproduces your scenario against a real SQLite — it seeds the stock allow-only rules, inserts both `FreedomFinalRulesReverseFix` and `FreedomFinalRulesPrivateEgressBlock` rows, runs the real `runSeeders(false)`, and requires the stored rules to be the hardened pair with the protocol id untouched; a second `runSeeders(false)` then has to leave the template alone, which is what pins the gate. (It skips under `CGO_ENABLED=0` locally and runs in CI.) `TestRewriteUppercaseFreedomFinalRules` covers the pure path: the three states the hardener accepts are rewritten, customized rules and a canonical `freedom` outbound are not. Both mutations bite — gutting the rewrite body reddens the three rewriting subtests, and making the spelling probe always answer yes reddens the canonical-spelling subtest.

**2. MEDIUM pre-existing — `rewriteFreedomDomainStrategy` at db.go:1708.** Already fixed on this head; the review was pinned to c3d43873, which was cut before the rebase. The fourth predicate is in 8ec7cbd9, and `TestRewriteFreedomDomainStrategySatisfiesCore/uppercase_protocol_spelling` drives the real core: the migrated outbound must stop warning.

**3. LOW pre-existing — the outbound editor's case-sensitive id.** Valid, and I am leaving it out of this PR on purpose: it is `rawOutboundToFormValues` in the frontend, this diff is backend-only, and the repo asks for focused diffs. I will open it separately.

BlindMaster24:
Rebased onto `main` again: #6519 merged as 39ce7cbc while this was open, and it registers its own seeder on the same line and in the same run order as the one added here, so the conflict was the seeders list. Resolved by keeping both — the pre-seed list now carries `UppercaseFreedomFinalRulesFix` and `DNSOutboundLegacyKeysFix`, both gates run, and the dns rewriter that arrived from `main` keeps its own `!strings.EqualFold(proto, "dns")`. Diff against the current `main` is `internal/database/db.go` plus the two test files; no marker, `gofmt`/`go vet` clean, and the rewriter tests are green.

BlindMaster24:
The `go-test` failure on the previous head was mine, not the change's: `TestUppercaseFreedomFinalRulesFixReachesHistoryGatedPanels` never reached the seeder because `InitDB` pre-seeds a fresh install's `history_of_seeders` rows, including this seeder's own, so the gate saw it as applied and the template stayed unhardened. The test now clears that one row first (the same way `TestMigrateFreedomDomainStrategyRewritesStoredTemplate` does) and asserts the two earlier rows are present instead of inserting them.

I reproduced the failure and the fix locally this time, running the package against a pure-Go SQLite driver so the DB tests execute without a C compiler: with the row left in place it fails with exactly the CI message (`stored finalRules are not the hardened pair`), and with the row cleared it passes. The rest of the package shows the same set of failures as a clean `main` under that driver, so nothing else moved.

BlindMaster24:
Closing the loop on finding 3 of this review, the LOW about the outbound editor reading the id exactly, which I said I would take separately rather than widen a backend-only diff:

- **#6522** — `rawOutboundToFormValues` normalises the id before its `switch`, so a row spelled `Blackhole` no longer falls into the `default:` arm and open as VLESS; without this, saving that row rewrote the block outbound as vless and dropped its settings, so traffic routed to `blocked` stopped being dropped.
- **#6523** (merged as `c0c2dd27`) — the six remaining filter readers: `useOutboundTags` / `useOutboundTagGroups`, the `dialerProxy` picker, `GeodataSection`, `GeneralTab` and the *Test All Outbounds* skip list, all now share `isOutboundProtocol`.
- **#6525** — the two readers that PR's premise left behind: `isUdpOutbound` (probe mode) and the mtproto egress picker's missing `excludeBlackhole`.

So all three findings from this review are delivered; three lines in this PR, the rest in the follow-ups.

PR #6522, opened 2026-09-14
fix(panel): read an outbound protocol id the way the core does

BlindMaster24:
## Summary

xray-core lowercases a protocol id before it looks the handler up, so a template that spells the direct outbound `Freedom` is that outbound — but the panel matched the id exactly. The outbound editor rendered it as an empty vless server, and the Basics tab could not find it and appended a **second** outbound tagged `direct`.

## Why

`infra/conf/loader.go` in the pinned core:

```go
func (v *JSONConfigLoader) LoadWithID(raw []byte, id string) (interface{}, error) {
	id = strings.ToLower(id)
```

`infra/conf/xray.go:329` hands it `c.Protocol` for every outbound, so the spelling never changes which handler runs. Two panel paths disagreed:

1. **The Basics tab could append a duplicate tag and stop xray from starting.** `setDirectFreedomStrategy` and the happy-eyeballs setter both look the direct outbound up with `o?.protocol === 'freedom' && o?.tag === 'direct'` and, on a miss, `push({ protocol: 'freedom', tag: 'direct', settings: {} })`. On a template spelled `Freedom` the lookup always missed, so changing the Freedom Protocol Strategy select (or the happy-eyeballs toggle) added a second `direct` outbound. The core rejects that outright — `app/proxyman/outbound/outbound.go:115`:

   ```go
   if _, found := m.taggedHandler[tag]; found {
       return errors.New("existing tag found: " + tag)
   }
   ```

   So the panel's own settings page could leave the box unable to start xray, and every routing rule pointing at `direct` would then resolve to the wrong handler.

2. **The outbound editor rendered the same outbound as an empty vless server.** `rawOutboundToFormValues` switched on the raw string and fell through `default:` to the vless branch, so the outbounds list showed a vless server with no address; the required-address rule then refused to save it, and saving from the JSON tab would have rewritten the outbound as vless.

User story: a user pastes their config into **Xray Settings → JSON** (or imports one) with `"protocol": "Freedom"` for the direct outbound — the core accepts it and keeps running, so nothing warns. Then they open **Xray → Basics** and change *Freedom Protocol Strategy*, and the panel saves a duplicate `direct`. The next restart fails with `existing tag found: direct`.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [x] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot

## How was this tested?

- `frontend/src/test/basics-freedom-strategy.test.ts` — a template whose direct outbound is `Freedom` with `sockopt.domainStrategy: UseIPv6` must report `UseIPv6` and must still hold **one** outbound after `setDirectFreedomStrategy(t, 'UseIPv4')`, with the value landing on that outbound. Reverting the predicate to the exact match reddens exactly this test (verified by mutating it back).
- `frontend/src/test/outbound-form-adapter.test.ts` — `"Freedom"` parses as freedom with its settings intact and saves back as the canonical id, and `"VLESS"` keeps its flat vless settings. Removing the normalisation reddens this test only (`1 failed | 34 passed`).
- `npm run typecheck`, `npm run lint` and `oxfmt --check` pass on the touched files.

Honest note: the local `unit` suite is `1155 passed, 1 failed`, and the failure is `src/test/input-number-guard.test.ts`, which is unrelated to this change — it also fails with my files stashed, so it is a local-environment difference in the oxlint CLI, not a regression here.

## Screenshots / recordings

N/A — the change is how an id is compared; the visible difference needs a config spelled `Freedom`.

## Breaking changes

None. Nothing is rewritten on load; the editor simply reads the id the way the core does, and saving writes the canonical lowercase id the core would have used anyway.

## Checklist

- [ ] `go build ./...` and the test suite pass locally. (no Go change in this PR)
- [x] For frontend changes: `npm run lint` and `npm run typecheck` pass, and the Vite build succeeds (`npx vite build` — the `npm run build` wrapper also runs `gen:api`, which this PR does not touch).


BlindMaster24:
Both pre-existing findings taken seriously:

**MEDIUM — a non-freedom egress holding the `direct` tag.** Agreed, and it is now guarded in f1e3a99e: both setters check the tag on its own before adding anything, so a `direct` already held by a socks/vless/http outbound keeps it and the edit is dropped instead of making the config unloadable. That matches `setDefaultOutboundTag`, which already looks a tag up by itself and never duplicates one. Test: a template whose only outbound is `{ protocol: 'socks', tag: 'direct' }` must still hold exactly one outbound after `setDirectFreedomStrategy(t, 'UseIPv4')`, unchanged. Removing the guard reddens exactly that test.

Worth knowing for the review: the panel cannot express "the freedom strategy of a `direct` tag that belongs to a socks outbound", so dropping the edit is the deliberate choice over writing a freedom-only key onto a foreign handler.

**LOW — the six remaining exact matches.** Fixed in #6523, which adds one `isOutboundProtocol` helper in `schemas/primitives/outbound-protocol.ts` and converts all six readers to it: `useOutboundTags` / `useOutboundTagGroups`, `OutboundsTab`'s `dialerProxy` picker, `GeodataSection`, `GeneralTab` and the *Test All Outbounds* skip list in `useXraySetting`. The consequence you named is the one that drives the change: a `Blackhole`-spelled block outbound was offered as an mtproto egress, a `dialerProxy` target and the geodata download egress, and each of those silently discards traffic. `use-outbound-tags.test.tsx` drives the real hooks so reverting the call sites turns it red.

PR #6523, opened 2026-09-14
fix(panel): read outbound protocol ids case-insensitively everywhere

BlindMaster24:
## Summary

Six more places read an outbound's protocol id by exact string match, while xray-core lowercases that id before it resolves the handler. An outbound spelled `Blackhole` therefore passed every `excludeBlackhole` filter, and one spelled `Freedom` was queued by *Test All Outbounds*. They all now share one `isOutboundProtocol` helper.

## Why

`infra/conf/loader.go` in the pinned core lowercases the id in `LoadWithID`, and `infra/conf/xray.go:329` hands it `c.Protocol` for every outbound, so the spelling never decides which handler runs. The panel's readers still disagreed with that:

- `api/queries/useOutboundTags.ts` (`useOutboundTags`, `useOutboundTagGroups`) — the tag lists that offer an egress for an **mtproto inbound** and for a `dialerProxy`, both requested with `excludeBlackhole`.
- `pages/xray/outbounds/OutboundsTab.tsx` — the `dialerProxy` picker in the outbound editor.
- `pages/index/GeodataSection.tsx` and `pages/settings/GeneralTab.tsx` — the **geodata download egress** pickers.
- `hooks/useXraySetting.ts` — the skip list for *Test All Outbounds*, which exists so direct/DNS/block outbounds are not probed.

User story: a config pasted into **Xray Settings → JSON** with `"protocol": "Blackhole"` for the block outbound runs exactly like `blackhole` — the core accepts it silently. The block outbound then appears in the mtproto egress dropdown, in the `dialerProxy` list and in the geodata download picker; choosing it looks like a working selection and discards the traffic, with nothing in the UI or the logs to say why. The mirror image on the same template: a direct outbound spelled `Freedom` is queued by *Test All Outbounds*, which probes it as if it were a proxy.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [x] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot

## How was this tested?

- `frontend/src/test/use-outbound-tags.test.tsx` drives the real hook over a mocked panel response whose outbounds are `direct`/`freedom`, `blocked`/`Blackhole`, `warp`/`wireguard`: with `excludeBlackhole` both pickers must return `['direct', 'warp']`, and without it all three tags are offered. Reverting the two call sites to the exact match reddens the first two tests (`expected [ 'direct', 'blocked', 'warp' ] to deeply equal [ 'direct', 'warp' ]`).
- `frontend/src/test/outbound-protocol-case.test.ts` pins the shared rule itself — `freedom`/`Freedom`/`FREEDOM`/`fReEdOm` match, another protocol or its spelling does not, and a missing, empty, non-string, `null` or `undefined` protocol never matches.
- `npm run typecheck`, `npm run lint`, `oxfmt --check` pass on the touched files, and the existing `geodata-section`, `general-tab` and `use-xray-setting` suites still pass.

Honest note on coverage: the two picker components (`GeodataSection`, `GeneralTab`) and `OutboundsTab` are one-line conversions to the same helper and I did not add per-component cases for them, so their exclusion behaviour is pinned by the helper test rather than by a test that renders those pages.

## Screenshots / recordings

N/A — the difference needs a template that spells a protocol id differently.

## Breaking changes

None. Nothing is rewritten on save; the readers simply accept the spellings the core accepts.

## Checklist

- [ ] `go build ./...` and the test suite pass locally. (no Go change in this PR)
- [x] For frontend changes: `npm run lint` and `npm run typecheck` pass, and the Vite build succeeds.


BlindMaster24:
The three pre-existing findings from this review are all closed now — one of them by the sibling PR, two by a follow-up that this branch could no longer carry, since the branch was merged while they were pending.

**#6525** — `https://github.com/MHSanaei/3x-ui/pull/6525` — takes the two that sit on the paths this PR's premise makes reachable:

- MEDIUM, `hooks/useXraySetting.ts:400` (`isUdpOutbound`). Agreed on the consequence you named: a `WireGuard`/`Hysteria` spelling got a dial-only TCP probe, so *Test All Outbounds* reported a working outbound as down. While fixing the protocol id I checked the other half of the same expression against the core and it is the same defect: `TransportProtocol.Build` (`infra/conf/transport_internet.go:16-17`) lowercases the transport name too, so a stream carrying `"network": "KCP"` built `mkcp` and was still probed over TCP. Both now go through the shared normalisation.
- LOW, `api/queries/useOutboundTags.ts:22`. Taken at the call site as you suggested — `pages/inbounds/form/protocols/mtproto.tsx:14` now passes `{ excludeBlackhole: true }`, which is what `HostSockoptForm.tsx:42` and `NodeFormModal.tsx:82` already did.

The test for the LOW renders the real `MtprotoFields` and opens the egress dropdown, so it fails on `expected [ 'direct', 'blocked', 'warp' ] to not include 'blocked'` with the fix reverted. One note on finding the field, in case it is useful: react-hook-form generates the field id and antd does not forward the `htmlFor` FormField passes, so the field cannot be looked up by name or label — the test keys on the fact that the egress picker is the form's only searchable select.

**MEDIUM, `outbound-form-adapter.ts:601` and `basics/helpers.ts:25` / `BasicsTab.tsx:132`** — the write paths, already fixed on #6522 and not duplicated here: `rawOutboundToFormValues` normalises the id before the switch (so a `Blackhole` row no longer opens as VLESS and no longer has its settings dropped on save), and both Basics setters check the `direct` tag on its own before appending, so a foreign holder of that tag keeps it and the edit is dropped instead of producing two outbounds the core refuses to load.

Thanks — the two findings on this PR were both real, and the probe one was reachable from a template this PR's own change already treated as valid.

PR #6525, opened 2026-09-14
fix(panel): probe UDP outbounds and hide the block outbound from the mtproto egress picker

BlindMaster24:
## Summary

Two remaining panel readers that read an xray template differently from the core: the outbound probe picks a TCP mode for UDP-only outbounds whose protocol id or transport name is spelled in a case the panel did not expect, and the mtproto egress picker still offers the block outbound.

Both were raised by the automated review on #6523, filed as pre-existing and non-blocking there. This PR is that follow-up.

## Why

The core normalises both names before it resolves them — `infra/conf/loader.go:46` lowercases the outbound protocol id, and `TransportProtocol.Build` (`infra/conf/transport_internet.go:16-17`) lowercases the transport name. Every panel reader that compares either string exactly disagrees with the core for a template that is spelled differently, and the template is user-editable JSON, so this is reachable from a paste or a hand edit.

**Scenario 1 — a working outbound reported as down.** An operator upgrades and their `direct`/`warp` outbound is spelled `"WireGuard"` (or `"Hysteria"`, or the stream carries `"network": "KCP"`). The core loads it and builds the UDP handler; nothing warns. In *Xray → Outbounds* the operator clicks the row's Test button, or *Test All Outbounds*. `isUdpOutbound` compared `o?.protocol === 'wireguard'` exactly, returned `false`, so the outbound was queued into the TCP lane, where a dial-only probe runs against a UDP-only endpoint. The result is a red "down" row for an outbound that is passing traffic right now, and *Test All Outbounds* shows the same for that entry. The mode is also picked from the same helper by the single-outbound probe, so both the per-row button and the batch run land the same way.

**Scenario 2 — the block outbound selectable as an mtproto egress.** An inbound of protocol mtproto with *Route through Xray* on offers an *Outbound* dropdown. That dropdown called `useOutboundTags()` without options, so nothing was filtered: the tag of the `blackhole` outbound was listed next to the real proxies and could be selected. Choosing it looks like a working selection and silently discards that inbound's Telegram traffic. The two other call sites of this data — the host `dialerProxy` picker and the node form — already pass `excludeBlackhole: true`; the mtproto picker was the one that did not.

**What changed**

- `isUdpOutbound` now matches the protocol id through the shared `isOutboundProtocol` helper and lowercases the transport name before comparing it, and recognises the `mkcp` alias the core resolves to the same mKCP transport.
- The mtproto egress picker asks for `{ excludeBlackhole: true }`, like its two siblings.

**Scope.** The remaining exact `network === …` comparisons in `lib/xray/*` (link builders, inbound tag flags, FinalMask) read values the panel writes itself from a fixed list, so this change does not touch them — they are a different surface and none of them is on the path fixed here. The two write-path readers the same review named (the outbound editor falling through to VLESS for a differently spelled id, and the Basics setters appending a second `direct`) are already fixed on #6522; this PR does not duplicate them.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot

## How was this tested?

`frontend/`, all four commands run locally:

- `npm run lint` — 0 warnings, 0 errors (567 files).
- `npm run typecheck` — clean.
- `npm run build` — succeeds; `gen:api` rewrote `frontend/public/openapi.json` with no diff.
- `npx vitest run --project components src/test/use-xray-setting.test.tsx src/test/mtproto-egress-picker.test.tsx` — 8 passed.

Both tests were confirmed red before their fix, then green with it:

- `use-xray-setting.test.tsx` drives the real hook: it renders `useXraySetting`, calls `testOutbound(0, outbound, 'tcp')` and asserts the `mode` the probe actually posts. With the fix reverted, the `"WireGuard"`, `"HyStErIa"` and `"KCP"` cases fail with `expected 'tcp' to be 'http'`; the canonical `wireguard` and a plain `vless` case pass either way.
- `mtproto-egress-picker.test.tsx` renders the real `MtprotoFields` with a spied config response and opens the egress dropdown. With the fix reverted it fails on `expected [ 'direct', 'blocked', 'warp' ] to not include 'blocked'`.

The picker test addresses the field by an explicit `id` on that `Select`, the same handle the inbound form's protocol select uses (`InboundFormModal.tsx:645`, `listSelectOptions('protocol')`), and reuses `listSelectOptions` from `test-utils` instead of a private copy of the dropdown reader. An earlier revision located the field positionally, which would have stayed green while asserting about a different select the day one above it gained `showSearch`.

## Screenshots / recordings

N/A — no visual change; the picker simply no longer lists the block tag.

## Breaking changes

None. Both changes narrow behaviour to match what the core already does: a probe switches mode for outbounds the core was already loading, and one dropdown loses an entry that could only discard traffic.

## Checklist

- [ ] `go build ./...` and the test suite pass locally. — No Go file is touched; the Go jobs are left to CI.
- [ ] I updated the Wiki / README / API docs if user-facing behavior changed. — No endpoint, schema or documented behaviour changed.


BlindMaster24:
All three taken — two fixed here, the third in its own PR because it is a different surface.

**MEDIUM — the transport-name reader is alias-blind (`mkcp`).** Right, and I should have caught it while I was in that expression: I had the core's switch open in front of me and read only the `kcp` arm. Fixed in a01533c7 — `isUdpOutbound` accepts `mkcp` now, and the trigger is the same one this PR already documented, so it belongs here rather than in a follow-up. New case in the probe-mode table: `{ protocol: 'vless', streamSettings: { network: 'mkcp' } }` must post `mode: 'http'`; with the alias removed it fails on `expected 'tcp' to be 'http'`, and the other six cases stay green.

**MEDIUM — the server-side twin.** Agreed, and it is worse than "the client pre-empts a server guarantee": the second half of your finding is what makes it a real failure rather than a lane choice. I reproduced it before touching anything — `TestOutbounds(batch, srv.URL, "", "tcp")` with `{"tag":"wg","protocol":"WireGuard"}` returns `Success:false Error:"No testable endpoint" Mode:"tcp"`, exactly as you described, because `extractOutboundEndpoints` switches on the raw id and matches no `case`.

That is a separate surface — a Go service and its own tests, on a path the browser no longer drives but the endpoint's documented contract still promises — so it goes in **#6526** rather than widening this frontend diff: `outboundTransportIsUDP` and `extractOutboundEndpoints` in `internal/web/service/outbound/outbound.go` now normalise the way the core does, including the `mkcp` alias, with three tests that are red before the fix (`No testable endpoint` for the `"WireGuard"` batch, five spellings in the gate table, and a `"SOCKS"` outbound against a real listener to pin the extractor on its own).

Two boundaries I left on purpose there, since they are visible in that diff: the literal `blackhole`/`loopback`/`freedom`/`dns` rejections at `probe_http.go:184-188` and the `blackhole`/`freedom` check in `testOutboundTCP` are still exact. The panel can't reach either with a different spelling now — the frontend skips those outbounds before the request leaves the browser — and for a direct API caller the outcome is a less specific error message, not a wrong result.

**LOW — the positional anchor and the duplicated helper.** Both were fair, and the second one is the better half of the point: `form/form-item id` was never the only handle available. `FormField` honours an explicit `id` on its child, so the field now carries one (`mtproto.tsx`), exactly like `InboundFormModal.tsx:645` and its `listSelectOptions('protocol')`, and the test calls `listSelectOptions` from `test-utils` instead of its own reader. On your positional objection: you are right that the old anchor would have gone green while asserting about `preferIp` — that is precisely the failure mode I was trying to avoid and instead wrote in. The rewritten test was re-checked red without the fix (`expected [ 'direct', 'blocked', 'warp' ] to not include 'blocked'`), so it is still pinning the behaviour and not the markup.

On `htmlFor`: you were right to leave it unverified. `FormField` does pass it to `Form.Item` (`FormField.tsx:79`) and antd v6 does not render it on the label element — the label comes out with `htmlFor=null`. The explicit `id` on the child is the handle that survives, which is what the fix uses.

Thanks — the `mkcp` miss and the positional anchor were both mine, in code I had just written.

PR #6526, opened 2026-09-14
fix(outbound): read the probe protocol id and transport name like the core

BlindMaster24:
## Summary

The outbound probe decides between its dial-only TCP lane and the real xray handshake probe by reading the protocol id and the stream transport name. Both readers compared those strings exactly, so a template the core is running was probed as if it were a different protocol.

Follow-up to a review finding on #6525, where the same reader was fixed in the browser.

## Why

The core normalises both names before it resolves them — `infra/conf/loader.go` lowercases the protocol id, and `TransportProtocol.Build` (`infra/conf/transport_internet.go:16-17`) lowercases the transport name and additionally resolves **both** `"kcp"` and `"mkcp"` to mKCP.

**Scenario.** An operator's config spells an outbound `"WireGuard"` — pasted, or hand-edited, or written by another tool — and the core runs it without a word. Someone calls `POST /panel/api/xray/testOutbound` with `mode=tcp` (or `testOutbounds`), as the panel itself did before #6525 and as a script polling the API still does. Then:

1. `outboundTransportIsUDP` compares `"WireGuard"` against `"wireguard"`, gets `false`, and keeps the outbound in the **dial-only TCP lane** — the lane it exists to keep UDP transports out of, because a bare dial neither proves reachability nor measures latency for a protocol that ignores unauthenticated packets.
2. `extractOutboundEndpoints` switches on the raw protocol string, matches no `case`, and returns nothing.
3. The caller gets `No testable endpoint` for an outbound that is passing traffic right now.

The transport-name half has the same shape and an extra edge: `"network": "mkcp"` is a spelling the core accepts and builds mKCP from, and it was not recognised at all. The endpoint's own documented contract (`frontend/src/pages/api-docs/endpoints.ts`) states that `mode=tcp` gives fast dial-only probes with UDP-transport outbounds still probed over HTTP — the first step above is where that promise was broken.

**What changed**

- `outboundTransportIsUDP` matches the proxy protocols and the transport names case-insensitively, and recognises the `mkcp` alias.
- `extractOutboundEndpoints` lowercases the protocol id before its switch.

**Not changed, deliberately — and one correction to what this description first claimed.** The literal comparisons at `probe_http.go:184-188` (the `blackhole` / `loopback` / `freedom` / `dns` rejections) are still exact in this PR, and this diff stays backend-only. An earlier version of this paragraph said the panel never reaches them with a different spelling, because the frontend skips those outbounds before the request leaves the browser. That is wrong, and this PR's own review disproved it. `isUntestable` (`frontend/src/pages/xray/outbounds/outbounds-tab-helpers.ts:57`) is the only gate on the row's Test button and it compares the id exactly as well, so a `"Freedom"` outbound is probed through the direct outbound and the API answers `success: true` with the panel host's own country and delay — the operator reads the panel's own reachability as a working tunnel. That is a separate, pre-existing bug, fixed in #6527 rather than here. The `blackhole`/`freedom` check in `testOutboundTCP` (`outbound.go:160`) stays exact too: it is message precision only, since `extractOutboundEndpoints` finds no endpoint for those ids either way.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Xray config generation
- [x] Statistics / traffic counters (the outbound test/probe path)
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot

## How was this tested?

Three new tests in `internal/web/service/outbound/`, all confirmed red before the fix and green after:

- `TestTestOutboundsTCPModeForcesCoreSpelledUDPToHTTPProbe` drives the real `TestOutbounds(..., "tcp")` through a stub xray process and asserts the result comes back from the HTTP probe. Before: `Success:false Error:"No testable endpoint" Mode:"tcp"`.
- `TestOutboundTransportIsUDPMatchesTheCore` is a table over both readers' spellings — `WireGuard`, `HYSTERIA`, `KCP`, `mkcp`, `MKCP` must be UDP; `tcp`, `ws` and a plain `vless` must not. Five substests fail before the fix.
- `TestTestOutboundsTCPLaneReadsProtocolIDCaseInsensitively` puts a `"SOCKS"` outbound in the TCP lane against a real local listener, so it fails on the extractor alone rather than through the lane gate.

Local verification, and one honest caveat about how. This machine has no C compiler, and `internal/database` is a dependency of the package, so it does not build with `CGO_ENABLED=0` on its own — `internal/database/db.go:2812` calls the cgo-only `sqlite3.SQLiteConn.Backup`. I verified with `go test -overlay` pointing at a copy of `db.go` in which that one call is replaced by an error return; **no file this PR changes is substituted**, only that single cgo call. Under that rig: `go test ./internal/web/service/outbound/` passes in full (the pre-existing suite included), `gofmt -l` is clean and `go vet` is clean. Real-CGo compilation, `race`, the Postgres job and the build matrix are left to CI.

## Screenshots / recordings

N/A — backend only.

## Breaking changes

None. Canonical spellings behave exactly as before; only templates the core already accepts and the probe previously misread change behaviour, and they change from a wrong answer to the right one.

## Checklist

- [ ] I tested the change locally and confirmed the described behavior. — Confirmed for the package under the overlay rig described above; `go build ./...` itself needs CGo on this machine.
- [ ] `go build ./...` and the test suite pass locally. — Blocked by the missing C compiler here; CI covers it.
- [ ] For frontend changes: `npm run lint`, `npm run typecheck`, and `npm run build` pass. — No frontend file is touched.
- [ ] I updated the Wiki / README / API docs if user-facing behavior changed. — No endpoint, request shape or response field changed; only which lane a request lands in.


BlindMaster24:
All four findings disposed of, and one correction to this description that finding 4 disproved.

**1. MEDIUM — the lane gate's browser twin (`outbound.go:230`).** Already fixed, by a later commit. `frontend/src/hooks/useXraySetting.ts:29` on `main` now compares the three proxy ids through `isOutboundProtocol`, lowercases `streamSettings.network`, and lists `kcp` and `mkcp` separately — the same set `outboundTransportIsUDP` matches after this PR. #6525 landed at 14:00 UTC as `84c5aef4`, after the head this review is pinned to (`f7907aad`), so the divergence described did exist there and does not exist at the merge. Both halves are pinned by their own tables: `TestOutboundTransportIsUDPMatchesTheCore` here, and the `mkcp` case in `use-xray-setting.test.tsx`. No lock contention remains for the spellings the finding lists.

**2. MEDIUM — the two readers inside `buildBatchTestConfig` (`outbound.go:225`).** Fixed in `a82476c9`, and this is the finding that was caused by the change, not merely reachable through it: folding `wireguard`/`amneziawg` is exactly what newly routed those spellings onto the readers. `probe_http.go:397` and `:421` now use `strings.EqualFold`. Red-first: the new `TestBuildBatchConfigReadsTheProtocolIDLikeTheCore` drives `buildBatchTestConfig` and fails both assertions without the change — `"WireGuard" settings = <nil>, want noKernelTun` and `"AmneziaWG" protocol = "AmneziaWG", want "socks"` — and passes with it.

**3. LOW — the four-line comment block.** Trimmed to two lines in the same commit.

**4. MEDIUM pre-existing — the testability gate.** Confirmed by reproduction rather than by reading, and this is the one that corrects the description. `isUntestable` (`frontend/src/pages/xray/outbounds/outbounds-tab-helpers.ts:57`) is the only gate on the row's Test button, and it compares the id exactly as well, so the browser half the description relied on does not exist. Driven through the real `TestOutbounds` before any fix:

```
SPELLING freedom  success=false err="Direct/DNS outbound cannot be tested"
SPELLING Freedom  success=true  err=""  egress=&{IPv4:198.51.100.2 Country:ZZ}
SPELLING FREEDOM  success=true  err=""  egress=&{IPv4:198.51.100.2 Country:ZZ}
SPELLING DNS      success=true  err=""  egress=&{IPv4:198.51.100.2 Country:ZZ}
```

A direct or DNS row therefore answers `success: true` with the panel host's own country and delay, not "a probe failure with a less specific message" as the description argued. Delivered separately as **#6527** — the Go gate plus `isUntestable`, red-first on both sides, four files — because it predates this change and this diff is backend-only.

The "Not changed, deliberately" paragraph above is corrected accordingly: it now says what actually holds, which is that the remaining comparisons are exact, that the browser does not cover them, and that the consequence is a false green rather than a vaguer error.

CI on `a82476c9`: every check green, `go-test`, `race`, `golangci`, `codegen`, `frontend`, `postgres-durable-first`, `govulncheck`, `fuzz-smoke`, CodeQL and the build matrix included.

PR #6527, opened 2026-09-14
fix(outbound): read the testability gate's ids like the core

BlindMaster24:
## Summary

The outbound probe has one gate that decides an outbound has no tunnel to test at all — direct, dns, loopback, blackhole. It compared the protocol id exactly, while the core lowercases it before resolving the handler, so a case-variant spelling slipped past the gate and the probe measured the panel host's own reachability. Both halves of that gate are fixed here: the batch gate in Go and the row's Test button in the browser.

## Why

The core normalises the id — `infra/conf/loader.go`, `LoadWithID`, at the pinned revision `52a412d9e2f5`:

```go
id = strings.ToLower(id)
```

So `"Freedom"` runs as freedom, and `"DNS"`, `"Loopback"`, `"Blackhole"` run as their lowercase ids. The panel's gate read the raw string instead.

**Scenario.** An operator's template spells an outbound `"Freedom"` — pasted from a config, written by another tool, or hand-edited in the template editor. They open the Outbounds page and click **Test** on that row (or a script calls `POST /panel/api/xray/testOutbounds`). Then:

1. `isUntestable` (`frontend/src/pages/xray/outbounds/outbounds-tab-helpers.ts:57`) is the **only** thing gating that button — `enqueue` in "Test All" already compares case-insensitively since #6523 — and it compared exactly, so the button stays enabled.
2. The request reaches `testOutboundsParsed`, whose switch (`internal/web/service/outbound/probe_http.go:182-193`) also compared exactly, so the outbound is not rejected and joins the HTTP batch.
3. The temp instance routes the probe **through the direct outbound**, and the API answers `success: true`, a real measured delay, and an `Egress` block carrying the panel's own IP and country.

The operator reads the panel host's own connectivity as a working tunnel. That is the one row where the honest answer is "there is nothing to test here", and a false green is worse than an error — on a DNS row it is the same false green against a resolver that may not exist.

Reproduced before the fix (stub xray process, `EgressTrace` stub), which is what the new test asserts:

```
SPELLING freedom  success=false err="Direct/DNS outbound cannot be tested"
SPELLING Freedom  success=true  err=""  egress=&{IPv4:198.51.100.2 Country:ZZ}
SPELLING FREEDOM  success=true  err=""  egress=&{IPv4:198.51.100.2 Country:ZZ}
SPELLING DNS      success=true  err=""  egress=&{IPv4:198.51.100.2 Country:ZZ}
SPELLING Blackhole success=true err=""  egress=&{IPv4:198.51.100.2 Country:ZZ}
```

Found as a `pre-existing` MEDIUM in the review of #6526 and deliberately kept out of that PR so its diff stayed backend-only; this is that finding, delivered separately.

**What changed**

- `probe_http.go:182` folds the id once before the switch. One line, so every comparison in that switch reads it the way the core does.
- `outbounds-tab-helpers.ts:57` routes `isUntestable` through the shared `isOutboundProtocol` helper, the same one #6523 introduced for the six other readers.

**Not changed, deliberately — and one correction to what this description first claimed.** The TCP lane's readers (`outboundTransportIsUDP` and `extractOutboundEndpoints`, `internal/web/service/outbound/outbound.go`) still compare the protocol id exactly. The first version of this paragraph called that harmless, and the review was right to reject that: it is harmless only for the *untestable* ids. A direct, dns, loopback or blackhole outbound is rejected by the lane gate anyway, so the only difference there is `No testable endpoint` instead of `Outbound has no testable endpoint` (`outbound.go:160`), which is why that line is left alone — no test can separate those two.

It is **not** harmless for a working proxy, and this is the part the description got wrong. A row spelled `"VLESS"` (or `"Trojan"`, `"Shadowsocks"`, `"Socks"`, `"HTTP"`, `"VMess"`) is correctly left testable by `isUntestable`, `isUdpOutbound` passes the mode through as `tcp`, `extractOutboundEndpoints` matches no `case` at `outbound.go:251-288` and the caller is told `No testable endpoint` for an outbound that tunnels fine — a working tunnel read as broken. That is the same defect in the opposite direction, and it is #6526's subject (open), where `TestTestOutboundsTCPLaneReadsProtocolIDCaseInsensitively` goes red on exactly this. It is not this diff, so it is not fixed here.

One more reader in this family is reported and left alone: `outboundAddresses` (`frontend/src/pages/xray/outbounds/outbounds-tab-helpers.ts:28`) still switches on `o.protocol` with exact labels, so a case-variant row renders an empty Address column in all three tables that use it. Display-only and pre-existing; widening this PR for it would be a different change.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Xray config generation
- [x] Statistics / traffic counters (the outbound test/probe path)
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot

## How was this tested?

Two new tests, both confirmed red before the fix and green after — the assertions enumerate the outcome, they do not settle for `err != nil`:

- `internal/web/service/outbound/probe_testability_gate_test.go` — `TestTestOutboundsRejectsCaseVariantUntestableIDs` drives the real `TestOutbounds(..., "http")` through a stub xray process and an `EgressTrace` stub, over six spellings. Before the fix, five of the six failed with `success=true err="" egress=&{IPv4:198.51.100.2 Country:ZZ}`; the sixth (`freedom`) is the control that the canonical path is unchanged. After: all six return the exact rejection string.
- `frontend/src/test/outbounds-testability-gate.test.ts` — twelve cases over the same spellings plus proxies that must stay testable (`vless`, `VMess`, `Trojan`, `Socks`). Before the fix, eight failed (`expected false to be true`); the four proxy cases passed either way and are the guard against over-correcting the gate. After: 12/12.

Local verification, and how. This machine has no C compiler, and `internal/database` is a dependency of the package, so it does not build with `CGO_ENABLED=0` on its own — `internal/database/db.go:2812` calls the cgo-only `sqlite3.SQLiteConn.Backup`. I verified with `go test -overlay` pointing at a copy of `db.go` in which that one call is replaced by an error return; **no file this PR changes is substituted**, only that single cgo call. Under that rig:

- `go build ./...` — clean.
- `go test ./internal/web/service/outbound/` — ok, the pre-existing suite included.
- `gofmt -l` and `go vet` — clean.
- Frontend: `npm run lint` (0 warnings, 0 errors), `npm run format:check`, `npm run typecheck` — all clean; the `unit` vitest project 23/23 including the new file; `npm run build` and `npm run build-storybook` — both pass, and `gen:api` left no generated artefact dirty.

Left to CI, honestly: real-CGo compilation, `race` (the detector needs cgo, so it cannot run here at all), the Postgres job, and the `storybook` vitest project — chromium is not installed in this environment. The `components` project runs 54 of 55 files green; the one failure is `inbound-form-modal.test.tsx › blocks adding TLS without a certificate`, a 5s timeout that reproduces only inside the full-suite run and passes when the file is run alone. That file imports nothing this PR touches (zero references to `isUntestable` or `outbounds-tab-helpers`), and the new test is a `.test.ts`, so it belongs to the `unit` project rather than `components` — a pre-existing load-sensitive timeout, not a finding of this change.

## Screenshots / recordings

N/A — the visible effect is the row's Test button being disabled instead of returning a green result; the assertions in `outbounds-testability-gate.test.ts` cover it, and no UI markup changed.

## Breaking changes

None. Every canonical spelling behaves exactly as before; only templates the core already accepts and the gate previously misread change behaviour, and they change from a wrong answer to the right one. The rejection strings are untouched, and no endpoint, request shape or response field changed.

## Checklist

- [x] I tested the change locally and confirmed the described behavior. — Confirmed for both halves under the rig described above, red before and green after.
- [ ] `go build ./...` and the test suite pass locally. — `go build ./...` passes under the overlay rig; the CGo build itself needs a C compiler this machine does not have, so CI covers it.
- [ ] I updated the Wiki / README / API docs if user-facing behavior changed. — No endpoint, request shape or response field changed; only which outbounds the probe refuses to test.


BlindMaster24:
Both findings taken, and the MEDIUM corrects a claim in the description that was simply wrong.

**MEDIUM pre-existing — the TCP lane's readers.** You are right, and the correction is accepted without qualification. My description said the exact comparisons in the TCP lane were "message precision only", and that holds only for the untestable ids, where the lane gate rejects them anyway. For a working proxy the outcome is the opposite direction of this PR's defect: `"VLESS"` is left testable by `isUntestable`, `isUdpOutbound` passes `tcp` through, `extractOutboundEndpoints` (`outbound.go:251-288`) matches no `case`, and the operator is told `No testable endpoint` for an outbound that is passing traffic. The description now says that, in those words, instead of the claim you rejected.

It is not fixed here, and it does not need to be: that is exactly #6526's subject, which is open. Its `TestTestOutboundsTCPLaneReadsProtocolIDCaseInsensitively` puts a `"SOCKS"` outbound in the TCP lane against a real local listener and went red on `No testable endpoint` before `extractOutboundEndpoints` folded the id, and the same PR folds `outboundTransportIsUDP`; every reader in your chain is in that diff. Separating it was deliberate — this diff is the batch gate and its browser twin, that one is the lane gate and the endpoint extractor.

The one line I still leave byte-identical is `outbound.go:160`, and only because both spellings reject there with a message and no test can separate the two; that is the half your finding does not touch.

**LOW pre-existing — `outboundAddresses`.** Correct, and reported rather than fixed. You anchored it accurately: the row's Address column, the card view and the subscription table all read through that one `switch`, and a `"VMess"` or `"WireGuard"` row with populated `settings` renders blank. It is display-only and it predates this change, so folding it belongs in its own diff with its own test rather than widening the gate fix; the description now carries it as a known gap so it does not get lost.

CI, so the red does not get misread. Everything that reads this diff is green: `go-test`, `frontend`, `golangci`, `codegen`, `postgres-durable-first`, `fuzz-smoke`, `govulncheck`, CodeQL, `race` and the Windows build. The three red `build (arm64|armv7|armv6)` jobs are not this change — they fail inside the same step *after* the Go binary is built and verified static (`Static binary confirmed`), on the geo-asset `fetch` that follows, and the log ends at `Process completed with exit code 4`, which is wget's network-failure code. Checked the four URLs directly while those jobs were red: Loyalsoldier `geoip.dat` 200, Loyalsoldier `geosite.dat` **504 Gateway Timeout**, chocolate4u `geoip.dat` 200, runetfreedom `geosite.dat` 200. The same jobs failed for other pull requests opened in the same window, and this workflow passed in full on my previous branch twenty minutes earlier. A re-run of only those jobs reproduced the same three failures while the asset was still timing out.

PR #6528, opened 2026-09-14
fix(panel): read the outbound protocol id in the Outbounds row like the core

BlindMaster24:
## Summary

The Outbounds row reads the outbound's protocol id the way the core does. Three readers were involved, all display-only:

1. `outboundAddresses` switched on `o.protocol` with exact `case` labels, so a row spelled `"VMess"` fell through to `default` and rendered an empty Address column.
2. The same switch had no `case` for `hysteria` or `amneziawg` — both selectable in the outbound form — so a canonically spelled row was blank too, and its VLESS branch read only the flat `settings.address`/`port`, so a `vnext`-shaped row rendered a bare `:`.
3. The identity cell gated the network/security tags on an exact-match `includes()` over four ids, so the same `"VMess"` row rendered without its `ws`/`tls` tags.

## Why

The core lowercases the id in `LoadWithID` (`infra/conf/loader.go:46` at the pinned revision `52a412d9e2f5`), so a template that spells an outbound `"VMess"` runs normally. The panel's readers do not all agree; these were the ones left in the Outbounds row.

**Scenario.** An operator's template comes from elsewhere — pasted, generated by another tool, hand-edited — and spells an outbound `"VMess"` with populated `settings.vnext`. On the Outbounds page the row is otherwise healthy: the tag is right, the protocol label is right, the Test button works. Only the **Address** column is blank and the stream tags are missing, even though both are in the settings. The same blank cell appears in the card view and in the subscription outbounds table, which read through the same helper. Nothing breaks functionally, but a row that looks like it has no server is exactly what sends someone to re-create an outbound that was already correct.

The hysteria/amneziawg half needs no unusual spelling at all: pick those protocols in the outbound form and the Address cell is blank from the moment the row is created.

Reproduced before the fix, which is what the new tests pin:

```
× reads a capitalised vmess id     expected [] to deeply equal [ 'a.example.com:443' ]
× reads a capitalised trojan id    expected [] to deeply equal [ 'b.example.com:8443' ]
× reads a capitalised wireguard id expected [] to deeply equal [ 'c.example.com:51820' ]
× reads the vnext server of a vless row          expected [] to deeply equal [ 'a.example.com:443' ]
× reads the flat server of a hysteria id         expected [] to deeply equal [ 'e.example.com:443' ]
× reads the peer endpoint of an amneziawg id     expected [] to deeply equal [ 'f.example.com:51820' ]
× renders the stream tags and the address of a "VMess" row
    expected [ 'VMess' ] to deeply equal [ 'VMess', 'ws', 'tls' ]
✓ leaves a canonical id unchanged
✓ still returns nothing for a protocol that carries no address
✓ returns no bare separator for a vless row that carries no server
```

**What changed**

- `outbounds-tab-helpers.ts`: the switch routes through the shared `isOutboundProtocol` (`schemas/primitives/outbound-protocol.ts`), the module that owns the rule, instead of folding the id inline — `isUntestable`, two functions below it in the same file, already reads it that way. The VLESS branch reads `vnext` first and falls back to the flat pair, the order the probe's extractor uses (`internal/web/service/outbound/outbound.go:259-269`), which also stops it emitting a bare `:`. `hysteria` joins that branch and `amneziawg` joins the `peers` branch, the shape each form writes.
- `useOutboundColumns.tsx`: the identity cell asks `isOutboundProtocol` per id instead of `includes(record.protocol as never)`.

**Relationship to #6526 and #6527.** All three are the same defect — a reader of a protocol id disagreeing with the core — split by subject: #6526 is the probe's lane gate and endpoint extractor (Go), #6527 is the probe's untestable gate (Go batch gate + `isUntestable`), and this is the Outbounds row. #6527 touches `isUntestable` in this same file; the hunks do not overlap, so whichever merges second needs no more than a trivial rebase.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot

## How was this tested?

- `frontend/src/test/outbounds-addresses-case.test.ts` — 11 cases: the three case-variant spellings, the `vnext`-shaped VLESS row, the VLESS flat fallback, hysteria, amneziawg, and four controls that are unchanged either way (a canonical `"vmess"` row, a `"freedom"` row that legitimately carries no address, a VLESS row with no server, a capitalised DNS row).
- `frontend/src/test/outbound-identity-tags-case.test.tsx` — renders `OutboundsTab` with a `"VMess"` row and asserts the identity cell's tags are exactly `['VMess', 'ws', 'tls']` and that the row shows `a.example.com:443`.

Each of the seven assertions above was watched red with its fix reverted and green with it restored.

Gates run locally (Node 24): `npm run lint` — 0 warnings, 0 errors; `npm run format:check` — clean; `npm run typecheck` — clean; `npm run build` — passes, and `gen:api` left no generated artefact dirty.

`npx vitest run --project unit` — 73 of 75 files pass. The two failures are not from this change: `input-number-guard.test.ts` (`expected [] to have a length of 3`) fails on the pristine tree too, in isolation, and lints a fixture under `tools/oxlint/` that this PR does not touch; `language-scope.test.ts` passes in isolation (2.7s against the default 5s timeout) and only fails inside the full run. `npx vitest run --project components` — 54 of 56 pass, and the failing files differ from run to run (`inbound-form-modal.test.tsx`, `calendar-expire-setting.test.tsx`, `geodata-section.test.tsx`, `subscription-general-tab.test.tsx` across three runs); each passes in isolation and none imports this PR's modules. The `storybook` project was not run here — chromium is not installed in this environment; CI covers it.

## Screenshots / recordings

N/A — no markup changed; the assertions above cover the rendered value.

## Breaking changes

None. Canonical spellings produce byte-identical output; only rows the core already runs and the readers previously misread change, and they change from a blank cell to the address and tags that are in the settings.

## Checklist

- [ ] `go build ./...` and the test suite pass locally. — No Go file is touched by this PR.
- [ ] I updated the Wiki / README / API docs if user-facing behavior changed. — No endpoint or response shape changed.


BlindMaster24:
All four disposed of, and three of the four are now fixed rather than explained away.

**1. LOW — the inline second copy of the protocol-id fold.** Right, and the precedent argument settles it: `isUdpOutbound` folds the *transport* name inline because no shared reader owns that field, and it asks `isOutboundProtocol` for the protocol id (`useXraySetting.ts:36` vs `:38-40`). `isUntestable`, two functions below it in this same file, already reads the id that way — so the file carried two idioms for one rule. The switch now asks `isOutboundProtocol` per case and the local fold is gone (`4f05774f`); the rule is once again only where you say it belongs.

**2. MEDIUM (pre-existing) — the two red legs.** Re-driven, and the cause is external. The step that dies is the geo-data block, and all six URLs returned 200 when I re-checked them at 15:37Z; Loyalsoldier's `geosite.dat` was answering 504 for the ~40 minutes those runs spanned (`geoip.dat` from the same release was 200 throughout, and PRs by other authors were red on the same workflow in the same window). The failed jobs are re-running here and on #6527.

The diagnosability half of your point is deliberately not folded in: `fetch()` at `release.yml:79` is a workflow change, and this diff is a display fix.

**3. LOW (pre-existing) — the two protocols the panel itself creates.** Taken, and fixed: `hysteria` now shares the flat-address branch and `amneziawg` joins the `peers` branch, which is the shape `protocols/amneziawg.tsx:123-124` writes. Your VLESS note is fixed the way you framed it — the branch reads `vnext` first and falls back to the flat pair, the order the extractor uses at `outbound.go:259-269`. That also removes the bare `:` you anchored on: a VLESS row with neither shape now renders `—`, like any other row without an address.

**4. LOW (pre-existing) — the stream tags on the very row.** Also taken: the identity cell asks `isOutboundProtocol` per id instead of `includes(record.protocol as never)`. You were right that it survives the whole series, and it was the only reader on that row still comparing the id exactly — a row that shows its address while hiding its `ws`/`tls` tags is still half-readable. Pinned end-to-end by rendering `OutboundsTab` with a `"VMess"` row: `['VMess']` before, `['VMess', 'ws', 'tls']` after.

Findings 3 and 4 are pre-existing, so they are your call: I fixed both because they are the same rule in the same row this change already touches, and the hysteria/amneziawg blank cell needs no unusual spelling at all — pick those protocols in the form and the cell is blank from creation. Say the word if you would rather have them as their own diff and I will split them out.

Verified here: each of the seven new assertions watched red with its fix reverted and green with it restored; `npm run lint` 0/0, `format:check` and `typecheck` clean, `npm run build` passes with no generated artefact dirty; unit 73/75 and components 54/56, the failures being `input-number-guard.test.ts` (also red on the pristine tree in isolation, and it lints a fixture this PR does not touch) and a load flake that moves between files run to run and passes in isolation.

PR #6531, opened 2026-09-14
fix(amneziawg): read the outbound pseudo-protocol id like the core

BlindMaster24:
## Summary

`amneziawg.IsAmneziaWGOutbound` compared the outbound protocol id exactly. It is the last reader of that id that does; the id is now read case-insensitively, like the core and like this repository's own probe lane already read it.

## Why

An outbound protocol id is not case-sensitive. The pinned core lowercases it before resolving the handler, and this repository already relies on that everywhere else:

- `internal/web/service/outbound/probe_http.go:399` — the probe lane reads this same id with `strings.EqualFold(p, "amneziawg")`, and `TestBuildBatchTestConfigReadsTheProtocolIDLikeTheCore` (merged in #6526) pins that an `"AmneziaWG"` entry is bridged to `socks` for the temp probe config.
- `internal/database/db.go` reads protocol ids with `strings.EqualFold` in five migrations (`:1634`, `:1720`, `:1835`, `:2153`, `:2256`), including `hasNonLowercaseFreedomOutbound` (`:2330`), which exists precisely because non-lowercase protocol ids were found in real stored templates.

`IsAmneziaWGOutbound` is the exception, and three paths depend on it:

**1. Config generation — Xray fails to start.** `transformAmneziaWGOutbounds` (`xray_amneziawg_outbound.go:25`) skips the entry, so the raw `"AmneziaWG"` pseudo-protocol is handed to the core. Verified by running the core's own validator:

```
"AmneziaWG"  -> ValidateOutboundConfig err = infra/conf: failed to load outbound detour config
                for protocol AmneziaWG > infra/conf: unknown config id: amneziawg
"amneziawg"  -> ValidateOutboundConfig err = ... unknown config id: amneziawg
```

The core does not know `amneziawg` at all — it is a panel pseudo-protocol that only works because the panel swaps it for a socks bridge before writing the config. Skipping the bridge therefore does not degrade the outbound, it makes the whole config unusable.

**2. The amneziawg job — silent no-op.** `internal/web/job/amneziawg_job.go:82` uses the same predicate, so the reconcile loop never builds an `OutboundDesired` for the entry: the outbound carries no tunnel and nothing is logged.

**3. The save path answered two ways for one spelling.** `CheckXrayConfig` (`xray_setting.go:64`) branches on the same predicate: the exact match goes to the panel's own validator, the case variant falls through to the core's. Same probe run, before the fix:

```
"AmneziaWG"  CheckXrayConfig err = xray core rejects outbound "t1": ... unknown config id: amneziawg
"amneziawg"  CheckXrayConfig err = amneziawg outbound "t1": privateKey is required
```

So an operator writing `"AmneziaWG"` was told the core has no such protocol, about a protocol the panel implements and whose lowercase spelling it accepts. After the fix both spellings get the panel's own message (`amneziawg outbound "t1": privateKey is required`), which is the actionable one.

### Reachability, stated honestly

The panel's save path rejects the case variant today — that is exactly why the operator sees the confusing message above rather than a stored template. So paths 1 and 2 need a template that did not come through that save: a restored backup, a direct DB edit, a scripted template, or a legacy DB. The `UppercaseFreedomFinalRulesFix` seeder (and the `hasNonLowercaseFreedomOutbound` predicate it is built on) is the repository's own evidence that such data exists in the wild, and it repairs it in place rather than rejecting it — the same choice this PR makes for the other pseudo-protocol.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [x] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot

## How was this tested?

Red-first: `TestTransformAmneziaWGOutbounds_ReadsTheProtocolIDLikeTheCore` drives the real generation path (`transformAmneziaWGOutbounds`) for three spellings, and the control case (`"amneziawg"`) passes while the two case variants fail before the fix:

```
--- FAIL: .../AmneziaWG
    protocol = "AmneziaWG", want "socks": the bridge never ran, so the raw pseudo-protocol reaches the core
    settings = {Address: Port:0 User:}, want the socks bridge for tag "awg-hop" on port 64900
--- FAIL: .../AMNEZIAWG
    (same two assertions)
--- PASS: .../amneziawg
```

After the change all three are bridged with the tag preserved and the settings rewritten (`address 127.0.0.1`, `port 64900`, `user awg-hop`). Revert `strings.EqualFold` back to `==` and the two failures return.

The save-path claim above is a command result rather than a test: the three lines per spelling come from a throwaway probe that called `IsAmneziaWGOutbound`, `xray.ValidateOutboundConfig` and `CheckXrayConfig` on the same outbound. The probe is not part of this PR.

**Local gate.** `go build ./...` clean, `gofmt -l` clean, `go vet` clean on `internal/amneziawg/`, `internal/web/service/` and `internal/web/job/`; `go test ./internal/amneziawg/` ok and the `TestTransformAmneziaWGOutbounds*` group green (4 tests). This machine has no C compiler, so the CGo build and the DB-backed tests in `internal/web/service` are CI's — the full package cannot run locally (it fails identically on unmodified `main`), which is environmental and not this change.

## Screenshots / recordings

N/A — backend and config generation only, no UI change.

## Breaking changes

None. The canonical spelling behaves exactly as before; the change only widens which entries the panel recognizes as its own amneziawg pseudo-protocol, and the control case pins that.

## Checklist

`go build ./...` passes here; the full `go test ./...` needs a C compiler for the SQLite driver, which CI has and this machine does not — the same caveat as my previous PRs.


BlindMaster24:
Addressed the review finding in `86236deb`.

**LOW — three-line comment block (`internal/web/service/xray_amneziawg_outbound_test.go:34`): fixed.** The block is now two lines and keeps only the why; the test name already carries the what:

```go
// The core folds the protocol id's case before resolving it, so a mixed-case
// spelling must bridge here too or the raw pseudo-protocol reaches the core.
```

**Red-first re-confirmed after the edit.** Restoring the pre-fix `probe.Protocol == "amneziawg"` through a Go `-overlay` (nothing on disk touched) and running only this test:

```
--- FAIL: .../ReadsTheProtocolIDLikeTheCore/AmneziaWG
    protocol = "AmneziaWG", want "socks": the bridge never ran, so the raw pseudo-protocol reaches the core
--- FAIL: .../ReadsTheProtocolIDLikeTheCore/AMNEZIAWG
    protocol = "AMNEZIAWG", want "socks": the bridge never ran, so the raw pseudo-protocol reaches the core
```

The canonical `amneziawg` subtest passes in both runs, so the table separates spelling from behaviour rather than going uniformly red.

**On `golangci`.** The failed job was not a code finding: it died while resolving a transitive dependency, `github.com/molecule-man/go-brrr@v1.1.0` (pulled in by `fasthttp` through `internal/web/service/tgbot`), with `stream error: stream ID 221; INTERNAL_ERROR; received from peer` — a module-proxy hiccup. This change touches `internal/amneziawg` only, and `go vet` plus `gofmt` are clean locally, so I re-ran the job instead of editing code around it.

PR #6539, opened 2026-09-15
fix(amneziawg): wrap the relay port window instead of refusing ids past it

### LINKED ISSUE #6537: [Bug]: Error during creating amneziawg
### Describe the bug

Something went wrong (amneziawg: inbound id 70350 exceeds the relay port window (ids above 435 are not supported))

1. Open Inbound
2. Create amneziawg

### Expected behavior

_No response_

### Actual behavior

_No response_

### Relevant logs

```shell

```

### Screenshots

_No response_

### 3x-ui version

3.8.0

### Xray-core version

_No response_

Docker / Docker Compose

### Operating system

Debian 12

_No response_

_No response_

None

### Before submitting

BlindMaster24:
## Summary

An AmneziaWG inbound's loopback relay port is derived from its row id, and the id is refused once that port would pass 65535 — so a long-lived database runs out of AmneziaWG inbounds it can ever create. The id now wraps into the same port window instead, with a guard for the collision that wrapping allows.

Fixes #6537.

## Why

`SOCKSPortForInbound` returns `SOCKSBasePort + inboundID` (`internal/amneziawgnet/socks_config.go:19`, `SOCKSBasePort = 65100`), and `AddInbound` refused the save when that exceeded 65535 (`internal/web/service/inbound.go:1241`). That is a lifetime budget of 435, not a concurrency limit: `inbounds` is `AUTOINCREMENT`, so an id is handed out once and never reused, and the counter is reset in exactly one place — when the table becomes completely empty (`DelInbound`, `internal/web/service/inbound.go:1445`, SQLite only). The reporter's counter is at 70350, so AmneziaWG has never worked on that database at all, and no amount of deleting inbounds fixes it.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [x] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot

## How was this tested?

- `TestSOCKSPortForInboundKeepsEverySlotInsideTheWindow` — no id derives a port outside `65101..65535`, and every id up to 435 keeps `SOCKSBasePort + id` byte for byte. Fails before the change (`id 436 derives relay port 65536`).
- `TestAddInbound_AmneziawgPastTheRelayPortWindowStillCreates` — the reported case: with the `inbounds` counter pushed to 70001, creating an AmneziaWG inbound succeeds and lands inside the window. Fails before the change with the `exceeds the relay port window` error.
- `TestAddInbound_AmneziawgRefusesAClaimedRelayPort` — a second AmneziaWG inbound whose id is one slot-window away now shares the first one's relay port; the create must be refused with the owner named. Sub-tests cover an enabled and a disabled blocker.
- `TestCheckPortConflict_AmneziawgRelayCollisionBlocksAnAdoptedInbound` — the update path, which is the only place an inbound adopted from a node can have its slot re-checked.

All four fail without the fix.

`go test ./internal/amneziawgnet/` is green here except four egress tests (`TestEgressConnectDomainResolvesThroughTunnel`, `TestEgressConnectDomainIPv6OnlyTunnelResolvesThroughTunnel`, `TestEgressUDPDatagramDomainForwardedIntoTunnel`, `TestEgressUDPDatagramDomainInterleavedClients`), which also fail on the base commit on this machine — they resolve domain targets through the tunnel and need working DNS.

The four DB-backed tests cannot run on this machine at all: it has `CGO_ENABLED=0`, so `go-sqlite3` is its stub (`InitDB: go-sqlite3 requires cgo to work`). They compile, and CI exercises them.

## Screenshots / recordings

N/A — no UI change.

## Breaking changes

None for existing rows: every id up to 435 keeps the exact relay port it had, so no stored config, running relay or generated Xray config moves, and nothing migrates. Two consequences worth knowing:

- The ceiling changes meaning — 435 AmneziaWG relays **at once** instead of 435 inbounds ever created. A create that lands on a live slot is refused with a message naming the inbound that owns it, instead of silently killing Xray's start.
- Rolling the binary back after creating an AmneziaWG inbound with an id above 435 reinstates the old formula, which derives an out-of-range port for that row and makes Xray refuse the config. Nothing can guard that from this side (the old build is frozen), so it belongs in the release notes if it is felt to matter.

## Checklist

- [ ] `go build ./...` and the test suite pass locally — cannot run on this machine (no C toolchain: `CGO_ENABLED=0`); see above, CI covers it.

## Repairs from the first review round

- The fixtures each built the same client email, which an email being panel-unique rejected before either new guard ran; each now derives its email from its own tag (`9ce17d11`).
- A refusal that reported an inbound's automatic relay port read as if that inbound listened on an unrelated port; `portConflictDetail.Relay` makes those messages say `relay port N`, and configured-port messages are unchanged (`3a3f82c1`).
- The sequence-push fixtures now skip under PostgreSQL instead of hard-failing, matching the repository's other dialect-bound tests (`3a3f82c1`).


BlindMaster24:
Round 1 dispositions, on head `3a3f82c1`.

**HIGH — duplicate client email in the new fixtures.** Fixed in `9ce17d11`. Every fixture built the same client email, and an email is unique across the panel (`internal/web/service/inbound.go:1142`), so the second create in a test was refused before either new guard ran. Each fixture now derives its email from its own tag. `go-test` is green on the head after that commit, so the #6537 regression case and the collision guard both execute now.

**MEDIUM — the refusal names an inbound listening on an unrelated port.** Fixed in `3a3f82c1`. `portConflictDetail` carries `Relay`, and both sites that report a derived relay port — `checkAmneziawgnetSocksRelayCollision` and `checkAmneziawgnetSocksReverseConflict` — now render `relay port 65101 (tcp) already used by inbound 'awg-blocker' (#1) on 127.0.0.1`. A site that reports a configured port renders byte for byte as before, and nothing pins the old sentence (`grep` for `already used by inbound` over `*_test.go` returns nothing). Verified locally by rendering the message with and without the branch: `port 65101 …` fails the assertion, `relay port 65101 …` passes.

The retry half I am deliberately not changing. On SQLite the rolled-back insert leaves the counter where it was, so a retry derives the same slot and is refused the same way; on PostgreSQL the sequence advances, so the same request can succeed on a different slot. A deterministic refusal is the better of those two — the same request succeeding or failing by dialect is worse than one that always fails the same way — and the remedy that works is freeing the slot, after which the retry succeeds. Making a retry advance to a free slot needs an id-allocation loop inside `AddInbound`: new behaviour next to a bug fix, so it belongs in its own change if the maintainer wants it.

**LOW — a node-adopted AmneziaWG inbound is impossible.** Wrong; three lines disprove it. `internal/web/service/inbound_node.go:715` creates the central row for a node-reported inbound with `NodeID: &nodeID` and `Protocol: snapIb.Protocol` verbatim, with no eligibility gate — `isNodeEligibleProtocol` is called only from the panel's own write paths (`internal/web/service/inbound.go:1118` and `:1700`). That such rows exist is also why the tree already guards against them twice: `internal/web/service/xray.go:739` skips `NodeID != nil` AmneziaWG inbounds when injecting relays, and `internal/web/service/port_conflict.go:284` filters `node_id IS NULL`. The test's premise stands.

**LOW — SQLite-only fixtures.** Fixed in `3a3f82c1`. `pushInboundIDSequence` now skips when `database.IsPostgres()`, the way this repository's other dialect-bound tests gate themselves. The other two assertions in that file depend on no dialect property and keep running everywhere.

BlindMaster24:
@claude review

Head `3a3f82c1` answers all four findings from the previous round. The HIGH is fixed by `9ce17d11` (each fixture derives its own client email), and `go-test` and `race` are green on this head. The MEDIUM is fixed by `3a3f82c1`, which makes both relay-port messages say `relay port N`; the retry behaviour it also raised is deliberately unchanged, with the reasoning in my reply above. The SQLite-only LOW is fixed by the same commit, and the node-adoption LOW is answered with the three lines that disprove it.

BlindMaster24:
Round 2 dispositions, on head `0b8758d3`.

**MEDIUM — the relay gate ignored `NodeID`.** Correct, and accepted as mine: `9ce17d11` and `3a3f82c1` are my repairs, and this gate was written by `58df4d90`. Fixed by `0b8758d3`, at both call sites, `checkPortConflictTx` (`internal/web/service/port_conflict.go:226`) and the post-`Save` block in `AddInbound` (`internal/web/service/inbound.go:1242`) — each now requires `inbound.NodeID == nil`, the same local-only predicate the forward check three lines up already used.

Red-first, and the red is an executed run rather than a claim: the test landed first as `df6295f9`, a test-only head, and CI's `go-test` failed on exactly one test with exactly the false refusal you described —

```
--- FAIL: TestCheckPortConflict_NodeAssignedAmneziawgOwnsNoRelaySlot (0.19s)
    inbound_amneziawg_relay_window_test.go:170: id 436 is node-assigned and binds no relay,
    so it cannot collide; got "relay port 65101 (tcp) already used by inbound 'awg-blocker' (#1) on 127.0.0.1"
```

One `--- FAIL:` line in that job, nothing else failed. `0b8758d3` turns it green: `go-test` and `golangci` pass on the new head, `race` was still running when I wrote this.

Both premises your finding rests on, re-verified against the tree rather than taken from it: the desired-instance query is `protocol = ? AND enable = ? AND node_id IS NULL` (`internal/web/service/inbound_amneziawg.go:31`) and the injector skips `inbound.Protocol != model.AmneziaWG || !inbound.Enable || inbound.NodeID != nil` (`internal/web/service/xray.go:739`), so a node-assigned row binds no loopback relay and owns no slot.

One thing to record rather than leave implicit: on the `AddInbound` call site the condition is currently unreachable, because `inbound.go:1118` refuses a node-assigned inbound whose protocol is not node-eligible and AmneziaWG is absent from `nodeEligibleProtocols`. I kept it anyway — the two guard functions query `node_id IS NULL` candidates and compare them against the *subject's* derived port, so "the subject is local" is their precondition, and stating it at both call sites is what keeps a future change to that eligibility list from silently reintroducing this. If you would rather see it dropped as dead, say so and I will; I read it as a guard, not as a hypothetical.

**The fixture you flagged — fixed too.** `TestCheckPortConflict_AmneziawgRelayCollisionBlocksAnAdoptedInbound` claimed a node premise and left `NodeID` nil, so it drove a local row and proved nothing about adoption; that was a mislabelled test of mine, and it is now `TestCheckPortConflict_LocalAmneziawgRelayCollisionBlocksTheEdit` with a comment stating what it actually asserts, while the new `TestCheckPortConflict_NodeAssignedAmneziawgOwnsNoRelaySlot` carries `NodeID` and is the red one above.

**Suppressed LOW — the forward message naming a relay port as the inbound's own.** Agreed, pre-existing and unchanged here: `checkAmneziawgnetSocksConflict` returns `Port: inbound.Port` without `Relay: true`, so it renders the checked inbound's configured port rather than the automatic one. It predates `58df4d90` and this pass leaves it, per the rounds rule.

**Nothing else from your round-1 list is open**: `9ce17d11` (fixture emails) and `3a3f82c1` (both messages say `relay port N`; sequence fixtures skip under PostgreSQL) are unchanged on this head, and the declined retry item stays declined for the reason given earlier.

BlindMaster24:
@claude review

Head `0b8758d3` answers the round-2 MEDIUM: both relay-guard call sites now require `NodeID == nil`, and the red is an executed CI run on the test-only head `df6295f9` (one failing test, the exact false refusal reported). Dispositions for every round-2 item are in the comment above.

BlindMaster24:
Round 3 dispositions, on head `0b8758d3`.

**MEDIUM (pre-existing) — the forward check's `enable = true` filter.** Disposition: acknowledged as pre-existing, left unfixed in this PR, and I am not going to quietly fold its repair in here. Your trace checks out against the tree, including the part that decides it: `SetInboundEnable` (`internal/web/service/inbound.go:1566-1588`) flips the column and runs no port check, and `injectAmneziawgnetSocks` (`internal/web/service/xray.go:739`) requires `inbound.Enable`, so a disabled row injects no relay while still holding the slot its id derives. The one thing I verified independently rather than accepting is the "not worsened by wrapping" clause, because that is what puts it out of this PR's scope: the reachable relay range is `65101..65535` for ids `1..435` before the wrap and after it, so no id that could take a slot yesterday can take a different one today. Agreed that the fix, if the maintainer wants it, belongs in `checkAmneziawgnetSocksConflict` in this file — a separate change with its own red test, not a rider on this one.

**Nothing else is open.** The round-2 MEDIUM is answered by `0b8758d3`, and the red-first evidence it carries stands as `df6295f9`: a test-only head whose `go-test` failed on exactly `TestCheckPortConflict_NodeAssignedAmneziawgOwnsNoRelaySlot` with the false refusal you reported, one `--- FAIL:` line in that job. On the unreachable `NodeID == nil` at `internal/web/service/inbound.go:1242` you are right that it is a maintainer call, not a defect; it stays, for the reason in the round-2 disposition.

PR #6540, opened 2026-09-15
fix(amneziawg): stop a disabled inbound's relay slot from being taken

BlindMaster24:
## Summary

A local AmneziaWG inbound owns the loopback relay port its id derives for as long as the row exists — but the forward port check only counted its *enabled* rows as owners. An ordinary inbound could therefore be saved onto that port while the AmneziaWG row was off, and turning that row back on then put two inbounds on `127.0.0.1:65101`, which makes Xray refuse its whole config and takes every other protocol on the host down with it.

Found as a pre-existing MEDIUM during the review of #6539; this is the second half of the same invariant that change wrote down in the other direction.

## Why it happens

`checkAmneziawgnetSocksConflict` filters its candidates with `enable = true` (`internal/web/service/port_conflict.go:290`), so a disabled AmneziaWG row is invisible when a new inbound's configured port is validated. Nothing else covers the gap: the relay is not a database row, so the general `port = ?` query cannot see it, and `SetInboundEnable` (`internal/web/service/inbound.go:1566-1588`) flips the column inside a transaction that runs no port check at all.

Reproduces entirely through the panel:

1. AmneziaWG inbound `#1` exists and is toggled off. `injectAmneziawgnetSocks` skips it (`internal/web/service/xray.go:739` requires `inbound.Enable`), so no relay is in the generated config — but the row still owns slot `65101`.
2. Any inbound is created on port `65101`. The forward check drops the disabled row, the DB query finds nothing, the save succeeds.
3. `#1` is toggled back on. No port check runs.
4. At the next Xray start both inbounds are appended on `127.0.0.1:65101`; Xray fails the second bind and refuses the entire config.

## Why this rule, and not a check on the enable path

Two designs close that sequence, and the review asked for the choice to be deliberate rather than inherited from a shared helper:

- **A (this PR):** a row owns the slot its id derives for as long as the row exists, disabled included, so the port is never handed out.
- **B:** a row owns it only while enabled, and `SetInboundEnable` validates the relay port when it switches the row on.

A is taken because the rule already exists in three of the four AmneziaWG checks in this file, including `checkAmneziawgnetSocksRelayCollision`, which shipped in #6539 and treats a disabled owner as an owner for exactly the reason that decides this: re-enabling runs no port check. B would make those three agree only by relaxing the merged one, and it moves a resource conflict onto a toggle that users flip as a routine on/off switch, where a refusal is far more surprising than at save time. The cost of A is the narrow case in **Breaking changes** below, which B would avoid.

## The change

One predicate: the forward check now counts every local AmneziaWG row, disabled ones included.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [x] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot

## How was this tested?

`TestCheckPortConflict_DisabledAmneziawgStillOwnsItsRelaySlot` — an AmneziaWG inbound is created disabled, and an ordinary inbound is then validated on that row's derived relay port; the save must be refused with the owner named.

It fails without the fix, on a test-only head whose CI run failed on exactly that one test:

```
--- FAIL: TestCheckPortConflict_DisabledAmneziawgStillOwnsItsRelaySlot (0.11s)
    inbound_amneziawg_relay_window_test.go:163: inbound #1 is disabled but still owns relay port 65101; the save must be refused
FAIL	github.com/mhsanaei/3x-ui/v3/internal/web/service	74.514s
```

`TestCheckPortConflict_AmneziawgnetSocksRelayIgnoredWhenDisabled` (`internal/web/service/port_conflict_test.go`) is deleted by this PR. It asserted the rule this change reverses, and it never reached the predicate it named: its fixture seeds `Settings: {}`, which `amneziawg.InstanceFromInbound` rejects on `parsed.Server == nil` (`internal/amneziawg/instance.go:32`) one statement before the `enable` column is looked at, so it passed with or without the filter. Leaving it would have had the package documenting both rules for the same operator state with nothing failing to flag the contradiction.

- [ ] `go build ./...` and the test suite pass locally — cannot run on this machine (no C toolchain: `CGO_ENABLED=0`, so `go-sqlite3` is its stub). The DB-backed tests compile here and CI exercises them.

## Breaking changes

No schema change, no migration, and no row's stored data moves. Xray's generated config is unchanged for every setup that starts today, apart from the case below.

**The one behaviour change:** a database that already holds a *disabled* AmneziaWG row **and** an ordinary inbound configured on that row's derived relay port (`65101`–`65535`) can no longer save that ordinary inbound while it keeps that port — not just a port change, but any save through that inbound's own update path, such as a remark or a TLS setting (adding a client to it goes through a different path that runs no port check). That state exists today and its config starts, because the disabled row injects no relay. The escape is the port change itself, which succeeds: the check compares the incoming port, so moving that inbound off the relay slot clears the refusal, and deleting the AmneziaWG row clears it too. Ports in `65101`–`65535` are panel-internal and never suggested, so reaching this state takes a deliberate choice of a very high port; the state is also the only one in which the AmneziaWG row can never be switched on.


BlindMaster24:
Round 1 dispositions, on head `dbb04d21`.

**HIGH — "the branch is the failing test without the fix". Wrong for the branch as it stands, right about the head you read.** You reviewed `538232ca`, which is a deliberate, transient artefact: on this machine no DB-backed test can run at all (`CGO_ENABLED=0`, so `go-sqlite3` is its stub, and there is no C compiler and no container runtime), so the red half of red-first can only be executed by CI. The method is therefore "push the test alone, read the failure, then push the fix", and the fix landed as `25538897`, one commit behind that head, with `go-test` green on it. The current head is `dbb04d21`. I take the process point — a head that is red by construction is visible to anything reading the PR in that window — and I would rather have that than a red-first claim I never executed; if there is a cheaper way to run this package's DB tests here, I will use it.

**MEDIUM — the added test inverts a rule another test states and that test cannot detect it. Fixed by `dbb04d21`, and your vacuity proof checks out independently.** `TestCheckPortConflict_AmneziawgnetSocksRelayIgnoredWhenDisabled` is deleted. I verified the mechanism rather than taking it: its fixture seeds `Settings: {}`, and `amneziawg.InstanceFromInbound` returns `ok=false` on `parsed.Server == nil` (`internal/amneziawg/instance.go:32`) at `port_conflict.go:297`, one statement before the `enable` column is read — so it is green with the filter, without it, and under either rule. Its doc comment and its name were the only thing stating the old rule, so it could not be left in place while the rule it names stops being true; the rule this PR pins is covered for real by `TestCheckPortConflict_DisabledAmneziawgStillOwnsItsRelaySlot`, whose fixture carries a qualifying server block and an enabled peer and which is the red one above. Its sibling `...RelayIgnoredWhenNoQualifyingPeer` is untouched: that one's premise — no qualifying peer means no relay — is unaffected by this change.

**MEDIUM — the pinned rule reaches the edit path, so "None for working setups" does not hold. Accepted; the description is corrected, and the design is now stated deliberately.** You are right on both counts: `UpdateInbound` runs the same `checkPortConflictTx` (`internal/web/service/inbound.go:1721`), and the owner row is never excluded from the candidates, so every edit of the ordinary inbound is refused while it keeps that port. The PR body now says so instead of claiming otherwise, and a new **"Why this rule, and not a check on the enable path"** section records the decision you asked for: the alternative is to keep `enable = true` and validate the relay port in `SetInboundEnable`, which is the more forgiving design, and it was rejected because the disabled-owner rule is already what the other three AmneziaWG checks in this file do — including `checkAmneziawgnetSocksRelayCollision`, merged in #6539 — for the same reason that decides this one, that re-enabling runs no port check. Taking the forgiving design would make those three consistent only by relaxing the merged one.

One correction to that finding's severity, offered as evidence rather than as a softening: the state is not frozen. The check compares the *incoming* port, so changing that inbound's port off the relay slot is itself an accepted save, which is the escape you also identify; deleting the AmneziaWG row clears it too. So the operator's way out is the action that resolves the underlying conflict, and the state is reachable only by having deliberately configured an inbound inside `65101`–`65535`.

BlindMaster24:
@claude review

Head `dbb04d21` answers all three round-1 findings: the HIGH was the transient test-only head (fix `25538897`, `go-test` green on it), the contradicting vacuous test is deleted in `dbb04d21`, and the edit-path consequence is corrected in the description with the design choice stated. Dispositions are in the comment above.

BlindMaster24:
Round 2 dispositions, on head `dbb04d21`.

**MEDIUM (pre-existing) — the peer gate releases the slot.** Acknowledged as pre-existing, left unfixed in this PR, and reported rather than folded in. The clause that puts it out of scope is the one you stated and I re-derived: ownership ended at *two* gates before this change — `enable` and `InstanceFromInbound` — and this PR removes only the first, so a row with zero qualifying peers released its slot identically before and after. I also confirmed both halves of the trigger rather than accepting them: `normalizeAmneziaWGSettings` (`internal/web/service/inbound_amneziawg.go:206-217`) leaves a fresh inbound with no peers, and `ClientService.AddInboundClient` (`internal/web/service/client_inbound_apply.go:315`) validates forwarded ports only — the relay checks run at `inbound.go:1229` and `:1721` and nowhere else. Worth recording that this hole is reached by the ordinary create-then-add-a-client workflow rather than by a deliberate disable, which makes it the more likely of the two to be hit; it wants its own change and its own red test, and the decision about where it belongs (the peer gate, or the client-add path) is the maintainer's.

**The body overstatement is corrected.** You were right that adding a client is not refused: the description now says any save through that inbound's own update path, and names the client path as the one that runs no port check.

**Round-1 items re-checked on this head, all as you re-verified them**: the transient test-only head is gone (`port_conflict.go:292` queries without `enable`; `go-test` `success`, package really ran, 66.013s), the contradicting vacuous test is deleted, and the edit-path escape is confirmed — the check compares the *incoming* `inbound.Port`, so moving that inbound off the slot is an accepted save.

No new production code is owed from this round.

PR #6542, opened 2026-09-15
fix(amneziawg): reserve the relay port before an AmneziaWG inbound has a peer

BlindMaster24:
## Summary

A local AmneziaWG inbound reserved its automatic loopback relay port only once it had a qualifying peer, so a freshly created AmneziaWG inbound owned nothing. An ordinary inbound could be saved onto the port that row derives, and adding the inbound's first client then put two inbounds on `127.0.0.1:65101`, which makes Xray refuse its whole config and takes every other protocol on the host down with it.

Found as a pre-existing MEDIUM during the review of #6540. That change removes the *other* gate on ownership, `enable`; this one removes the peer gate, and the two are independent.

> **Red-first head.** On this machine no DB-backed test can run at all — no C toolchain (`go-sqlite3` needs cgo), no container runtime — so the failing half of red-first can only be executed by CI. The first commit here is deliberately the test alone; `go-test` is red on it, and the fix is the commit after. This pull request is opened as a draft for exactly that window, so the review job does not read a head that is red by construction.

## Why it happens

`checkAmneziawgnetSocksConflict` skips a candidate when `amneziawg.InstanceFromInbound` returns `ok=false` (`internal/web/service/port_conflict.go:300`), and `InstanceFromInbound` requires at least one enabled peer with a public key and allowed IPs (`internal/amneziawg/instance.go:38-52`). A fresh AmneziaWG inbound has none: `normalizeAmneziaWGSettings` writes `Clients: []model.Client{}` for empty settings (`internal/web/service/inbound_amneziawg.go:206-217`), and `injectAmneziawgnetSocks` skips a row with no peer emails (`internal/web/service/xray.go:756-760`).

Nothing else covers the gap, and nothing re-checks it later: only two call sites run `checkPortConflictTx` at all — `AddInbound` (`internal/web/service/inbound.go:1229`) and `UpdateInbound` (`:1721`) — while the client paths that can create the first peer, `ClientService.AddInboundClient` (`internal/web/service/client_inbound_apply.go:315`) among them, run no port check.

Reproduces entirely through the panel, with the ordinary workflow rather than a deliberate toggle:

1. Create an AmneziaWG inbound and give it no client yet. The row owns no slot, and `InstanceFromInbound` is `false`, so no relay is generated.
2. Save any inbound on that row's derived port (`65101`–`65535`). The forward check drops the AmneziaWG row, the DB query cannot see a relay that is not a row, and the save succeeds.
3. Add the inbound's first client. No port check runs.
4. At the next Xray start the relay is generated next to the inbound from step 2; Xray fails the second bind and refuses the entire config.

## Where the guard belongs

Two shapes close this, and this PR takes the first:

- **Ownership follows the row (this PR)**, so a local AmneziaWG row reserves its derived port from creation. One predicate removed, and the check then states the same rule as its two siblings — `checkAmneziawgnetSocksRelayCollision` and `checkAmneziawgnetSocksReverseConflict` key on `protocol` and `node_id IS NULL` alone (`internal/web/service/port_conflict.go:320-327`, `:347-354`).
- **Ownership follows what is injected**, which needs a port check added to every path that can create the first peer or re-enable the row — at least `AddInboundClient`, `UpdateInboundClient`, both client-enable setters and `SetInboundEnable`. That is five new validation sites, each able to refuse a client operation because of a loopback port the operator cannot see or act on from where the refusal appears.

The first is also the only shape consistent with what already ships: #6539's collision check has no peer gate.

## The change

The peer gate is removed from the forward check's candidate loop, and the now-unused `amneziawg` import goes with it.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [x] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [ ] Telegram bot

## How was this tested?

`TestCheckPortConflict_AmneziawgnetSocksRelayReservedBeforeTheFirstPeer` (`internal/web/service/port_conflict_test.go`) replaces `TestCheckPortConflict_AmneziawgnetSocksRelayIgnoredWhenNoQualifyingPeer`, which asserted the rule this change removes. The rewrite keeps the test's position and shape and changes three things: its name, its doc comment, and the direction of its assertion. Its fixture is also corrected — the old one seeded `Settings: `{}``, a shape the service never writes, and the new one seeds what `normalizeAmneziaWGSettings` writes for a fresh inbound: a server block with `"clients":[]`.

It fails without the fix, on the test-only head:

```
--- FAIL: TestCheckPortConflict_AmneziawgnetSocksRelayReservedBeforeTheFirstPeer (0.11s)
    port_conflict_test.go:898: an AmneziaWG inbound with no peer yet still owns relay port 65101; the save must be refused
```

Executed in [run 34944524141, job `go-test`](https://github.com/MHSanaei/3x-ui/actions/runs/34944524141/job/104300664280) — the only `--- FAIL:` line in that job, and `golangci` passed alongside it.

- [ ] `go build ./...` and the test suite pass locally — cannot run on this machine (no C toolchain: `CGO_ENABLED=0`, so `go-sqlite3` is its stub). The DB-backed tests compile here and CI exercises them.

## Breaking changes

No schema change, no migration, and no row's stored data moves. Xray's generated config is unchanged for every setup that starts today.

**The one behaviour change:** saving an inbound onto the derived relay port of a local AmneziaWG row that has no qualifying peer is now refused, where it used to be accepted. That is the bug above, so the refusal is the fix; it also means a row whose settings cannot produce a relay at all reserves its port until the row is deleted or its settings are made valid — the edit that makes them valid runs the port check, so a squatter is caught there rather than silently surviving.

**Residual, not fixed here:** a database that already reached the broken state before this change — a peerless AmneziaWG row plus an ordinary inbound on its derived port — still breaks Xray when that row's first client is added, because the client paths run no port check. Closing that needs a decision (refuse the client operation, or re-check the relay on the peer transitions), so it is reported rather than folded in. Ports in `65101`–`65535` are panel-internal and never suggested, so reaching this state takes a deliberately chosen high port.


BlindMaster24:
Round 1 dispositions, on head `17f069a0`.

**LOW — the doc block's justification clause now contradicts the function it cites. Correct, and fixed in `17f069a0`.** You are right that this is mine: this PR rewrote the sentence's subject and left the clause, so the block claimed every matching row gets a relay inbound while the rows this change newly reserves for are exactly the ones `injectAmneziawgnetSocks` skips. A reader following the cross-reference lands on the guard this branch removes and reads it as the rule, which is how a removed guard comes back.

The fix replaces the twelve-line block rather than patching the clause. What is left is the two things that are true and load-bearing — what the check compares, that the reservation is broader than what is emitted, and why `db` is a parameter — which also puts the block inside `CLAUDE.md`'s two-line cap instead of five times over it. The peerless reason stays where it is actionable, in the two-line comment above the candidate loop. I am flagging the size of that edit deliberately: it is a pre-existing over-long block, and I shortened it because the finding landed on it, not as tidying. The other over-long blocks in this file are untouched.

**MEDIUM (pre-existing) — the reservation is only consulted from `AddInbound`/`UpdateInbound`.** Acknowledged as pre-existing, disclosed in the description's Residual paragraph, and not fixed here. Your two transitions are the right pair: `AddInboundClient` (`client_inbound_apply.go:315`) creates the first peer, and `SetInboundEnable` (`inbound.go:1566-1588`) flips the column, and neither runs a port check. Closing it means deciding what a *client* operation should do about a loopback port the operator cannot see from there, which is a design call, not a predicate — so it is a line in the report. Your second consequence is already handled by the sibling PR: `#6540` removes the `enable = ?` filter, and this PR does not, which is why the two are separate.

**On the failure text being unverified in your read.** Fair, and the description now carries the job link, so it is checkable from the PR: [run 34944524141](https://github.com/MHSanaei/3x-ui/actions/runs/34944524141/job/104300664280), one `--- FAIL:` line, `port_conflict_test.go:898`.

**One disclosure about the test, since you reviewed it after the red run.** The assertion you saw that names the owning inbound was added in the fix commit, not in the test-only head. The red property is unchanged by it: it can only fire when the conflict is non-nil, and on the test-only head the check returned nil, so the earlier `Fatalf` was the one that ran — which is what the CI log shows. The line number moved from the head you would have seen to `:898` for the same reason.

BlindMaster24:
@claude review

Head `17f069a0` answers the round-1 LOW: the doc block no longer claims every matching row gets a relay, and it is now two lines. The pre-existing MEDIUM is acknowledged and disclosed as a residual rather than fixed. Dispositions are in the comment above.

BlindMaster24:
@claude review

Rebased onto `2d8d3048` after #6540 merged, so this head is `b41118cb`. One conflict, in this function's doc block, where #6540 edited a word inside the twelve-line block this PR replaces — resolved by keeping this PR's two-line block and #6540's query change, which is what the two changes mean together. `enable = ?` is gone from the candidate query and the peer gate is gone from the candidate loop; both inline comments survive. Diff against the new base is the same two files, +19/-23.

PR #6544, opened 2026-09-15
fix(amneziawg): refuse a row's own relay port and keep a disabled row's slot reserved

BlindMaster24:
Closes two relay-slot gaps in the same guard family, both of which the #6542 review had left `pre-existing`. Both had the same shape: every guard that reserves an AmneziaWG row's own ports skipped a row it should have counted — the row being saved.

## 1. A row could claim its own relay port

All three relay checks exclude the id whose slot they compute (`id != ignoreId`), so nothing compared an AmneziaWG row's own WireGuard listen port with the relay port its own id derives. Saving a row on that exact port puts two UDP listeners on the same port:

- the embedded device binds UDP on `inbound.Listen:inbound.Port` (`internal/amneziawgnet/device.go:137`, `ListenPort: ib.Port` at `internal/amneziawg/instance.go:65`);
- its injected relay binds TCP **and** UDP on `127.0.0.1:port` (`internal/web/service/xray.go:770`, `udp:true` at `internal/amneziawgnet/relay.go:47-61`).

Whichever loses the bind race dies — and if it is the relay, Xray refuses the whole config and every other protocol on the host goes down with it. It took nothing exotic to reach: the first AmneziaWG inbound on port `65101` is id 1, whose derived relay port is exactly `65101`.

`amneziawgnetSocksSelfConflict` now states the rule the three siblings already state — a local row owns the slot its id derives — from both call sites that know the id: `AddInbound`'s post-Save block and `checkPortConflictTx`'s `ignoreId > 0` block. A node-hosted row is untouched: it binds no relay on this host, so its WireGuard port may legitimately be the port its id would derive.

## 2. A new row's own clients' forwards were validated before it existed

`normalizeAmneziaWGSettings` validates every client's `ForwardedPorts` before the row is saved, so the new row was never a candidate for itself in that pass — a client could forward exactly the relay port the row's own id derives, or its own WireGuard listen port, and the create was accepted. At runtime the panel's wildcard forward listener and Xray's loopback relay then race for the same port. `AddInbound`'s post-`Save` block, the only frame where the id exists, now re-runs the guard over the settings submitted in that same request; both callers share one copy of the collision message via `amneziaWGForwardedPortsConflict`.

This one reads only what the request submitted, so it can never refuse a previously-stored spec.

## Reverted after round 2: reserving a disabled row's relay slot

An earlier commit here dropped `enable = true` from `loadPortConflictContext` so that a disabled AmneziaWG row's relay slot could not be claimed by a client's `ForwardedPorts`. Round 2 showed what that costs, and the code confirms it: `UpdateInboundClient` carries a stored spec forward for a partial edit (`client_inbound_apply.go:763-765`) and re-validates it (`:772`, `:909`), so after an in-place upgrade an operation that never submitted the field — a bot enable/expiry toggle — is refused over a slot the operator did not touch, for a relay that is not even emitted while the row is disabled. The inbound-save path re-validates every stored spec the same way (`inbound_amneziawg.go:281`).

The slot that protects is claimable only by a spec an operator authors onto `65101-65535`; the cost lands on unrelated operations. Nothing forces that trade here, so the widening is undone and the slot returns to the residual list below. The precise fix — refuse a *newly claimed* spec rather than a stored one, and check the enable transition in `SetInboundEnable`, which is where the conflict is actually created — is larger than the hole and belongs in its own change.

## Red-first

Every fix here was watched red before it was written, on the same tree, by running the real paths against a throwaway SQLite database:

```
--- FAIL: TestAddInbound_AmneziawgRefusesItsOwnRelayPort (0.21s)
    inbound_amneziawg_relay_window_test.go:190: WireGuard port 65102 is inbound #2's own relay port; the create must be refused
--- FAIL: TestUpdateInbound_AmneziawgRefusesItsOwnRelayPort (0.20s)
    inbound_amneziawg_relay_window_test.go:212: WireGuard port 65101 is inbound #1's own relay port; the save must be refused
--- FAIL: TestAddInbound_AmneziawgRefusesAClientForwardingItsOwnRelayPort (0.30s)
    inbound_amneziawg_relay_window_test.go:196: inbound #2 derives relay port 65102 and its own client forwards that port; the create must be refused
```

The two self-collision tests pin two genuinely different call sites: `AddInbound` reaches the checks after `Save` (the id only exists then), `UpdateInbound` through `checkPortConflictTx` with `ignoreId > 0`. A fix in either place alone leaves the other red.

`TestCheckPortConflict_NodeAssignedAmneziawgOwnsNoRelaySlot` gains the same rule for a node-hosted row's own port. That assertion cannot be red on the unfixed tree (nothing refused it before), so it is a regression guard against the new refusal firing on rows that own no slot, not a fix verification. The test that pinned the reverted disabled-row rule was deleted with it.

## Gate

- `go test ./internal/web/service/ -count=1 -shuffle=on` — **green** on every head (whole package, DB-backed tests included). It runs here through an overlay that swaps the cgo SQLite driver for a pure-Go one, since this host has no C compiler; no test or production file is substituted.
- `gofmt` clean, `go vet` clean on `internal/web/service`, `internal/amneziawgnet`, `internal/amneziawg`.
- CI on the final head: 21 check runs green, `race` included. `Analyze (actions)` failed once with "job was not started because it repeatedly failed to be acquired" and is green on re-run — infra, not a code signal.
- Not covered by any local run: PostgreSQL (the new tests' fixtures are SQLite-only, like their siblings in this file) and the Xray gRPC end-to-end path (`XRAY_E2E_BINARY` unset).

## Residual, left alone

- **A disabled row's own listen port stays claimable**, and `SetInboundEnable` (`inbound.go:1569-1590`) flips the column with no port check. Closing it means refusing the *enable* — the operation that actually collides — rather than widening the forward guard, which would stop specs that work today on a port that is genuinely free. Needs the maintainer's call.
- **The forward guard compares a client's spec with the inbound's stored listen port** (`inbound_amneziawg.go:281`), so moving an AmneziaWG inbound onto a port one of its own clients already forwards is accepted on the update path; the post-`Save` re-run in this PR closes it on create only. The update-path half needs the in-memory row overlaid onto the guard's context, which is its own change.
- **Client operations still run no port check**, so a database already in a broken state is not repaired — it is refused on the next save of the offending row, which is the point.


BlindMaster24:
Thanks — read both inline findings against the tree, and took the first one here.

**Round 1, `internal/web/service/inbound.go:1243` (create path, own clients' forwards) — fixed in `80eb5712`.**

You are right about the mechanism, and the fix is exactly where you put it. `normalizeAmneziaWGSettings` validates the clients' `ForwardedPorts` at `inbound_amneziawg.go:275` against a context read before the row exists, so the new row was never a candidate for itself. `AddInbound`'s post-`Save` block is the only place with an id, so it now re-runs the guard there:

```go
// The clients' forward specs were validated while this row had no id,
// so the ports it now derives were never in the guard's context.
if aErr := s.checkAmneziaWGForwardedPorts(tx, inbound.Settings); aErr != nil {
    return aErr
}
```

Two notes on the shape:

- The re-run closes **both** ports the row owns, not just the relay slot: with the row now in `ctx.inbounds`, a client's spec matching the row's own WireGuard listen port is refused too — the same wildcard-listener-versus-device-bind race, and previously accepted for the same reason. That is a new refusal on the create path, and I treat it as correct: a spec sitting on its own inbound's listen port was never going to work, it silently lost the bind.
- The collision message now lives in one place. `amneziaWGForwardedPortsConflict` is shared by the pre-`Save` loop (which delegates to it instead of formatting inline) and the post-`Save` re-run, rather than a second copy of the string.

Second-order effect you flagged — the row becoming un-editable until the forward is removed — is gone for new rows, and for a row already in that state the next save still refuses it, which is the point: no lockout, the forward has to move.

Red-first, and watched red twice before this commit existed:

```
--- FAIL: TestAddInbound_AmneziawgRefusesAClientForwardingItsOwnRelayPort (0.30s)
    inbound_amneziawg_relay_window_test.go:196: inbound #2 derives relay port 65102 and its own client forwards that port; the create must be refused
```

Gate: whole `internal/web/service` green, `-shuffle=on`, 139s (DB-backed tests included, run against a throwaway SQLite database through the same pure-Go driver overlay described in the description). `gofmt` and `go vet` clean.

**Round 1, `internal/web/service/inbound_amneziawg.go:351` (a disabled row's own listen port) — open, and I am not fixing it here.**

The argument transfers, and I am not going to pretend it does not: I reserved the relay slot on exactly this reasoning and left the row's own port claimable. Two differences make it a decision rather than the missing half of this PR.

- A disabled row's own port **is** free while it is disabled, and `TestCheckForwardedPortsConflict_IgnoresDisabledInboundPort` pins that as intended behaviour for every protocol, not just AmneziaWG. Widening the guard means every disabled inbound's port stops being forwardable — a refusal on a spec that works today, for inbounds that will never come back.
- The precise fix is not in this guard at all. The conflict is created by the *enable*, not by the forward: `SetInboundEnable` (`inbound.go:1569-1590`) flips the column with no port check. Refusing the enable while a forward listener holds the port rejects only the operation that actually collides, and leaves the forward alone. That is a new refusal on a different endpoint with its own blast radius, so it belongs in its own change and in the maintainer's call, not appended to a guard fix.

I have flagged it as the next step in the series rather than silently widening this one.

BlindMaster24:
@claude review

BlindMaster24:
**Round 2, `internal/web/service/inbound_amneziawg.go:331` (introduced) — accepted, fixed in `0f90bc6b` by undoing the widening.**

You are right, and the tree confirms it: `client_inbound_apply.go:763-765` carries a stored spec forward and `:772` / `:909` re-validate it, so an operation that never submitted the field is refused over a slot the operator did not touch. The inbound-save path re-validates every stored spec the same way, so the mine sits on two paths, not one.

Weighed properly, the trade does not pay: the slot this reserves is claimable only by a spec an operator authors onto `65101-65535`, while the cost lands on unrelated operations — and on a relay that `injectAmneziawgnetSocks` does not emit while the row is disabled at all. So the commit that widened the context is undone here, and `TestCheckForwardedPortsConflict_DisabledAmneziawgRelayPortIsReserved` is deleted with it rather than left asserting a rule that no longer holds.

The precise fix is what you named elsewhere in the review, and it is bigger than the hole: refuse a **newly claimed** spec rather than a stored one, and check the enable transition in `SetInboundEnable` (`inbound.go:1569-1590`), where the conflict is actually created. That is its own change, not a widening of a guard six call sites share.

`80eb5712` is untouched by the revert and does not carry the same risk: its re-run reads the settings submitted in the same request, so a stored value can never reach it. `TestAddInbound_AmneziawgRefusesAClientForwardingItsOwnRelayPort` still passes.

**Round 2, `internal/web/service/inbound_amneziawg.go:281` (pre-existing) — open, and deliberately not in this PR.**

Confirmed by reading the line: `portCtx` is built from database rows at `:275`, so on an update the guard sees the row's stored port, never the one the same request is saving. Your create-path half is closed here; the update-path half needs the in-memory row overlaid onto the context (replacing its stored entry, and adding it when a disabled row is being enabled) — a context-shaped change of its own, which is why it is not bolted on here.

With `0f90bc6b` this PR has no introduced findings outstanding: two fixes, both red-first, both limited to what the request itself submits.

BlindMaster24:
@claude review

PR #6548, opened 2026-09-15
perf(nodes): reuse one pooled client per node instead of rebuilding it

BlindMaster24:
## What this fixes

A 100-node panel polls each enabled node continuously: heartbeat and traffic sync are `@every 5s` compile-time constants (`web.go:303-304`), so one node sees ~6 requests per 5 s tick type, ~153 HTTPS requests/second across 100 nodes. Two things made that far more expensive than it needs to be.

**A client per tick.** `probe()` (`service/node.go:1274`) called `runtime.HTTPClientForNode` directly. For `skip`, `pin` and `mtls` modes that function built a **new** `http.Client` with a **new** `http.Transport` on every call — empty pool, so every heartbeat paid a full TCP + TLS handshake, and the CPU in a TLS handshake is server-certificate signature verification, which is exactly what shows up in monitoring.

**A pool too small for the fleet.** `MaxIdleConns: 64` with one node per host: 100 hosts cannot fit in 64 global idle slots, so Go closed the excess and those connections re-handshook even where the client was already cached.

What was *already* fine, so this does not overclaim: the sync path went through `Manager.RemoteFor`, which caches `Remote` per node, so `Remote.httpClient()`'s `sync.Once` already pooled there; and `verify` mode plus plain `http` got the shared `defaultNodeHTTPClient`. The cost was the heartbeat probe (all non-verify nodes), the ad-hoc `runtime.NewRemote(n, nil)` calls in `service/node.go:728`, and the global idle cap on every path.

## The change

* `HTTPClientForNode` now caches the client per **node identity** — `id, mode, scheme, address, port, pinned cert` — with the proxy URL as a variant of it, because the heartbeat reaches a node directly while the sync path reaches it through its egress. A hit returns the pooled client.
* When a node's identity changes, its previous client is closed (`CloseIdleConnections`) and dropped: no request can be served by a client built for a trust decision that no longer applies. `InvalidateMasterClientConnections` still works unchanged — the credential-rotating transport observes the epoch on its next `RoundTrip`, which is what its doc comment already promised.
* Idle caps raised to `512` global / `8` per host, in one `newNodeTransport` helper that replaces the same struct literal in three places. Arithmetic: one idle connection per node host, plus the sync and push requests, so 512 covers a 200-node fleet with room for concurrency; the memory is a few KB per idle connection.
* Client construction is delegated to a private `buildNodeHTTPClient`; the now-unreachable `verify` fallthrough at the end of the old function is gone, since the wrapper owns that decision.

Rejected: caching keyed by node id alone, which would serve a `skip` client to a node the operator has since switched to `pin`, and would hand a direct client to a proxied request.

`TestHTTPClientForNodeReusesOneConnectionAcrossCalls` runs two calls against an `httptest` TLS server counting `http.StateNew` connections. Probe — build a client per call (the pre-change behaviour):

```
--- FAIL: TestHTTPClientForNodeReusesOneConnectionAcrossCalls (0.02s)
    tls_client_test.go:130: TLS handshakes = 2, want 1: a rebuilt client re-handshakes on every tick
```

Green after the change, and `TestHTTPClientForNodeRebuildsWhenNodeIdentityChanges` pins the cache key the security argument rests on: a node switched to `pin` (or given a proxy) must never be served the client built to skip verification.

The pre-existing tests in the same file still pass unchanged — `HTTPClientForNodeVerifyShared`, `HTTPClientForNodeMTLSRebuildsTLSConfigAfterCredentialInvalidation`, `HTTPClientForNodeProxyMTLSRebuildKeepsProxyAndNewCredential`, `HTTPClientForNodePinInvalid`, `HTTPClientForNode_ProxyPinPreservesPinEnforcement`, `HTTPClientForNode_ProxyVerifyNoPin`. That is the check that the cache did not weaken credential rotation or pin enforcement.

## Gate

* `go test ./internal/web/runtime` — green (local run through a `-overlay` that supplies a pure-Go SQLite driver; the tree's `go.mod`/`go.sum` are unchanged).
* `gofmt -l` and `go vet` on the package — clean.

## What I did not do, and why

* **Proxy paths keep `http.DefaultTransport`'s caps** (100 global / 2 per host), since `netproxy` owns that transport and only egress-tagged nodes use it. Sequential per-node ticks need one connection, so the per-host cap is not the constraint there; raising it would change every other `netproxy` caller.
* **`GetAllInboundClientIps()` is still called inside the per-node loop** (`node_traffic_sync_job.go:450`) — one full table read per node per tick. Same family, different mechanism (DB reads and pushed bytes rather than handshakes) and a separate diff; it belongs in its own PR.
* **The largest traffic win is a protocol change and needs your call:** every tick pulls the node's whole `panel/api/inbounds/list` with settings, streamSettings, sniffing and client stats, although it rarely changes. A conditional request on the existing route (node answers `304` when the body would be identical, master keeps the previous snapshot) would cut the bulk of the 1.8 KB average response. I have not written it — it changes master↔node wire behaviour and version compatibility.


BlindMaster24:
Round-1 findings, both answered in 31bc1362 and 7bc0588b.

**MEDIUM `tls_client.go:221` — the eviction loop retained every proxy variant of the current identity.** Fixed by replacing the key-prefix filter with one entry per node (`dropNodeClients`): on insert, every other entry of that node is closed and dropped, and the `verify`/`http` early return now clears the node as well — which closes the third leak you named, a node switched to `verify` keeping its old `skip`/`pin` entry forever. The cache is now bounded by the fleet size rather than by the number of operator "test node" clicks.

`TestHTTPClientForNodeKeepsOneClientPerNode` fails without this — watched red first, `cached clients for the node after "socks5://127.0.0.1:41002" = 2, want 1` — and pins the verify-mode cleanup on the same cache.

One thing I deliberately did not do: evict entries for nodes that no longer exist. This function has no view of the node table, and giving it one puts a DB read on the heartbeat path. A deleted node now costs exactly one unreachable client, the same bound as the fleet; if that should go too, it needs an explicit deletion hook instead.

**LOW — three 3-line comment blocks.** Trimmed to two lines each.

Gate on this head: `go test -overlay <pure-Go SQLite rig> ./internal/web/runtime -shuffle=on -count=1` → `ok 0.324s`; `gofmt` clean; `go.mod`/`go.sum` unmodified (the rig exists only because this container has no C compiler). The 100-node handshake reduction remains unverified here — no fleet to measure.

PR #6549, opened 2026-09-15
fix(inbounds): check ports when an inbound is enabled, not only when it is saved

BlindMaster24:
## What breaks today

`SetInboundEnable` (`inbound.go:1574`) flips the `enable` column and, for a node row, polls the remote — with **no port check on that path at all**. Every guard in the panel lives on the save path: `checkPortConflictTx` runs from `AddInbound` (`inbound.go:1229`) and `UpdateInbound` (`inbound.go:1729`), and nowhere else. So a row can reach the enabled state having never been checked, whenever the pair was not produced through a guarded save:

* a row that predates the guards — they landed over three PRs this cycle (`#6539`, `#6540`, `#6542`), so an installation upgrading across them can hold a same-port pair in its DB;
* a restored backup or a hand-edited row, which never went through `AddInbound`.

Once the flag moves, the generated config has two inbounds on one socket, and the core does not skip the second one — it exits the whole process on the failed listen (`main/run.go:94`), taking every protocol on the host down, then the one-second watchdog retries it. Reached from the UI with the enable switch; no unusual steps.

## The change

Check the same rules before the flag moves, and only for a local enable — nodes run their own Xray, so a node row sharing a local port stays legal, and enabling is the direction that can create a collision (disabling cannot):

```go
if enable && inbound.NodeID == nil {
    conflict, err := checkPortConflictTx(db, inbound, inbound.Id)
    ...
}
```

The refusal is `common.NewError(conflict.String())`, the same shape `AddInbound`/`UpdateInbound` already return, so the panel shows *which* row owns the port (`'holder' (#2) port 44431 (tcp)`). Because the check runs before the transaction, a refused enable leaves the row disabled — no half-applied switch.

Reuse, not a new rule: this is the existing checker with all of its special cases (tcp/udp coexistence, reserved API and egress ports, AmneziaWG relay slots, node exemption), so an enable can no longer be refused for something a save would have allowed.

Note this is the same funnel question the previous PRs closed from the config side: with #6547 a colliding config is refused before the core is swapped, so the failure is loud either way — but refusing at the switch is where the operator can act, and it keeps the log from being the only report.

`TestSetInboundEnableRefusesAPortAnotherEnabledInboundServes` — two rows on `0.0.0.0:44431`, the second disabled through the DB, then the enable switch:

```
--- FAIL: TestSetInboundEnableRefusesAPortAnotherEnabledInboundServes (0.27s)
    inbound_enable_port_test.go:39: enabling a row onto a port another enabled inbound serves must be refused
```

Green after the change, and it asserts the two things that matter beyond the error: the message names `holder`, and the row is still disabled afterwards.

The state is seeded with `seedInboundConflict` (a direct `db.Create`) plus a raw `Update("enable", false)` on purpose: `AddInbound` treats a **disabled** row as owning its port (`port_conflict.go:245-252` has no `enable` predicate), so the API refuses to build this pair in the first place — which is exactly the class of row this guard is for.

Two guards against the fix being too broad, both green before and after:

* `TestSetInboundEnableAllowsTCPUDPCoexistence` — a `vless/tcp` row enabling onto a port a `hysteria/udp` row holds, the legal coexistence the rest of the guards allow.
* `TestSetInboundEnableAllowsANodeRowOnALocalPort` — a node row enabling onto a local row's port.

The pre-existing `TestSetInboundEnableRoutedMtprotoRequestsRestart` still passes, so the MTProto restart path through this function is untouched.

## Gate

* `go test ./internal/web/service -shuffle=on` — green, 255 s (local run through a `-overlay` supplying a pure-Go SQLite driver; the tree's `go.mod`/`go.sum` are unchanged). All `TestAddInbound*`, `TestUpdateInbound*` and `TestSetInboundEnable*` tests included.
* `go test ./internal/web/job` and `./internal/web/controller` — green.
* `gofmt -l` and `go vet` on the touched packages — clean.


BlindMaster24:
Round-1 findings.

**LOW `inbound.go:1585` — the recorded reason was wrong.** You are right and I have corrected it in bd324355, both in the code and in the PR description. `checkPortConflictTx`'s candidate query (`port_conflict.go:245-252`) has no `enable` predicate, so a disabled row owns its port and `AddInbound` refuses the pair I described as step 2 — which is why the test seeds it with a direct `db.Create`. The comment now states the reason that survives that correction: every guard in the panel lives on the save path (`AddInbound`, `UpdateInbound`), and `SetInboundEnable` had none, so a row reaches the enabled state having never been checked whenever the pair was not produced by a guarded save — a row predating the guards (`#6539`, `#6540`, `#6542` all landed this cycle, so an upgrade across them can carry such a pair in its DB), a restored backup, or a hand-edited row. The PR body now says that instead of the repro that cannot happen.

**MEDIUM pre-existing `inbound.go:1587` — an inbound's port against existing forwarded ports.** Corroborated and left as pre-existing, per the rules here: this change did not introduce it, and fixing it adds a second direction to `checkForwardedPortsConflict`'s reach — a query for forwarded ports across every AmneziaWG row on the same host, wired into both `checkPortConflictTx` and the enable path, plus the port-forward side. That is a guard-shaped change of its own, not something to attach to a comment fix. Recorded for a follow-up PR; the trigger you describe (AWG client forwards 8443, then a VLESS row is created or enabled on 8443) is reachable on the current tree and I have not softened it.

Gate on this head: `go test -overlay <pure-Go SQLite rig> ./internal/web/service -run TestSetInboundEnable -count=1 -v` → all four green, including the pre-existing MTProto restart case; `gofmt` clean; `go.mod`/`go.sum` unmodified.

PR #6550, opened 2026-09-15
fix(xray): restart when a diff strands a client's live session

### LINKED ISSUE #6533: [Bug]: Disabled client keeps active connection
### Describe the bug

When a client is disabled in the 3x-ui panel, an already established connection may remain active and continue passing traffic.

Connect using a client.
Disable this client in the 3x-ui panel.
Check the existing connection.

### Expected behavior

The active connection should stop after the client is disabled.

### Actual behavior

The client is shown as disabled in the panel, but the existing connection continues to work.

### Relevant logs

```shell

```

### Screenshots

_No response_

### 3x-ui version

3.8.0

### Xray-core version

26.9.9

install.sh script

### Operating system

Debian 13

_No response_

_No response_

None

### Before submitting

BlindMaster24:
Closes #6533.

## Is it a bug? Yes — and the cause is in the core, not in the panel's intent

Disabling a client does take it out of the config: `buildInboundForLocalRuntime` (`inbound.go:2037-2060`) drops any client whose DB or JSON `enable` is false, so a disabled client cannot open *new* connections. What it cannot do is end the connection it already has, and reading the core explains why:

| core symbol | what removal actually does |
|---|---|
| `proxy/vless/inbound/inbound.go:245` | `RemoveReverse` + `validator.Del(e)` |
| `proxy/vmess/inbound/inbound.go:177` | `usersByEmail.Remove(email)` + `clients.Remove(email)` |
| `proxy/trojan/server.go:124` | `validator.Del(e)` |
| `proxy/shadowsocks/server.go:62`, `proxy/hysteria/server.go:66` | `validator.Del(...)` |

Every one of them touches the credential store only. An established session was validated once at handshake and lives in its own goroutine afterwards, so it never consults that store again. The core has no other lever: `app/proxyman/command/command.proto` defines exactly `AddUserOperation` and `RemoveUserOperation`, and there is no `KickUser`, `DisconnectUser` or `CloseUserConnections` anywhere in the module. **There is no API that closes one client's session.**

The panel already behaves as if it knew this: when a client is disabled *automatically* by quota or expiry, the traffic jobs call `RestartXray(true)` under the `restartXrayOnClientDisable` setting (default **true**, `setting.go:164`), which is the only thing that actually ends the session. The manual path did not — it took the hot route (`diffInboundUsers`, `internal/xray/hot_diff.go:180` → `AlterInbound`), which reported success while the session kept flowing. That asymmetry is the bug.

(The same illusion sits in `check_client_ip_job.go:615`, whose comment says removing a user "disconnect[s] all connections". For vless/vmess/trojan/shadowsocks that cannot hold, per the table above — that one is a separate finding and I have not touched it.)

## The fix

A diff tells the two cases apart: a disable or a delete *removes* a user, an edit re-adds the same email under the same tag with new values. `HotDiff.DropsUsers()` reports the first, and `tryHotApply` steps aside for the restart the operator's own setting asks for:

```go
if s.restartToDropClients(diff) {
    logger.Info("hot apply: clients left the config, restarting to drop their live sessions")
    return false
}
```

Why gate it on `restartXrayOnClientDisable` rather than always restarting: with the setting off an operator has explicitly said "do not restart Xray when clients get disabled", and that choice has to survive its own auto-disable path, which would otherwise reach the same state and restart anyway on the next config change.

Scope of the behaviour change, stated plainly: with the setting on (the default), disabling or deleting a client now restarts Xray, so **every** client's session blips for a moment — the same trade the auto-disable path already makes. With it off, nothing changes from today.

## What the setting means now

| event | setting on (default) | setting off |
|---|---|---|
| client disabled by hand (UI, bot, bulk) | diff drops the user, hot path steps aside, Xray restarts, **session ends** | hot `RemoveUser`: new connections rejected, open session keeps flowing |
| client deleted | same as above | same as above |
| client edited (limit, UUID, remark) | email is re-added, hot apply, no restart | no restart |
| client added, inbound or routing edited | hot apply, no restart | no restart |
| client auto-disabled by quota or expiry | restart, as today | no restart, as today |

The cost of the setting being on is not new: a restart blips **every** session on the panel, not just the disabled client's, which is what the auto-disable path already pays today.

## i18n

`restartXrayOnClientDisableDesc` now says "auto-disabled by expiration or traffic limit, or disabled or deleted by hand" in all 13 locales, and the key-parity test (`src/test/i18n-dead-keys.test.ts`) passes. The label itself still reads "Restart Xray After Auto Disable" -- renaming it is a wording decision for you, not a requirement of the behaviour. If you would rather keep the setting strictly about the automatic path, say so and I will drop both the wording change and the gating on it.

Both layers, with the predicate disabled:

```
--- FAIL: TestHotDiffDropsUsers
    --- FAIL: TestHotDiffDropsUsers/client_taken_out_of_the_config
    --- FAIL: TestHotDiffDropsUsers/renamed
    --- FAIL: TestHotDiffDropsUsers/one_dropped,_one_edited
--- FAIL: TestRestartToDropClients
    --- FAIL: TestRestartToDropClients/disabled_client,_setting_untouched
    --- FAIL: TestRestartToDropClients/disabled_client,_setting_on
```

Exactly the cases that encode the rule fail; `edited in place`, `added`, `unchanged` and `setting off` stay green — which is what shows the predicate splits "stops being served" from "changed while staying served" rather than flagging any user change.

## Gate

* `go test ./internal/xray` — green.
* `go test ./internal/web/service -shuffle=on` — green, 173 s, DB-backed tests included (local run through a `-overlay` supplying a pure-Go SQLite driver; the tree's `go.mod`/`go.sum` are unchanged).
* `go test ./internal/web/job ./internal/web/controller` — green.
* `gofmt -l` clean.

What is not covered by a test, honestly: the three lines inside `tryHotApply` that call the predicate. Reaching them needs a running core and a gRPC API, so they are verified by reading; the decision they make is what the tests above pin.


BlindMaster24:
Round-1 findings, all answered in 9deea036 and efabd391.

**HIGH `xray.go:1428` — the guard was in a path the manual flow never enters.** Confirmed exactly as you wrote it, and that is a design error on my part, not a placement nit: `needRestart` ends up false after a successful `rt.RemoveUser`, so none of the three schedulers fires and `tryHotApply` is never entered. The repair went to the `needRestart` bookkeeping you pointed at: when the client actually leaves the config — the update path's remove-without-re-add, the delete path, and bulk disable — the apply layer now asks for the restart `restartXrayOnClientDisable` promises, and only for local inbounds so a node row cannot make the master restart its own core.

The guard in `tryHotApply` stays as the funnel for everything else, but it is no longer the only thing between a disabled client and its session.

Red first: `TestManualClientDisableHonoursRestartSetting` drives the UI switch (`SetClientEnableByEmail`) with a DB-backed local inbound, the local runtime override and a real `SettingService`; with the fix neutralised it fails `needRestart = false, want true` on the setting-on case, and the setting-off case pins that it does not restart when the operator turned it off.

On the second consequence you named — the stale `process.SetConfig` snapshot turning the *next* unrelated change into a restart — I traced it rather than assumed: with the setting on, the restart now happens immediately after the disable, so the snapshot is refreshed then; with it off, the stale snapshot makes the next diff report the user as removed, `DropsUsers()` is true, and `restartToDropClients` reads the setting and lets the hot apply proceed. So the blast radius is the one restart the setting asks for, not a latent restart on an unrelated edit later.

**MEDIUM `internal/xray/hot_diff.go:35` — shadowsocks and hysteria could not reach the predicate.** Confirmed on the code path: `RemovedUsers` is only produced for the protocols `diffInboundUsers` will diff, so those two fell through to the remove+add branch with the guard bypassed. Fixed without touching that gate: the diff now also compares `settings.clients` of an inbound present in both configs and records what disappeared as `DroppedClients`, which `DropsUsers()` reads. `settings.clients` with an email per entry is the one shape every account list here shares, so the signal is protocol-agnostic and needs no new per-protocol rule; the RPC plan is unchanged, which is why I did not widen `userDiffableProtocols` — that would have changed how the hot apply drives those protocols, not just the restart decision.

Red first: `TestHotDiffDropsUsersOnProtocolsItCannotDiff` fails for both protocols without the change (`DropsUsers = false ... removed=[] added=[] dropped=[]`), and its edit case proves an edited client is still served.

**LOW — two 3-line comment blocks.** Both trimmed to two lines.

**LOW — the description outran the title and the docs.** The setting's title no longer says "Auto" in all 13 locales (`restartXrayOnClientDisable` now reads as what it does), and the callouts in `docs/content/docs/{en,ru,zh,fa}/config/clients.mdx` describe the manual case too. The docs change is text only — no endpoint or generated output is involved, so nothing needs regenerating there.

Gate on this head: `go test ./internal/xray -shuffle=on` → ok; `go test -overlay <pure-Go SQLite rig> ./internal/web/service ./internal/web/job ./internal/web/controller -shuffle=on` → ok 165s / 9.4s / 8.0s; all 13 locale files parse; `gofmt` clean; `go.mod`/`go.sum` unmodified.

PR #6551, opened 2026-09-15
docs(limit-ip): correct what the temporary disconnect can actually do

BlindMaster24:
## What the comment claimed vs what the core does

`disconnectClientTemporarily` said it "removes and re-adds a client to **force disconnect** banned connections", and the call site repeated it ("Remove user to disconnect all connections"). Reading the core at the version this repo pins, removal cannot end a live session for any protocol the function's gate admits:

| protocol (gate at `check_client_ip_job.go:644`) | what `RemoveUser` reaches |
|---|---|
| vless | `proxy/vless/inbound/inbound.go:245` — `RemoveReverse` + `validator.Del(e)` |
| vmess | `proxy/vmess/inbound/inbound.go:177` — `usersByEmail.Remove(email)` + `clients.Remove(email)` |
| trojan | `proxy/trojan/server.go:124` — `validator.Del(e)` |
| shadowsocks | `proxy/shadowsocks/server.go:62` — `validator.Del(e)` |
| hysteria | `proxy/hysteria/server.go:66` → `account/config.go:64-76` — deletes the auth and the uuid from two maps |

Every one of them touches a credential store. A session was validated once at handshake and lives in its own goroutine afterwards, so it never consults that store again. The core exposes exactly two user operations — `AddUserOperation` and `RemoveUserOperation` (`app/proxyman/command/command.proto`) — and no `KickUser`, `DisconnectUser` or `CloseUserConnections` exists anywhere in the module. There is no API that closes one client's session.

What actually enforces the IP limit is one layer down: the job logs `[LIMIT_IP] Email = ... || Disconnecting OLD IP = ...` (`:561`) for the fail2ban `failregex` documented right above it (`:557`, installed by `x-ui.sh` / `DockerEntrypoint.sh`), so the banned IP is dropped in the kernel. The remove-and-re-add is a nudge at best: it makes new handshakes fail for those 100 ms.

This is the same root as #6533, which I fixed separately in #6550: the panel can only end a session by replacing the process that owns it.

## What this PR changes

Comments only — three of them:

* the function doc now says it drops the credential for a moment so new handshakes are refused, and that fail2ban is what ends live traffic;
* the call site says `RemoveUser` only clears the validator, naming the five protocols;
* "Wait a moment for disconnection to take effect" became "Keep the credential out long enough for the core to apply the removal".

Behaviour, the protocol gate and `TestDisconnectClientTemporarilyAllowsHysteria` are untouched — that test pins the *gate*, and its own comment ("the skip left over-limit Hysteria2 sessions alive until the fail2ban ban caught up") is compatible with what I describe here: the ban is what ends the session.

## If you want a real immediate disconnect

Three options, my recommendation first:

1. **Leave it to fail2ban and delete the RPC + sleep.** The kernel drop is already the enforcement; then `disconnectClientTemporarily` and its test can go, saving a gRPC round trip and 100 ms per banned client per scan. Needs your say-so because it removes a feature-shaped call.
2. **Restart Xray per violation** — this is what #6550 does for a manual disable, but a cron that can fire for several clients per tick makes a full process restart expensive and disruptive, so I did not do it here.
3. **Ask upstream for a kick API.** Nothing in the panel can substitute for it.

## Verification

* `go test ./internal/web/job` — green (local run through a `-overlay` supplying a pure-Go SQLite driver; the tree's `go.mod`/`go.sum` are unchanged), including `TestDisconnectClientTemporarilyAllowsHysteria`.
* `gofmt -l` clean; the change is comment-only, so no runtime behaviour is claimed and no red-first test applies.

Not verified, stated plainly: this is a reading of the pinned `xray-core` sources, not a reproduction. I have no Xray binary in this environment (`XRAY_E2E_BINARY` unset), so I did not observe a live session surviving a removal — I observed that nothing in the code path could end one. If you want that settled empirically before deciding between the options above, say so and I will build the test.


BlindMaster24:
Round-1 findings, all three answered in e9b31c1a.

**LOW `check_client_ip_job.go:668` — "only clears its validator" is wrong for vless.** Corrected: the comment now says the validator is cleared and a session already up keeps running, **except vless**, where the removal also drops a reverse handler (`proxy/vless/inbound/inbound.go:245-247`). I read the same lines you did; the sentence no longer presents the RPC as a harmless credential flip.

**LOW `check_client_ip_job.go:675` — the sleep was documented as waiting for the core.** Corrected: `AlterInbound` applies `RemoveUserOperation` inline, so nothing is pending when the sleep starts, and the comment now says the 100 ms only widens the window in which new handshakes fail — which is what this PR's description claims and what the code does. Whether the sleep should exist at all is the maintainer's call and this change does not make it.

**HIGH pre-existing `check_client_ip_job.go:678` — a vless reverse client loses its tunnel.** Recorded, not fixed here — this PR touches three comment blocks and nothing else, and the repair is a behaviour change in a different place than the sentences it corrects. The shape I would use, if you agree: gate the disconnect cycle on the client carrying a reverse tag (the data is already in `clientConfig["reverse"]` / `model.Client.Reverse`) and skip the remove/re-add for it, since `buildUserAccount` (`internal/xray/api.go:544-579`) drops `Reverse` and the re-added account therefore fails `GetReverse` until a restart. `internal/xray/hot_diff.go:124-127` is the existing precedent for refusing to hot-swap a reverse client, so the gate belongs on the same idea, not in `buildUserAccount` alone. Kept as a separate change rather than folded into a docs PR.

Gate on this head: `go test -overlay <pure-Go SQLite rig> ./internal/web/job -count=1` → `ok 8.459s` (the `limit_ip_disconnect_test.go` protocol-gate tests included); `gofmt` clean; all three comment blocks are two lines or fewer.

PR #6553, opened 2026-09-15
fix(limit-ip): leave a reverse client out of the temporary disconnect

BlindMaster24:
## What breaks today

`disconnectClientTemporarily` handles an over-limit client by removing it from the running Xray and adding it back 100 ms later (`internal/web/job/check_client_ip_job.go:669-685`). For a **vless client with a reverse config** that cycle cannot be undone:

1. `RemoveUser` → `proxy/vless/inbound/inbound.go:246` → `RemoveReverse` (`:219-226`) → `outboundHandlerManager.RemoveHandler(a.Reverse.Tag)` (`app/proxyman/outbound/outbound.go:131-146`): the reverse outbound handler is deleted from the tag table.
2. `AddUser` → `buildUserAccount` (`internal/xray/api.go:544-579`) builds `vless.Account{Id, Flow, Testseed, Testpre}` — no `Reverse`, even though the client config carries one.
3. On reconnect `GetReverse` (`inbound.go:193-201`) reads the account from the validator, finds no reverse and fails with *"reverse: user … is not allowed to create reverse proxy"*; the handler is never rebuilt, so the tunnel stays down until Xray restarts.
4. `inbound.go:542-544` rejects forward-proxy use *"for safety reasons"* only while `account.Reverse != nil`, so after step 2 that guard no longer fires for this client either.

Reachable without unusual steps: the reverse handshake goes through `dispatcher.WrapLink`, which calls `trackOnlineIP` with the user's email and source IP — the same online-IP data this job scans — so a reverse client with **Limit IP** set is a normal candidate for a temporary disconnect.

The repo already encodes this constraint elsewhere: `internal/xray/hot_diff.go:124-127` refuses to hot-swap an inbound when `inboundHasReverseClient` and forces a full restart, with a comment saying why. This job had no equivalent gate.

## The change

Skip the cycle for a client whose config carries `reverse`, and log the skip instead of doing it silently:

```go
if reverseClient {
    logger.Warningf("[LIMIT_IP] Not disconnecting %s: its reverse proxy config does not survive a temporary removal", clientEmail)
    return
}
```

What this trades, explicitly: for a reverse client the limit violation stays until the fail2ban ban catches up, which is what actually ends live traffic anyway (the removal only refuses *new* handshakes for 100 ms). Breaking a tunnel that needs an Xray restart to come back is the worse of the two, and it was silent.

The alternative — carrying `reverse` through `buildUserAccount` so the re-added account keeps it — is not in this PR: the removed outbound handler is not rebuilt by an account field alone, so that fix needs the core's behaviour verified against a live Xray (`XRAY_E2E_BINARY`), and it would change the account shape for every vless add.

`TestDisconnectClientTemporarilySkipsReverseClient` drives the function with a DB and a vless client carrying `reverse`, then asserts on the log: with the skip neutralised it fails because the client *is* removed and re-added:

```
--- FAIL: TestDisconnectClientTemporarilySkipsReverseClient (0.21s)
```

Green after the change, and it asserts the skip is *reported* (`Not disconnecting <email>`), not silent — the existing `TestDisconnectClientTemporarilyAllowsHysteria` still passes, so the protocol gate is untouched.

## Gate

* `go test ./internal/web/job -shuffle=on` — green, 9.9 s (local run through an `-overlay` supplying a pure-Go SQLite driver; `go.mod`/`go.sum` are unchanged).
* `gofmt` clean; both comment blocks within the 2-line cap.
* Not covered: nothing here runs against a real core, so the core-side chain above is a reading of the pinned module source, not a reproduction.


BlindMaster24:
Fixed in `e696699c`.

The block was three lines against the 2-line cap this repo sets, and the PR
description claimed otherwise — the description is corrected in the same breath,
so the assertion now matches the code. Nothing else in the PR touches a comment
block over the limit; the reworded block and the test's own block are both at 2.

## MEDIUM pre-existing — confirmed, and it is not this job's to fix

You are right, and I can add the `file:line` for the half you flagged as
unverified. `trafficAddUser` builds its account map from
`internal/web/service/inbound_traffic.go:723-731`, and the field list you quoted
is exactly what it carries — but even carrying `reverse` there would not be
enough: `buildUserAccount` (`internal/xray/api.go:544-579`) never reads
`Reverse` at all and emits no `reverse` key, so the handler could not come back
on this path whatever the caller passed. The fix therefore lands in two places
(`buildUserAccount` plus whatever re-adds the account), which is wider than the
one-gate change this PR is and reaches callers well outside the LIMIT_IP job.

Left in place as you classified it: pre-existing, reported rather than fixed
here, and the natural subject of its own change on the traffic-apply side.

**Gate on this head:** `internal/web/job` and `internal/xray` green (DB-backed
job test through the same local pure-Go SQLite overlay used earlier on this PR).
CI on `e696699c` is running; I will not merge before it and the next review round
are clean.

BlindMaster24:
@claude review

BlindMaster24:
## Round-1 LOW — resolved

Confirmed by your own round 2 (`e696699`, both blocks at `check_client_ip_job.go:656-657` and `:676-677`). Nothing to add.

## MEDIUM pre-existing — the client-edit path: confirmed, and I can now name the fix

You are right that it is a second defect, not the same one. I went back into the
pinned core for both, and your core-side premise (unverifiable in your sandbox,
as you flagged both rounds) is **not** the blocker I described last round — the
handler *can* come back without a restart, and the missing piece is one field:

- `Handler.AddUser` is only `h.validator.Add(u)` (`xray-core/proxy/vless/inbound/inbound.go:233`).
- `Handler.RemoveUser` → `RemoveReverse` → `outboundHandlerManager.RemoveHandler` (`:245`, `:219-227`) — your reading of the loss is correct.
- `Handler.GetReverse` (`:193-205`) rebuilds the handler from the **stored**
  account's `a.Reverse.Tag` when the client next opens a reverse-command
  connection, and it looks that account up through the validator — so a re-added
  account carrying `reverse` gets its handler back on the next reverse dial,
  while an account without the field is permanently refused at `:199-201`
  (`user ... is not allowed to create reverse proxy`).

So the repair is to carry `reverse` into the account on every remove/re-add path:
`buildUserAccount` (`internal/xray/api.go:544-579`) plus the three callers your
two rounds found (`inbound_traffic.go:723-731`, `client_inbound_apply.go:1021-1041`,
`client_bulk.go:1828`). That is a change to the account shape for **every** vless
`AddUser`, across the traffic, client-apply and bulk subsystems — three of them,
each with its own red-first test — which is why it is not going in here rather
than a reason it cannot be done. Recorded as pre-existing; it is the subject of
its own change.

## MEDIUM pre-existing — the `go-test` red: a flake, and half of your diagnosis does not hold

Agreed that it is nondeterministic and not this PR's doing; two corrections to
the evidence.

- **The `:58911` collision is not the handout racing itself.** It is a fixture:
  `internal/amneziawgnet/portfwd_test.go:191-192` holds `collidingPort` open on
  purpose to prove `Reconcile` tolerates a failing bind, so its `address already
  in use` line is the test working, not a fault. Reading it as evidence of a
  race between tests is the one claim in this review I think is wrong.
- **The `:58930` collision is a real flake but not reproducible from this
  change.** The delta on this head is a two-line comment (§ above), so attempt 1
  and the green run at `234d479` are the same machine code; I also ran
  `TestForwardedPortsOnlyChangeStillReconcilesPortForwards` five times shuffled
  on a local machine, all green, and attempt 2 of the same workflow concluded
  success. Nothing in this PR touches `internal/amneziawgnet`.

Not fixed, and not something to fix inside a PR about LIMIT_IP: it is a
fixed-port test (`manager_test.go:289`) whose runner-level collision deserves its
own investigation, and I am reporting it here rather than sweeping it into this
branch.

PR #6554, opened 2026-09-15
fix(ports): refuse an inbound on a port an AmneziaWG peer forwards

BlindMaster24:
## What breaks today

An AmneziaWG peer's `forwardedPorts` are bound by the forward listener on every interface — TCP and UDP, one listener per port (`internal/amneziawgnet/portfwd.go`, `attachTCP`/`attachUDP`). Nothing compares that list with an ordinary inbound's port:

* `checkPortConflictTx` matches the port against other rows' `port` column and against derived SOCKS relay ports — a forwarded port is neither, and it is not a column at all (`internal/web/service/port_conflict.go:245-274`, `:278-307`).
* `checkForwardedPortsConflict` (`internal/web/service/inbound_amneziawg.go:367-397`) does know about them, but it is reached only from the AmneziaWG save paths (`inbound.go:1262`, `client_inbound_apply.go:470`, `:540`, `:772`, `:909`) and only in the other direction: it checks the *client's* forwards against inbounds that already exist.

Trigger: an AmneziaWG inbound has a client with `forwardedPorts` containing 8443; a VLESS inbound is then created (or enabled — see #6549) on 8443. Every guard passes. On the forward side the loser logs a warning and returns `nil`, leaving the peer's forward silently dead (`internal/amneziawgnet/portfwd.go:287-291`); on the Xray side the inbound fails its listen.

## The change

`checkPortConflictTx` now resolves the owner of the port the same way the relay-slot checks next to it do — same host, peers derived from the stored settings through the shared `amneziawg.InstanceFromInbound` — and refuses with a message that names the peer:

```
port 8443 (tcp) already forwarded on inbound 'awg-forward' (#1) on * by its client awg-forward@relay-window
```

Placing it inside `checkPortConflictTx` is deliberate: that is the function both save paths (`AddInbound`, `UpdateInbound`) and the enable check added in #6549 call, so the new direction is enforced at every point where a row's ports can enter the running config — and only for rows on the same host, since the forward listener runs where the AmneziaWG row runs.

`TestAddInboundRefusesAPortAnAmneziaWGPeerForwards` seeds an enabled AmneziaWG row whose client forwards 8443 and creates a VLESS row on it. With the new check neutralised the create is *allowed*:

```
--- FAIL: TestAddInboundRefusesAPortAnAmneziaWGPeerForwards
    --- FAIL: TestAddInboundRefusesAPortAnAmneziaWGPeerForwards/the_forwarded_port
        port_conflict_forwarded_test.go:42: port 8443 is forwarded by a peer of another inbound; the create must be refused
```

Green after the change, with a free-port case in the same table, and `TestAddInboundAllowsANodeRowOnALocallyForwardedPort` pins the scoping that keeps a node row legal on a locally forwarded port.

## Gate

* `go test ./internal/web/service -shuffle=on` — green, 203 s (local run through an `-overlay` supplying a pure-Go SQLite driver; `go.mod`/`go.sum` are unchanged). All existing port-conflict, relay-window and enable-port tests included.
* `gofmt` clean; comments within the 2-line cap.
* Not covered here: PostgreSQL and the Xray gRPC end-to-end tests, as in the neighbouring PRs.


BlindMaster24:
Fixed in `81e63539`. All four LOWs, each red-first except the third, which is
not observable from a test; reasoning below.

### LOW `port_conflict.go:323` — owner set diverged from the listener set

Fixed. The lookup read `instance.Peers` + `ForwardedPortsInclude` while the
supervisor opens a socket only for a peer with an email, a port spec and a
target `desiredPeerTargets` resolves, so the two could disagree. Rather than
copy the gate, `internal/amneziawgnet/portfwd.go` now exposes
`ForwardedPortOwner(inst, port)` and `desiredPortForwardKeys` is built on the
same `forwardingPeers` helper — one definition of "this peer binds something",
used by both sides.

Both halves of your finding are covered by one new test,
`TestAddInboundAllowsAPortNoPeerCanActuallyForward`, which fails on the head you
reviewed:

- `a peer with no email` → was refused with the message you called out verbatim:
  `port 8443 (tcp) already used by inbound 'awg-forward' (#1) on *` — a row whose
  `port` column is 51820, nothing pointing at the forward, because `ForwardedBy`
  was `""` and `String()` took the generic branch.
- `an IPv6-only peer on a row without IPv6` → was refused although no listener
  opens.

With the gate shared, the owner is always a peer that really binds, so
`ForwardedBy` is never empty on this detail and the generic branch is no longer
reachable from here.

### LOW `port_conflict.go:320` — the detail repeated the candidate's address

Fixed: the detail now reports the wildcard bind the forward actually takes
(`Listen: ""` → `on *`), with the reason kept to two lines at the field.
`TestForwardedPortRefusalNamesTheWildcardBind` creates on `10.0.0.5` and asserts
` on * by its client `; before the fix it got
`... on 10.0.0.5 by its client awg-forward@relay-window`.

### LOW `port_conflict.go:303` — a failed `Find` read as "no conflict"

Fixed by changing the helper's signature to `(*portConflictDetail, error)` and
propagating at the call site, matching `checkAmneziawgnetSocksConflict` and the
two checks beside it. No test: a single failure of that one query is not
reachable without dialect-specific failure injection, and on this path it is not
even observable at the outer level — the statement immediately after it queries
the same table and would fail too, so the operator-visible outcome is already an
aborted save. Fixed as the error-handling defect it is and reported as such
rather than dressed up with a test that would pass either way.

### LOW `port_conflict_forwarded_test.go:13` — comment block

Fixed: 2 lines.

## MEDIUM pre-existing — partially closed by this change, remainder confirmed

Your path is now blocked at its entry point rather than left open.
`SetInboundEnable` runs `checkPortConflictTx` (`inbound.go:1586`), and the owner
lookup this PR added runs inside it for the *ordinary* inbound being enabled, so
"enable an ordinary inbound onto a port an AWG peer forwards" is refused in both
save and enable directions, and the AWG-side spec can no longer be saved against
a disabled victim and then quietly collided with later — every ordering of
save/enable between the two rows now meets one of the two guards.

What remains, as you said, is a row that never went through those paths: a
hand-edited or legacy `settings` whose spec collides with an inbound enabled
afterwards, reached by enabling the AWG row itself, which passes its own id as
`ignoreId` and so is skipped by the forwarded-port lookup. Closing that means
reusing `checkAmneziaWGForwardedPorts` in the enable branch, which is a
deliberate decision about the enable path, not a repair to this lookup — left
`pre-existing` and recorded, exactly as you classified it.

One deliberate consequence worth naming, since it reads as a bug otherwise: the
owner lookup counts a *disabled* AWG row's stored spec, so a create onto a port
whose forward is dormant is refused. That mirrors `checkAmneziawgnetSocksConflict`,
which reserves a disabled row's id-derived slot for the same reason — the row
wants the port back the moment it is enabled, and the alternative is the losing
bind this whole family of guards exists to prevent.

**Gate on this head:** `internal/amneziawgnet`, `internal/amneziawg`,
`internal/web/service`, `internal/web/job`, `internal/xray`,
`internal/web/controller` — all green through the local pure-Go SQLite overlay.
The four `egress_domain_test.go` DNS tests fail on this machine both with and
without this change (verified by reverting it), which matches `go-test` being
green on the head you reviewed. CI on `81e63539` is running.

BlindMaster24:
@claude review

BlindMaster24:
Fixed in `35b7d9b6`.

The `desiredPortForwardKeys` doc block was the one of the three I reworded that
still ran to three lines; it is two now. Nothing else in the diff has a block
over the cap — checked mechanically over the added lines of all three commits,
not by eye.

Round 2's judgement on the disabled-row question matches what I intended, and I
have nothing to add to it: the check claims a port for a peer that cannot bind
it *yet*, and the narrower set would reopen the same hole in the other ordering.

BlindMaster24:
@claude review

PR #6558, opened 2026-09-15
fix(clients): keep a vless reverse client's handler across a re-add

BlindMaster24:
## The bug

`RemoveUser` does two things to a VLESS client: it drops the credential, and it
drops the client's reverse outbound handler (`RemoveReverse` →
`outboundManager.RemoveHandler`). Every live remove/re-add path in the panel
rebuilds the account from a hand-built map, and `buildUserAccount` read
`id`, `flow`, `testseed` and `testpre` — never `reverse`. So a reverse client
that was removed and added back could still authenticate, but its tunnel stayed
down until Xray restarted, with nothing logged. The account type is what decides
this: `GetReverse` refuses with `user … is not allowed to create reverse proxy`
when the stored account has no tag.

`claude`'s review of #6553 found the quota path; the same defect is on four
paths, all of which this PR fixes:

| Path | Trigger |
|---|---|
| `client_inbound_apply.go:1040` | editing a client (comment, expiry, totalGB, limitIp…) — the most ordinary action |
| `client_bulk.go:1802` | bulk re-enable from the client list |
| `inbound_traffic.go:727` | traffic reset: a depleted client is removed and re-added on every renewal |
| `client_inbound_apply.go:611`, `runtime/local.go` | adding a client to an inbound that already has one, and the local runtime adds both node-push paths go through |

The LIMIT_IP job is deliberately untouched: it now skips reverse clients entirely
(#6553), so it never removes one.

## The fix

`buildUserAccount` carries the tag, reading it from either shape a caller can
hold — the settings JSON object or the typed client value — via a small
`reverseTag` helper. The five account maps those paths build now include the
client's `reverse`, and the two identical maps in `internal/web/runtime/local.go`
became one `clientUserMap` so a field can no longer be missing from one of them
only.

Not changed, deliberately:

- The account shape for every other protocol, and for a vless client without a
  reverse — nothing is sent when the tag is empty.
- The hot-swap path in `internal/xray/hot_diff.go`, which already refuses to
  hot-swap any inbound carrying a reverse client and restarts instead.
- The four account maps that differ in which fields they carry (some omit
  WireGuard keys). Unifying them would change what a bulk enable sends to a
  WireGuard inbound, which is a behaviour change this fix does not need.

## Core chain, read from the pinned `xray-core`

Static reading of `xray-core@v1.260327.1-0.20260908222543-52a412d9e2f5`:

1. `AddUserOperation.ApplyInbound` → `op.User.ToMemoryUser()`
   (`app/proxyman/command/command.go:38-52`).
2. `User.GetTypedAccount()` deserialises the typed message into the registered
   `*vless.Account` (`common/protocol/user.go:8-24`).
3. `vless.Account.AsAccount()` copies `Reverse: a.Reverse` into the
   `MemoryAccount` the validator stores (`proxy/vless/account.go:24`).
4. `Handler.GetReverse` rebuilds the handler from the stored account's
   `Reverse.Tag` on the client's next reverse-command connection
   (`proxy/vless/inbound/inbound.go:193-205`).

No `-test`/live-core run backs this: `XRAY_E2E_BINARY` is set by no CI job, so
the chain is a reading of the module source, not a reproduction.

## Red first

Every test fails on the commit before the fix:

```
--- FAIL: TestBuildUserAccountCarriesTheReverseTag/settings_json_shape  the account carries reverse tag "", want "portal"
--- FAIL: TestBuildUserAccountCarriesTheReverseTag/typed_client_field   the account carries reverse tag "", want "portal"
--- FAIL: TestClientUserMapCarriesTheReverseTag                        the map carries reverse <nil>, want the client's own *ClientReverse
--- FAIL: TestClientEditKeepsTheReverseTag                             the re-add of rev-edit@example.test carries reverse <nil>, want its stored tag portal
--- FAIL: TestBulkReEnableKeepsTheReverseTag                           the re-add of rev-bulk@example.test carries reverse <nil>, want its stored tag portal
--- FAIL: TestTrafficResetKeepsTheReverseTag                           the re-add of rev-quota@example.test carries reverse <nil>, want its stored tag portal
--- FAIL: TestAddingClientsKeepsTheReverseTag                          the re-add of rev-add-second@example.test carries reverse <nil>, want its stored tag portal
```

The four service tests drive the real service methods (`ClientService.Update`,
`BulkSetEnable`, `ResetClientTraffic`, `AddInboundClient`) against a real temp
DB, with the runtime override recording the maps the panel would push.

## Gate

- `go test` over `internal/xray`, `internal/web/runtime`, `internal/web/service`,
  `internal/web/job`, `internal/web/controller` — green (the DB-backed service
  suite via an `-overlay` supplying a pure-Go SQLite driver; `go.mod`/`go.sum`
  are unchanged).
- `gofmt` clean, added comment blocks within the 2-line cap.
- Not covered here: the live-core behaviour above, and PostgreSQL
  (`postgres-durable-first` is the CI job that would).


BlindMaster24:
Right, and fixed in `67159b68` — the `local.go` hunk and its test are gone, the
file is back to its pre-PR state.

I verified the reachability claim rather than taking it on faith, and it holds
exactly as you described: `grep -rn "\.AddClient(\|\.UpdateUser(" ` over
`internal/` returns only those four sites, and each is on the node side —
`client_inbound_apply.go:641` sits in the `} else {` of `if oldInbound.NodeID ==
nil {` (`:579`), `:1066` and `client_bulk.go:1829` in its own `} else if push {`,
and `client_bulk.go:854` under `if oldInbound.NodeID != nil && !flowChanged …`
(`:835`). `Remote.AddUser` ignoring the map (`internal/web/runtime/remote.go:561`
→ `UpdateInbound`) closes it: on a node inbound the client arrives as a settings
snapshot and is applied by the node's own local branch, which is the map at
`client_inbound_apply.go:611`.

Two corrections to what I wrote, both yours: the extraction was over a dead path,
and the commit message's "the local runtime adds both node-push paths go through"
was simply wrong. The commit is not amended (it is already pushed), so the
history carries the mistake and its removal rather than a silent rewrite.

What the fix consists of after the removal — all four maps reachable through the
local branch, each with its own red-first test:

| Map | Path |
|---|---|
| `client_inbound_apply.go:611` | `AddInboundClient` on a local inbound |
| `client_inbound_apply.go:1050` | `UpdateInboundClient` — the client edit |
| `client_bulk.go:1802` | `bulkSetEnableInboundClients`, the re-enable |
| `inbound_traffic.go:732` | `resetClientTrafficLocked`, the quota renewal |

plus `buildUserAccount`, whose account-level test fails on both input shapes.
The service tests drive the real methods with the runtime boundary stubbed by a
recording `Runtime` override — the maps are observed where production would hand
them to `Local.AddUser`, which passes them straight through to
`xray.XrayAPI.AddUser`.

**Gate on this head:** `internal/xray`, `internal/web/runtime`,
`internal/web/service` green (`-shuffle=on`, DB suite through the local
pure-Go-SQLite overlay); `gofmt`/`go vet` clean. CI is re-running on `67159b68`.

BlindMaster24:
@claude review

PR #6567, opened 2026-09-16
fix(sub): send stable X-HWID on external subscription fetch

### LINKED ISSUE #6559: [Feature]: External subscription fetch sends no X-HWID, so an HWID-limited 3x-ui provider returns 404
### Describe the bug

Hello! I would like to report an architectural issue in version 3.8.0 regarding node synchronization and HWID limits.

**Architecture Context:**
I have 3 servers running 3X-UI. All panels are placed behind Nginx reverse proxies to hide direct ports. Server 1 acts as a "Master" node, and the other two are "Donor" nodes.
To aggregate subscriptions, on the Master node, I add the Donor nodes' subscription links directly into the user's profile settings within the 3X-UI panel (External Subscriptions feature).

**Description of the Problem:**
When the Master node attempts to fetch these external subscription links from the Donor nodes, the Master's internal HTTP client correctly spoofs its User-Agent (e.g., `v2rayNG/1.8.5`), but it **does not send an `X-HWID` header** at all (it is empty).

If the client on the Donor node has the `Max HWIDs` limit enabled (e.g., set to 5), the Donor's panel strictly expects an HWID. Because the Master server sends an empty HWID, the Donor panel blocks the request and returns a `404 Not Found` error.

**Steps to Reproduce:**
1. Install 3X-UI v3.8.0 on both Donor and Master nodes (running behind Nginx).
2. Create a client on the Donor node and set `Max HWIDs: 5`. Copy this client's subscription link.
3. On the Master node, open a client's settings and paste the Donor's subscription link into the external subscriptions field.
4. When the Master node tries to sync/fetch this link, it fails and receives an HTTP 404 error from the Donor node because it doesn't provide an `X-HWID`.

**Current Workaround:**
To fix this temporarily, I had to configure Nginx on the Donor node to intercept requests from the Master's IP and manually inject a fake HWID into the headers:
```nginx
if ($remote_addr = "MASTER_IP") {
    set $target_hwid "Master-Server-ID";
}
proxy_set_header X-Hwid $target_hwid;

1. Set up Node 1 (Master) and Node 2 (Donor).
2. On Node 2 (Donor), set `Max HWIDs` > 0 (e.g., 5) for a client and copy their subscription link.
3. On Node 1 (Master), paste this link into a client's "External Subscriptions".
4. **Actual result:** Node 1 fails to fetch and add the subscription.
5. **Reason:** Node 2 strictly expects an HWID due to the limit, but the 3X-UI panel on Node 1 does not transmit the `X-HWID` header when requesting external links. This results in an HTTP 404 error from Node 2.

### Expected behavior

Once Nginx injects this, the Donor panel hashes it, accepts the request (200 OK), and writes it to the database, consuming one HWID slot for the Master server.

Suggested Solution:
The internal fetcher (Golang HTTP client) that processes external subscription links in 3X-UI should automatically generate and append a persistent X-HWID header (e.g., based on the server's MAC address or a static UUID) when making requests. Alternatively, please add an option to whitelist specific IP addresses to bypass HWID enforcement for node synchronization.

Thank you for your great work on this panel!

### Actual behavior

_No response_

### Relevant logs

```shell

```

### Screenshots

_No response_

### 3x-ui version

3.8.0

### Xray-core version

_No response_

install.sh script

### Operating system

Ubuntu 24.04

_No response_

None

### Before submitting

sdhfsl:
## Problem

Fixes #6559.

A Master panel fetching a donor's subscription link sends no `X-HWID` header (`doFetchSubscriptionLinks` sets only `User-Agent`). When the donor client has an HWID limit, `EnforceHwidForSubID` rejects the empty HWID (`NotSupported`, `Allowed=false`) and the donor answers 404 — Master/Donor aggregation is broken for every HWID-limited client.

## Fix

- `doFetchSubscriptionLinks` now sends `X-HWID` from new `serverHwid()`: a stable per-installation id (`3x-ui-server-<uuid>`), created once and persisted in the settings table, so the donor registers exactly one device slot for the fetching panel (no per-request randomness burning slots; no header at all when the DB is unreachable).
- Donor side untouched — it already hashes and tracks the header correctly.

## Tests

New `TestServerHwidStableAcrossCalls`: non-empty, expected prefix, long enough for the donor's minimum length, stable across calls, persisted in settings. Full `internal/sub` suite passes, `go vet` and gofmt clean.


sdhfsl:
All five points addressed in 227ed818:

- **MEDIUM (concurrent double-UUID)**: `serverHwid()` is now serialized with a process-wide mutex, so concurrent first-time fetches of different URLs cannot mint two ids (single instance owns its DB, so in-process locking is sufficient).
- **LOW (goimports)**: `uuid` moved to the third-party group per `.golangci.yml` local-prefixes — this was the `golangci` failure.
- **LOW (id sent to third parties)**: added `externalSubSendHwid` opt-out (default send, so donor sync works out of the box; set to `false` to stop identifying this panel to non-3x-ui providers).
- **LOW (test coverage)**: new `TestFetchSendsStableHwid` drives `fetchSubscriptionLinks` against httptest and asserts the header is sent by default with the stable value, and omitted after opt-out. Removing the header lines fails it.
- **LOW (comment length)**: both blocks trimmed to the 2-line cap.

`internal/sub` suite green, `go vet` + gofmt clean.

Issue #6022, opened 2026-07-17
[Feature]: Adding the ability to apply additional links and subscriptions to new customers


bl00dychains:
### Is your feature request related to a problem?

I don't consider this a problem, but it's frustrating to manually add an external subscription or link for each user, especially if there are more than 100 users.

### Describe the solution you'd like

Please add a separate function for automatically adding external links and subscriptions to new customers

### Concrete use case

For example, a separate button will appear on the "Customers" tab, or a button will be added when you click the "Advanced" button, which allows you to customize the overall behavior of all existing and future subscriptions.

### Mockups, screenshots, or examples

<img width="473" height="286" alt="Image" src="https://github.com/user-attachments/assets/25f01a54-a212-484f-8734-cacb862a5400" />

<img width="1092" height="145" alt="Image" src="https://github.com/user-attachments/assets/2dbf084e-ab78-477d-9aa3-15042306f579" />

### Which parts of the panel would this affect?

Frontend (UI / panel pages), Subscription (share links / Clash / JSON)

### Are you willing to help?

- [ ] I'd like to implement this feature myself and open a PR.
- [x] I can help test once a PR is open.

### Before submitting

- [x] I searched [existing issues](../issues?q=is%3Aissue) and this feature has not been requested.
- [x] This feature request is written in English.

Issue #5694, opened 2026-06-30
[Feature]: Unified Subscription Link for Groups (Reusing existing aggregation engine)


De-naber:
### Is your feature request related to a problem?

First of all, thank you for the incredible work! The software is absolutely top-notch, and the recent improvements to the "Groups" functionality in the latest releases are perfect.

However, there is a missing piece for a very common real-world scenario. Currently, if I want to provide access to a "Family" (which has multiple users and multiple devices), I have to generate and send a separate subscription link for each client/device. This is quite tedious for the end-user. 

It would be much more logical and user-friendly to share just one single subscription link for the entire Group. The user's app (v2rayNG, Karing, etc.) would then fetch this group link and automatically get all the inbounds/configs assigned to the clients within that group.

### Describe the solution you'd like

Implement a Group Subscription Link.

I took a look at the SQLite database structure and the current backend logic, and it seems almost zero new logic needs to be written. The engine already knows how to do this!

Currently, the backend engine is already capable of aggregating clients that share the same sub_id into a single subscription link (even if the UI normally prevents duplicate sub IDs, the underlying mechanism still works perfectly). We can just reuse this exact mechanism for Groups.

**Proposed Technical Implementation (Backend):**
1. In the database, we just need a new column for group_sub_id (varchar). This can be added either to the client_groups table OR directly to the clients table — whichever is easier to code and requires fewer DB queries.
2. Point the existing subscription aggregation engine to group clients by this new group_sub_id instead of (or alongside) the individual client's sub_id. Since the clients table already contains the group_name mapping, linking them is straightforward.

**Proposed Implementation (Frontend/UI):**
We can reuse the existing UI components.
Currently, when you click on a client, the "Client Information" modal pops up, displaying all their inbounds and subscription links. We can reuse this exact modal/logic for Groups. When clicking on a Group, it should open a similar window displaying the new group_sub_id and the aggregated subscription link.

**Additional Context:**
By reusing the existing aggregation engine and the existing "Client Information" UI modal, this feature can be implemented with minimal code changes while providing a massive quality-of-life improvement for panel admins.

### Alternatives you've considered

The only current workaround is to create a single "Client" and share its subscription link across all family devices. But this completely ruins the ability to track traffic statistics per user/device and breaks the whole purpose of the panel's detailed stats.

### Are you willing to help?

- [ ] I'd like to implement this feature myself and open a PR.
- [ ] I can help test once a PR is open.

### Before submitting

[...]

Issue #5533, opened 2026-06-24
[Feature]: Centralized Links Management and Bulk Assignment to Clients


saeedhasani:
### Is your feature request related to a problem?

The new External Links / External Subscriptions feature introduced in v3.4.0 is very useful, but currently links can only be added individually inside each client.

For small deployments this is acceptable, but for larger panels with hundreds or thousands of users it becomes extremely difficult to manage.

A common use case is adding:

- Backup subscriptions
- Emergency nodes
- Third-party configurations
- Shared VPN endpoints
- Cloudflare-related subscriptions

Currently administrators must manually open every client and add the same link repeatedly.

This does not scale and makes maintenance difficult.

### Describe the solution you'd like

Introduce a dedicated **Links** section in the sidebar, similar to how **Inbounds** are managed.

Example:

Sidebar
- Inbounds
- Clients
- Groups
- Nodes
- Links

Inside the Links section, administrators can:

- Create External Links
- Create External Subscriptions
- Edit existing links
- Enable/disable links
- Delete links

After creating a link, it should be possible to assign it to multiple clients at once, similar to the current inbound assignment workflow.

---

### Example Workflow

1. Create a link in the Links section.
2. Select one or more clients.
3. Assign the link to the selected clients.
4. The assigned link automatically appears in their generated subscription.

If the administrator later updates the link, all assigned clients automatically receive the updated version.

---

### Benefits

- Centralized management
- Much easier maintenance
- No need to edit clients one by one
- Consistent configuration across users
- Better scalability for large deployments
- Similar UX to existing inbound management

---

### Additional Suggestion

It would be useful if Links could support:

- Tags
- Categories
- Enable/disable status
- Usage count (how many clients are using the link)

This would make link management much easier for large-scale installations.

Thank you for your great work on 3x-ui.

### Are you willing to help?

- [x] I'd like to implement this feature myself and open a PR.
- [x] I can help test once a PR is open.

### Before submitting

- [x] I searched [existing issues](../issues?q=is%3Aissue) and this feature has not been requested.
- [x] This feature request is written in English.

PR #6571, opened 2026-09-16
feat(links): make external links a panel-wide library with scopes

BlindMaster24:
## What this is

Turns external links from a per-client list into a **panel-wide library with scopes**, and makes the panel a conforming client when it fetches someone else's subscription.

Today a link can only belong to one client (`client_external_links.client_id`), so "the same backup provider for 300 clients" is 300 copies that drift, and none of the requests behind it (#5533, #6022, #5293, #5694) can be built on top.

## Model

- `external_links` — the shared row: `kind` (`link` / `subscription`), `value`, `remark`, `namePrefix`, `enable`, `expiryTime`, `sortIndex`, its own fetch identity (`userAgent`, `headers`, `cacheTtl`) and the last good expansion (`lastLinks`). Unique on `(kind, value)`.
- `external_link_assignments` — binds a library row to a target with `targetType` = `client` / `group` / `inbound` / `global` / `new_clients`, plus per-binding overrides (`enable`, `expiryTime`, `remark`, `namePrefix`, `sortIndex`). Unique on `(link_id, target_type, target_id)`. `expiryTime` there is `-1` (never) / `0` (inherit) / a timestamp, because a client's own "never" has to stay distinguishable from inheriting a shared expiry.
- Resolver: effective links for a client are the union over its own, its group, its inbounds and the panel, most specific scope winning for overrides, display order stable.
- Migration `migrateClientExternalLinksToLibrary` is idempotent: one library row per distinct `(kind, value)`, one assignment per legacy row keeping that client's own overrides, order and expiry, then the legacy table is retired as `client_external_links_legacy`. No client's expiry is copied onto the shared row.

## Fetch layer (`internal/sub`)

- **Conditional requests** — `If-None-Match` / `If-Modified-Since`; a `304` refreshes the entry nearly for free.
- **Stale-while-revalidate** — a client is answered from cache while the refresh runs behind it, instead of waiting on a 6 s fetch.
- **Negative cache with backoff** honouring `Retry-After`, plus jitter and a per-host bound that covers only background revalidations, so a wave after a restart does not serialise behind them.
- **Cache key = URL + request identity** (UA / headers), so two rows pointing at one URL with different identity do not poison each other.
- **Stored expansion** (`lastLinks`) is read back when a fetch yields nothing, so a restart with the provider down still serves the client.
- **`X-HWID`** the panel sends is the mechanism merged in #6567 (`serverHwid`, persisted per installation), which is the fix for the `404` from an HWID-limited donor (#6559); this PR builds on it rather than duplicating it. Per-row `headers` are layered on top.
- Manual **refresh** endpoint bypasses the cache on the operator's command.

## Wiring

- `GET/POST /panel/api/links/{list,add,del/:id,enable/:id,reorder,targets/:id,assign/:id,unassign/:id,client/:clientId,refresh/:id}`, registered in `internal/web/controller/link.go`, documented in `endpoints.ts`, `openapi.json` regenerated, `docs/` reference page regenerated.
- Cascades: deleting a client, a group or an inbound removes the assignments that pointed at it.
- Client export/import carries the client's own assignments instead of dropping them. `ExportAll`/`ImportClients` are used only by the panel's export/import endpoints, so those links stay panel-local and are never pushed to a node.
- New clients inherit the `new_clients` scope through a batched materialise on creation and on bulk creation.

## Review round 1

Nine findings were posted and each has a disposition in [this reply](https://github.com/MHSanaei/3x-ui/pull/6571#issuecomment-5696370990): both HIGHs (the usage aggregate had no `GROUP BY`; an assignment could not express "never expires") and the MEDIUMs (`last_links` was written but never read, stale generated output, a partial library edit resetting order and identity, a DB read on the fetch path without a nil guard, the host gate serialising the cold wave) are fixed, and the LOW documentation gap is closed. Each fix that can be tested carries a test that fails without it.

While re-reading the same code one more such hole was found and fixed before the review reached it: a body that omits `enable` used to re-enable a disabled row, so a "rename only" edit flipped its state.

## Verification

CI on this branch: `build` (386, amd64, arm64, armv5/6/7, s390x) and `Build for Windows`, `go-test`, `race`, `golangci`, `govulncheck`, `fuzz-smoke`, `codegen`, `frontend`, `postgres-durable-first`, `CodeQL` and `Docs CI` — all green.

What the local machine could and could not do, plainly: this workstation has no C toolchain and its commit limit is exhausted, so `go build` / `go test` / `golangci-lint` die with `out of memory allocating heap arena map` — no local Go build or test run happens for this branch at all, which is why every test here is proven by CI rather than by me. What was run locally: `gofmt -e -l` on every touched file, the unused-import and comment-cap checks, `go run ./tools/openapigen` + `node scripts/build-openapi.mjs` + the docs generator (all of which produced the committed output), and the route ⇄ `endpoints.ts` ⇄ `openapi.json` comparison.

Still unverified by any job: the PostgreSQL **dialect** run beyond `postgres-durable-first` (which runs a single test), so the `GROUP BY` fix rests on PostgreSQL's documented rule plus a SQLite-executed test, not on an executed Postgres query. The Xray gRPC end-to-end tests need `XRAY_E2E_BINARY`, which CI does not set; this change does not touch that path.


BlindMaster24:
Thanks — the two HIGHs were real and are fixed on the new head `8184e114`. Per finding:

**1. HIGH `external_link_library.go:411` — usage aggregate without `GROUP BY`.** Fixed: the outer select is now `… ) AS reach GROUP BY link_id`. Red-first: `TestLibraryUsageCountsEachRowOnItsOwn` seeds two library rows with different reach (one panel-wide, one client-only) and asserts `wide=2, narrow=1`; without the clause SQLite hands one row the panel total and the other 0, so it goes red.

**2. HIGH `external_link_resolver.go:253` — an assignment cannot express "never expires".** Fixed with an explicit sentinel rather than a guess: `model.ExternalLinkExpiryNever` (`-1`) means "never", `0` still means "inherit", anything else wins, applied through one shared `resolvedExpiry` helper used by the resolver, the client form's read-back and the export. The migration no longer puts any client's expiry on the shared row and stores each legacy row's own expiry on its assignment (`0` → sentinel), so an upgrade cannot cut one client off with a sibling's timestamp; the client form's `0` keeps its pre-PR meaning of "never", and import maps `0` the same way. Red-first: `TestAssignmentExpiryNeverBeatsTheSharedRow` (owner keeps the link while an inheriting client loses it) plus the migration assertions now pin the shared row at `0` and the `0`-legacy assignment at the sentinel.

**3. MEDIUM `model.go:1072` — `last_links` written but never read.** Fixed by reading it: `LastExternalLinkLinksByValue` is consulted in `expandEntry` when a subscription fetch yields nothing, which is the case the column's comment claims. The claim in the body and commit message is now true.

**4. MEDIUM stale generated artefacts.** Fixed — `openapigen` + `build-openapi.mjs` + the docs generator were re-run; `codegen` passes on this head.

**5. MEDIUM `external_link_library.go:127` — partial edit resets order and identity.** Fixed: the update path builds an explicit column map instead of a `Select`-forced struct, so `sort_index` is never touched (reorder owns it) and `user_agent` / `headers` / `cache_ttl` are written only when the request carries them. `RowsAffected == 0` now reports "link not found".

**6. MEDIUM `internal/sub/external_subscription.go:272` — DB reads/writes from the fetch path with no `InitDB`.** Fixed: `RecordExternalLinkFetchByValue` and the new reader both return early on a nil `*gorm.DB` (the X-HWID helper already did).

**7. MEDIUM `external_subscription.go:303` — per-host gate serialises the cold-cache wave.** Fixed: the host bound now covers only background revalidations; a caller that actually waits for the fetch is no longer queued behind them.

**8. MEDIUM no CI on the head.** Resolved — the branch is rebased onto `f8db7f6c`, the PR is mergeable, and CI does run now. Its first pass found two real defects, both fixed: `serverHwid`/`serverHwidKey`/`serverHwidMu` ended up declared twice when this branch met #6567's mechanism, and the generated files were stale. Everything except `go-test`/`race` was green on that head; those two are re-running on `8184e114`.

**9. LOW `client_portable.go:23` — export/import payload changed silently.** Fixed: both `endpoints.ts` summaries now state that `externalLinks` travels with the payload and is panel-local, the example shows it, and openapi + the docs MDX are regenerated.

One note for the record, since it changed this PR: `d59b77bc` (#6567) already shipped the X-HWID fix for #6559 using a persisted `3x-ui-server-<uuid>`, so my own panelGuid-based variant was dropped and this branch now builds on that mechanism instead of duplicating it.

PR #6579, opened 2026-09-17
fix(sub): send panel guid as X-HWID on outbound subscription fetch

### LINKED ISSUE #6574: [Bug]: Outbound Subscription fetch fails with HTTP 404 due to missing X-HWID (related to #6559)
### Describe the bug

Hello! 

**Note:** The logic for this feature was already successfully implemented for client links in: **[Feature]: External subscription fetch sends no X-HWID, so an HWID-limited 3x-ui provider returns 404 #6559** (PR #6567).

I tested the latest `dev` build and the External Subscriptions sync now works perfectly. Thank you! 
However, I noticed that the exact same problem still exists in a different part of the panel, specifically in:
**Outbounds -> Subscriptions -> Outbound Subscriptions**

**Description:**
When I add an HWID-limited subscription link from a Donor node into the Master node's "Outbound Subscriptions" menu, it fails to connect and throws the following error:
`Failed to fetch/parse subscription (http 404)`

Just like in the previous issue, the donor's HWID gate blocks the request because the internal fetcher for *Outbound Subscriptions* also does not send the `X-HWID` header. 

**Suggested Solution:**
Could the exact same logic (`GetPanelGuid()` + `X-Device-OS: 3x-ui`) implemented in #6559 for `external_subscription.go` be applied to the fetcher in `outbound_subscription.go` as well?

Thank you!

1. Set up Node 1 (Master) and Node 2 (Donor).
2. On Node 2 (Donor), set `Max HWIDs` > 0 (e.g., 5) for a client and copy their subscription link.
3. On Node 1 (Master), navigate to **Outbounds -> Subscriptions -> Outbound Subscriptions**.
4. Add a new outbound subscription and paste the link from Node 2.
5. **Actual result:** The panel fails to fetch the config with `Failed to fetch/parse subscription (http 404)`.

### Expected behavior

_No response_

### Actual behavior

_No response_

### Relevant logs

```shell

```

### Screenshots

_No response_

### 3x-ui version

3.8.5

### Xray-core version

_No response_

install.sh script

### Operating system

Ubuntu 24.04

_No response_

None

### Before submitting

sdhfsl:
## Problem

Fixes #6574 — same LOT as #6559 (fixed by #6567), other fetch path.

Outbound Subscriptions (`fetchAndStore`) set only `User-Agent`. Against an HWID-limited donor, `EnforceHwidForSubID` rejects the empty HWID and answers 404: `Failed to fetch/parse subscription (http 404)`. #6567 fixed the client-external-links path; this path was missed.

## Fix

- Send the panel's stable id via official `GetPanelGuid()` as `X-HWID` plus `X-Device-OS: 3x-ui` (exactly what the issue suggests), honoring the same `externalSubSendHwid` opt-out introduced in #6567 (default send).
- Uses the official guid instead of minting another id, so a panel occupies one donor slot across both fetch paths.

## Tests

New `TestOutboundFetchSendsHwid` drives `Refresh` against httptest: header present with the panel guid by default, absent after opt-out. Mutation-checked (header lines removed -> test red, restored -> green). `go vet` + gofmt clean.


sdhfsl:
The single red check (`review`) is an infrastructure failure of the Claude review action, not a problem in this PR:

- All code checks are green: `go-test`, `race`, `golangci`, `govulncheck`, `fuzz-smoke`, `codegen`, `frontend`, `postgres-durable-first`, CodeQL and all `build` jobs pass.
- The `review` job failed only at its final guard step ("Fail if the review posted nothing"): the `anthropics/claude-code-action@v1` step exited with `is_error:true` and empty `modelUsage`, so no review comment was ever posted.
- The likely cause is the model argument in `.github/workflows/claude-pr-review.yml:121` (`--model claude-opus-5`), which the installed action / OAuth account does not accept. Either a supported model alias or removing the explicit `--model` so the action uses its default should fix it.

No change to the PR source is indicated by this job. Could a maintainer please fix or re-run the review job? Thanks!

PR #6582, opened 2026-09-17
fix(install): disable firewalld pulled in by fail2ban on RHEL family

### LINKED ISSUE #6569: [Bug]: EL7 - fail2ban setup pulls in firewalld and blocks all panel/inbound ports after reboot
### Describe the bug

On RHEL/CentOS 7 (EL7), the automatic Fail2ban setup that runs during install/upgrade
silently pulls in **`firewalld`** as a transitive dependency and leaves the unit
**enabled**.

`firewalld` is *not* started by the package transaction, so nothing appears to break at
that moment — the panel keeps working normally. But on the **next reboot** it starts with
its default `public` zone, which only allows `ssh` and `dhcpv6-client`. Every 3x-ui port
is then rejected by the host firewall:

* the panel port (e.g. `44328`)
* the subscription port (`2096`)
* every Xray inbound port

Clients get `Connection refused` (the firewall rejects with `icmp-host-prohibited`),
which looks exactly like "the panel died" even though `x-ui` is running and listening
correctly on all ports.

This is especially confusing to diagnose because the operator never configured a
firewall in the first place: on EL7 the dependency is pulled in silently, by a package
transaction that the 3x-ui script triggers on their behalf.

1. Start from a RHEL/CentOS 7 system where `firewalld` is **not** installed. Many
   minimal/VPS images remove it at provisioning time, so verify first:
   `rpm -q firewalld` → `package firewalld is not installed`.
2. Install or upgrade 3x-ui with the official script. During the run the installer
   prints `Setting up Fail2ban for the IP Limit feature...` and then executes:
   `yum makecache -y && yum install epel-release -y`
   `yum -y install fail2ban nftables`
3. Do **not** change anything else. At this point the panel and all inbounds are still
   reachable — nothing looks wrong.
4. Reboot the host (`reboot`).
5. After the reboot, the panel and every inbound port are unreachable
   (`curl: (7) Failed to connect ... Connection refused`), while SSH on port 22 still
   works.
6. Confirm the cause:
   `firewall-cmd --list-all` → `services: dhcpv6-client ssh`, `ports:` empty
   `iptables -L INPUT -n -v` → final `REJECT ... reject-with icmp-host-prohibited`

### Expected behavior

Installing or upgrading 3x-ui should not leave the host in a state where a firewall the
operator never asked for blocks the panel and all inbounds on the next boot.

If `firewalld` really has to be installed on EL7 as a side effect of `fail2ban`, the
installer should either leave it disabled (the `3x-ipl` jail does not need it), or open
the ports 3x-ui relies on, or at minimum warn loudly that a firewall was just enabled.

### Actual behavior

`firewalld` is installed and enabled as a side effect of the fail2ban dependency chain,
with an empty `public` zone. After the next reboot it rejects all traffic except SSH,
taking the panel, the subscription service, and every inbound offline — with no
indication in the 3x-ui output that anything firewall-related was installed.

### Relevant logs

The yum transaction triggered by the installer (note `firewalld` arriving as a
`Dep-Install`, not as a user-requested package):

```text
# yum history info <id>
Command Line   : -y install fail2ban nftables
Packages Altered:
    Install     fail2ban-0.11.2-3.el7.noarch                @epel
    Dep-Install fail2ban-firewalld-0.11.2-3.el7.noarch      @epel
    Dep-Install fail2ban-server-0.11.2-3.el7.noarch         @epel
    Dep-Install firewalld-0.6.3-13.el7_9.noarch             @updates
    Dep-Install python-firewall-0.6.3-13.el7_9.noarch       @updates
    Dep-Install firewalld-filesystem-0.6.3-13.el7_9.noarch  @updates
    Install     nftables-1:0.8-14.el7.x86_64                @base
    ...
```

`firewalld` is the only reverse dependency on EL7, so there is no way to install
`fail2ban` there without it:

```text
# rpm -q --whatrequires firewalld
fail2ban-firewalld-0.11.2-3.el7.noarch
```

The resulting zone / ruleset (empty `ports:` is the whole problem):

```text
# firewall-cmd --list-all
public (active)
  target: default
  interfaces: eth0
  services: dhcpv6-client ssh
  ports:
```

```text
# iptables -L INPUT -n -v --line-numbers
Chain INPUT (policy ACCEPT 0 packets, 0 bytes)
num   pkts bytes target     prot opt in     out     source        destination
1    19668 5223K ACCEPT     all  --  *      *       0.0.0.0/0     0.0.0.0/0     ctstate RELATED,ESTABLISHED
2      727 43600 ACCEPT     all  --  lo     *       0.0.0.0/0     0.0.0.0/0
...
7    22881 4478K REJECT     all  --  *      *       0.0.0.0/0     0.0.0.0/0     reject-with icmp-host-prohibited

# iptables -L IN_public_allow -n -v
Chain IN_public_allow (1 references)
 pkts bytes target     prot opt in     out     source        destination
   18  1104 ACCEPT     tcp  --  *      *       0.0.0.0/0     0.0.0.0/0     tcp dpt:22 ctstate NEW,UNTRACKED
```

`firewalld` was enabled at install time but only *started* at the reboot — the firewall
log contains a single line, timestamped at boot, not at package-install time:

```text
# systemctl is-enabled firewalld ; systemctl is-active firewalld
enabled
active

# cat /var/log/firewalld
2026-09-16 09:39:34 WARNING: AllowZoneDrifting is enabled. This is considered an insecure configuration option. ...
```

Meanwhile the panel and Xray are up and listening — only the firewall is blocking them:

```text
# ss -tlnp | grep -E '44328|2096|3000'
LISTEN 0 4096 [::]:2096   *:*  users:(("x-ui",pid=854,fd=16))
LISTEN 0 4096 [::]:30000  *:*  users:(("xray-linux-amd6",pid=865,fd=9))
LISTEN 0 4096 [::]:30001  *:*  users:(("xray-linux-amd6",pid=865,fd=11))
LISTEN 0 4096 [::]:30003  *:*  users:(("xray-linux-amd6",pid=865,fd=10))
LISTEN 0 4096 [::]:30007  *:*  users:(("xray-linux-amd6",pid=865,fd=5))
LISTEN 0 4096 [::]:44328  *:*  users:(("x-ui",pid=854,fd=10))

# curl -sk -o /dev/null -w '%{http_code}\n' https://127.0.0.1:44328/<basePath>/
200
```

### 3x-ui version

3.8.5

### Xray-core version

26.9.9

install.sh script

### Operating system

CentOS 7.9 (EL7), kernel 5.4.x, KVM VPS. The image had `firewalld` removed at
provisioning time (`yum history` shows an `-y remove firewalld linux-firmware`
transaction), so the host had no firewall at all before the upgrade.

Install / upgrade script

No — direct access

### Before submitting

- [x] I searched [existing issues](https://github.com/MHSanaei/3x-ui/issues?q=is%3Aissue) and this bug has not been reported.

### Additional context

**Root cause.** `x-ui.sh` → `setup_fail2ban_iplimit()` → the `centos)` branch:

```bash
centos)
    if [[ "${VERSION_ID}" =~ ^7 ]]; then
        yum makecache -y && yum install epel-release -y
        yum -y install fail2ban nftables
    else
        dnf makecache -y && dnf -y install fail2ban nftables
    fi
    ;;
```

The comment above this block explains that `nftables` was added so the *default sshd
jail* would not log `nft: not found`, and it explicitly mentions "fresh RHEL-family"
systems. The problem is that on EL7 the same command also drags in `firewalld`, and
nothing in the script ever accounts for that: `install_firewall()` only implements
`ufw`, so there is no firewalld code path anywhere, and no port is ever opened in its
zone.

Because the RPM only *enables* the unit (systemd preset) without starting it, the
breakage is deferred to the next boot — which makes it very hard to connect the symptom
back to the install/upgrade that caused it. Debian/Ubuntu are unaffected, which likely
explains why this has gone unnoticed.

**Impact.** Any EL7 deployment that did not already have `firewalld` installed loses all
non-SSH connectivity after the first reboot following a 3x-ui install or upgrade.

**Suggested fix (any of these).**

1. On the RHEL family, if `firewalld` was pulled in by *this* transaction and was not
   present/running before, leave it disabled afterwards (`systemctl disable --now
   firewalld`). The `3x-ipl` jail uses plain iptables
   (`before = iptables-allports.conf`) and does not need firewalld, so IP Limit keeps
   working.
2. Or open the panel port in the firewalld zone before reloading, and tell the operator
   that inbound ports must be opened manually.
3. Or, at minimum, print a prominent warning when the transaction introduces
   `firewalld`, so the operator is not left debugging a "panel is down" symptom that has
   nothing to do with the panel.

**Workaround for affected users** (IP Limit / fail2ban keeps working):

```bash
systemctl disable --now firewalld
```

**Note.** As of this report, `x-ui.sh` on `main` still contains the same block.

sdhfsl:
## Problem

Fixes #6569.

On RHEL-family hosts where firewalld is not installed, `yum/dnf install fail2ban` can pull firewalld in as a dependency, arriving enabled. Nothing breaks immediately, but after the next reboot its default `public` zone (ssh + dhcpv6-client only) rejects the panel port, the subscription port and every inbound port — looking exactly like a dead panel while `x-ui` runs fine.

## Fix

In `setup_fail2ban_iplimit` (`x-ui.sh`): snapshot firewalld presence before the install; if the transaction introduced it on a host that never had it, `systemctl disable --now firewalld` and print the re-enable command. Hosts already using firewalld, and installs where fail2ban was already present, are untouched. Failure of the disable step never blocks the install.

## Tests

`bash -n` syntax check passes. Logic walked through all four states (had/has matrix); no shellcheck in CI for this repo, verified by reading. Live EL7 verification isn't available in this environment — flagging that honestly.


BlindMaster24:
> ## Problem
> Fixes #6569.
> 
> On RHEL-family hosts where firewalld is not installed, `yum/dnf install fail2ban` can pull firewalld in as a dependency, arriving enabled. Nothing breaks immediately, but after the next reboot its default `public` zone (ssh + dhcpv6-client only) rejects the panel port, the subscription port and every inbound port — looking exactly like a dead panel while `x-ui` runs fine.
> 
> ## Fix
> In `setup_fail2ban_iplimit` (`x-ui.sh`): snapshot firewalld presence before the install; if the transaction introduced it on a host that never had it, `systemctl disable --now firewalld` and print the re-enable command. Hosts already using firewalld, and installs where fail2ban was already present, are untouched. Failure of the disable step never blocks the install.
> 
> ## Tests
> `bash -n` syntax check passes. Logic walked through all four states (had/has matrix); no shellcheck in CI for this repo, verified by reading. Live EL7 verification isn't available in this environment — flagging that honestly.

Check it yourself if you can. If not, tell me so I'll take it over.

sdhfsl:
> > ## 问题
> > 修复了#6569。
> > 在未安装 firewalld 的 RHEL 系列主机上，`yum/dnf install fail2ban`可以将其作为依赖项引入，并启用。虽然不会立即出现任何问题，但在下次重启后，其默认`public`区域（仅支持 SSH 和 DHCPv6 客户端）会拒绝面板端口、订阅端口以及所有入站端口——看起来就像面板完全失效，但实际上`x-ui`运行正常。
> > ## 使固定
> > 在`setup_fail2ban_iplimit`（`x-ui.sh`）中：快照显示安装前 firewalld 的存在情况；如果事务在从未安装过 firewalld 的主机上引入了 firewalld，`systemctl disable --now firewalld`则打印重新启用命令。已在使用 firewalld 的主机以及已安装 fail2ban 的主机将保持不变。禁用步骤的失败不会阻止安装。
> > ## 测试
> > `bash -n`语法检查通过。逻辑遍历了所有四种状态（had/has 矩阵）；此仓库的 CI 中没有 shellcheck，已通过读取验证。此环境中不支持实时 EL7 验证——如实指出。
> 
> 如果可以的话，请自行检查。如果不行，请告诉我，我接手。

Okay, you can take over, thank you.

PR #6585, opened 2026-09-17
fix(ci): keep a refused Claude credential from reddening a pull request

BlindMaster24:
An expired Claude subscription reddens `Claude PR Review` on every pull request, and nothing about the pull request is wrong. This fixes that, and the same refusal in the issue analyst.

## Why it is red

The workflow already knows that an API refusal is not a broken workflow: `Report a review the API refused to run` recognises an exhausted usage window and an overloaded API, says so, and leaves the job green. Its condition is `steps.review.outcome == 'failure'`, because those refusals make the action exit non-zero.

A refused credential does not. From run [35187380192](https://github.com/MHSanaei/3x-ui/actions/runs/35187380192):

| step | conclusion |
|---|---|
| `Run anthropics/claude-code-action@v1` | **success** |
| `Report a review the API refused to run` | success, found no cause |
| `Fail if the review posted nothing` | **failure** |

The refusal is only in the uploaded transcript: `{"error":"oauth_org_not_allowed"}` on the assistant entry, and a result entry with `"is_error": true, "api_error_status": 403`, `Your organization has disabled Claude subscription access for Claude Code`. Neither existing branch matches, so the last step sees "posted nothing" and goes red. The same shape is behind all five failures in the last day: `35159059540`, `35184688775`, `35185722358`, `35186543654`, `35187380192`.

## The change

`claude-pr-review.yml`: a new step, `Report a review the credential refused`, reads the transcript for that 401/403 result (or an `oauth_`/`authentication_`/`invalid_api_key` error), records `skipped=true` and emits a `::warning::`. `Fail if the review posted nothing` gains `steps.refused.outputs.skipped != 'true'`. The existing classifier is untouched, so the usage-limit and overload paths behave exactly as before.

`claude-issue-analyst.yml`: the same step and the same clause on `Fail if the analysis posted no reply`. It never runs for a pull request -- its trigger excludes them -- but it failed seven times in the last day for the same reason, and each failure is a notification.

Deliberately **no comment on the pull request** for this cause: the credential is the maintainer's own condition, and a note about it under every contributor's PR is noise. The maintainer sees the yellow annotation on the run instead of a red check.

## Verification

- Both files parse as YAML; the step ids stay unique.
- The predicate was evaluated against three real transcripts: the refused run (fires), a usage-limited run with `429` + `rate_limit_event: rejected` (does not fire, still handled by the old step) and a completed review of 629 entries with `is_error: false` (does not fire). So only the refused credential changes behaviour.
- No local gate reaches `.github/`: `make verify` does not build or test workflows, so this is inspection plus the transcript evidence above, not a test run.
- `pull_request_target` executes the workflow from the base branch, so this takes effect after it is merged, not while it is open.


BlindMaster24:
This pull request is itself the reproduction: run [35195780703](https://github.com/MHSanaei/3x-ui/actions/runs/35195780703) just failed on `Fail if the review posted nothing` (step 11) with the transcript `{"error":"oauth_org_not_allowed"}` and `api_error_status: 403` — the exact shape this change classifies. With the new step that run ends green with a warning annotation instead. The check goes away once this is merged, because `pull_request_target` runs the workflow from `main`.

PR #6604, opened 2026-09-19
fix(tgbot): scope the add-client wizard to the admin, not the chat

### LINKED ISSUE #6593: Add-client draft is shared between admins in the same group chat
`1691c9ca` gave each chat its own add-client draft, which fixed two admins in
two different chats filling in one client between them. In a group chat they
are in the same chat, so the draft is still shared and the same thing happens
between two people at one keyboard's distance.

`clientDrafts` keys by chat alone
(`internal/web/service/tgbot/tgbot.go`):

```go
type clientDrafts struct {
	mu     sync.Mutex
	drafts map[int64]*clientDraft
}

func (s *clientDrafts) forChat(chatID int64) *clientDraft {
```

so every admin in a group reaches the same `*clientDraft`. `userStateMgr` is
keyed the same way, so the wizard step each of them is on is shared too.

## Reproduction

Against `1c0ce80e`, in `internal/web/service/tgbot`:

```go
func TestProbeTwoUsersOneChatShareDraft(t *testing.T) {
	const groupChat int64 = -1001234567890
	t.Cleanup(func() { addClientDrafts.reset(groupChat) })

	first := addClientDrafts.forChat(groupChat)
	first.Lock()
	first.email = "admin-a@example.com"
	first.Unlock()

	second := addClientDrafts.forChat(groupChat)
	second.Lock()
	got := second.email
	second.email = "admin-b@example.com"
	second.Unlock()

	if first != second {
		t.Fatalf("drafts are distinct; the collision does not reproduce")
	}
	t.Logf("second admin opened the wizard and saw %q already filled in", got)
	first.Lock()
	defer first.Unlock()
	t.Logf("first admin's draft now reads %q", first.email)
}
```

```
second admin opened the wizard and saw "admin-a@example.com" already filled in
first admin's draft now reads "admin-b@example.com"
```

In practice: the second admin's wizard opens pre-filled with the first one's
email and limits, and whichever of them taps a control last decides what the
other creates. Either admin's tap also re-renders the other's inline keyboard,
since the callback carries no owner.

## Notes

Keying the draft and the conversation state by `(chat, user)` rather than by
chat is the shape that worked for us — `message.From.ID` and
`callbackQuery.From.ID` are in scope at each call site in `tgbot_router.go`
(118, 323, 1196, 1238, 1251) and `resetAll` needs no change.

I am not opening a PR for it: we ran a fork of this wizard and have just
dropped it in favour of yours, so I no longer run this code and cannot judge
the UX consequences of the change in anger. The defect seemed worth writing
down rather than leaving with the fork.

sdhfsl:
## Summary

Keys the add-client wizard's draft and conversation state by `(chat, user)` instead of by chat alone, so two admins in one group chat no longer fill in one client between them.

## Why

Closes #6593.

`clientDrafts` and `userStateStore` were keyed by chat id only (`internal/web/service/tgbot/tgbot.go`). That fixed two admins in *different* chats, but in a group chat both admins are in the same chat, so they still reached one `*clientDraft` and one wizard step: the second admin's wizard opened pre-filled with the first one's email and limits, and whichever of them tapped a control last decided what the other created.

A private chat is unaffected — its chat id and user id are equal, so the new key matches what the chat id alone used to be. A message with no sender (a channel post) keys to user 0, an id no admin can hold.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tests only
- [ ] Build / CI / tooling
- [ ] Other

## Areas affected

- [ ] Xray config generation
- [ ] Database / migrations
- [ ] Docker image
- [ ] Multi-node (sub-nodes)
- [x] Telegram bot

## How was this tested?

New regression test `TestAddClientDraftIsPerAdminInGroupChat` (`internal/web/service/tgbot/tgbot_client_draft_per_admin_test.go`): two admins tap `add_client_to` in the same group chat (same chat id, different `From.ID`) and then both render their card again.

Before the change it fails with the defect the issue describes:

```
tgbot_client_draft_per_admin_test.go:53: admin A's card shows email "ix3orqd7",
want its own "pdenag8i" from admin B's draft
```

After it passes, and it also pins the second half of the issue — the step each admin is on is their own: after admin A taps `add_client_ch_default_email`, A is `awaiting_email` and B has no step at all.

Commands run:

- `go test ./internal/web/service/tgbot/` — pass
- `go test -race -shuffle=on -count=1 ./internal/web/service/tgbot/` — pass, no data race
- `gofmt -l internal/web/service/tgbot/` — clean
- `go build ./...` — passes for every package except the root and `internal/web`, which need `frontend/dist` (gitignored). This is unrelated to the change and reproduces on a clean `main` in the same checkout.
- `go test ./...` — the tgbot package and `internal/web/service` pass. On my Windows box five packages fail for environment reasons: root and `internal/web` (`all:dist: no matching files found`), `internal/amneziawgnet` (tunnel tests needing real network), `internal/crypto/nodetoken` (`TestFileKeySourceRejectsLoosePerms` — no POSIX permission bits), `internal/logger` and `internal/web/service/panel` (one timing and one env-var test each). I re-ran each of these on a clean checkout without this change and they fail identically, so none of them is a regression from this PR.

Existing tests that used the old key were updated to the new one (`tgbot_add_client_expiry_test.go`, `tgbot_client_draft_per_chat_test.go`, `tgbot_delete_after_test.go`, `tgbot_draft_render_test.go`); no assertions were weakened.

## Screenshots / recordings

N/A — no UI change.

## Breaking changes

None. The wizard's behavior in private chats and the callback data are unchanged.

## Checklist

- [x] `go build ./...` and the test suite pass locally. *(see note above on the Windows-only environment failures)*
- [ ] For frontend changes: `npm run lint`, `npm run typecheck`, and `npm run build` pass. — N/A, no frontend change.
- [ ] I updated the Wiki / README / API docs if user-facing behavior changed. — N/A

## Scope note

The issue also mentions that both admins in a group still see and can tap the *same* card, so either admin's tap re-renders the other's inline keyboard. This PR does not change that: making the card itself per-admin means deciding where a group's wizard card should live (e.g. moving the flow to a private chat), which is a UX decision the issue author explicitly left open. The change here is limited to the draft and the step, which is the scope the issue's Notes describe.


sdhfsl:
### Implementation notes and what I verified

**Scope vs. #6593's Notes.** The key is `(chat, user)` as the issue proposes: `message.From.ID` on the message path, `callbackQuery.From.ID` on the callback path, all five `addClientDrafts` sites the Notes lists (pre-change `tgbot_router.go:118, 323, 1196, 1238, 1251`), and `resetAll`'s logic untouched.

**Two places I went past the Notes — worth a look:**

1. **The conversation-state store had call sites the Notes did not list.** Leaving them chat-keyed would have left the second half of the defect in place: `userStateMgr.clear(...)` ran on *every* command and *every* callback query (pre-change `tgbot_router.go:85, 98, 107`), so any admin's tap anywhere in a group would have cleared the other admin's pending step. Same for the three `set(chatId, "awaiting_*")` calls in the wizard (pre-change `1069, 1079, 1089`). All of them now take the same key.

2. **`messageActor` guards a nil `message.From`** (a channel post has no sender; the field is optional) and keys it to user 0, an id no admin can hold. The sketch in the issue dereferences `message.From.ID` directly, and the two handlers that call it do not filter out senderless posts.

**Verified:**

- `TestAddClientDraftIsPerAdminInGroupChat` fails before the change with the reported symptom and passes after:

  ```
  admin A's card shows email "ix3orqd7", want its own "pdenag8i" from admin B's draft
  ```

  It drives the real `answerCallback` entry point against a fake Telegram API server and a real DB, and also pins the second half of the issue (after admin A taps "change email", A is `awaiting_email` and B has no step).
- `TestMessageActorSeparatesAdminsInOneChat` pins the key the *typed* steps read, since that handler is a closure inside `OnReceive()` that no test in the repo drives. Mutation-checked: reverting the key to chat-only turns it red (`keys collide: {chatID:… userID:0} ×3`).
- `go test -race -shuffle=on -count=1 ./internal/web/service/tgbot/` — pass, no data race. `gofmt` clean.
- Repo-wide `go test ./...`: the five packages that fail on my Windows box (missing `frontend/dist`, no real network for the tunnel tests, no POSIX permission bits, one timing test, one env-var test) fail identically on a clean checkout without this change, so none is a regression.

**Not verified, so it is not claimed:**

- **No live test.** I have not run this against a real bot in a real group chat — that needs a bot token, two accounts, a group and a running panel. Everything above is at the test level.
- The "admin replies with the email" path is only covered indirectly, through the key it reads. Driving it end-to-end would mean adding long-polling test infrastructure that this repo does not have today; I did not want to put a new harness inside a bug-fix PR.

The scope note in the description still stands: two admins in a group still share one card and one keyboard, and this PR does not change that.

sdhfsl:
Heads-up on the one red check: `build (armv6)` is failing **after** the binary is built, in the release workflow's third-party asset downloads, not in the code.

Evidence that it is not this PR:

- The step's own output shows the armv6 build succeeding — `xui-release: ELF 32-bit LSB executable, ARM, EABI5 … statically linked` then `Static binary confirmed` — before it dies on `##[error]Process completed with exit code 4` ~12s later, in the silent section that follows (the six `fetch` calls for `geoip`/`geosite` from Loyalsoldier / chocolate4u / runetfreedom). `wget`'s exit 4 is a network failure, and it comes back too fast for the `--tries=5 --waitretry=10` retries, i.e. a resolution failure rather than a missing asset.
- The same job passed on this PR's previous commit (`4687a854`, run 35434659089), and all seven other platforms — which run the identical downloads in the same two runs — passed: amd64, arm64, armv5, armv7, 386, s390x and Windows.
- The only change between the passing and failing commits is an added `_test.go` file, which `go build` does not compile.

I cannot re-run the job myself (no write access). Could someone re-run the failed `build (armv6)` job when convenient? All code checks are green otherwise: `go-test`, `race`, `golangci`, `govulncheck`, `fuzz-smoke`, `codegen`, `frontend`, `postgres-durable-first`, CodeQL, `review` and the other seven build platforms.

